PR #8, opened 2026-07-28
Session sharing: publish a session's trace as a Hub dataset

thomwolf:
Share one agent session with a teammate: publish its trace as a Hub dataset, notify them, and (next) let them read it in a panel and fork it.

Design and rationale live in `docs/session-sharing.md`. Draft because the trace panel and the accept/decline half of the inbox are still to come.

## What works

- **All five harnesses.** Claude Code and Codex ship the trace **verbatim** (the Hub reads both natively). Hermes, opencode and OpenClaw are converted to the Hub's documented [STS-Format](https://huggingface.co/docs/hub/session-traces-format).
- **One code path for public and private.** A share always produces a payload dataset; visibility is a flag. "Private" is a gated repo plus `grant_access` per recipient, which the Hub allows without the recipient requesting anything.
- **Redaction blocks, it doesn't warn.** A public share is refused (409) on anything credential-shaped, naming the rule, and creates no repo at all. Token patterns, plus **value** matching against the Space's own secret-bearing env vars, plus every derived artifact — the briefing quotes prompts verbatim, so redacting only the trace would leak what the trace hid. `file-history-snapshot` lines are dropped: they embed whole files.
- **Notification without inbound connectivity.** A pull request on `<recipient>/am-inbox` adding `incoming/<envelope-id>.json`. Merging accepts, closing declines. A non-collaborator can open a PR, so nobody grants anyone write access — and the repo's existence is the opt-in.
- **Receiving defaults to an empty whitelist**, because an inbound trace ends up in front of a coding agent.

## Why the transport is Hub-mediated

Direct Space-to-Space HTTP is ruled out by this app's own security model, not by convenience. `visibility.js` only serves a usable backend when the Space is **private**, so an AM's API is unreachable from outside — and anyone who *could* reach `/api/*` would have `POST /api/sessions/:id/input`, i.e. a shell. Both sides therefore only make outbound calls to the Hub.

## Fixes found along the way

- **`claude --session-id` is not always honoured.** The launch line ends in `|| exec claude`, and when the first invocation exits non-zero (fresh-install onboarding does), Claude picks an id of its own. The pin then matches nothing on disk — which silently broke the Overview digest and `--resume` too, not just sharing. Now re-pinned after launch, as codex and opencode already are.
- **`scripts/` wasn't shipped in the image**, so the exporter the server runs as a child process wouldn't exist in the container.

## Notes for review

- The heavy work (a regex sweep over a multi-MB transcript) runs in a **child process**; this app has wedged on synchronous work before, see `watchdog.js`.
- The SQLite harnesses are read by selecting **one conversation**, never by copying the db — opencode's holds OAuth tokens.
- **Not yet verified: cross-account delivery.** Every notify test was to the sender's own inbox, where the token owns the repo, so the non-collaborator PR path is still untested.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -120,6 +120,9 @@ USER node
 
 # App code + built frontend + runtime config (tmux + prompt rcfile).
 COPY --chown=node:node server/ server/
+# scripts/ is not developer-only: share.js runs scripts/share-session.mjs as a
+# child process to build a share bundle off the event loop, so it must ship.
+COPY --chown=node:node scripts/ scripts/
 COPY --chown=node:node --from=web /web/dist /app/public
 COPY --chown=node:node entrypoint.sh /app/entrypoint.sh
 COPY --chown=node:node tmux.conf session.bashrc /app/
--- a/docs/session-sharing.md
+++ b/docs/session-sharing.md
@@ -189,13 +189,17 @@ sides only ever talk to huggingface.co.
 | Piece | What it is |
 |---|---|
 | **Inbox** | `<recipient>/am-inbox`, a small **public** dataset. Its existence *is* the opt-in — no repo, nobody can send you anything, enforced by the Hub. |
-| **Delivery** | The sender opens a **pull request** adding `incoming/<envelope-id>.json`. Anyone HF-authenticated can open a PR without write access; the author is Hub-authenticated identity that cannot be forged. |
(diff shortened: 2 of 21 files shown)


thomwolf:
Superseded by #10 — same branch content, raised from `share/session-traces` inside this repo.

This one couldn't be reopened: `thomwolf/agent-manager` is no longer a fork of this repo (`fork: false`, `parent: null`, and this repo now reports `forks_count: 0`), which is why this PR was auto-closed on 2026-07-29T07:26Z, why it reports `commits: 0` and `changed_files: 0`, and why both `reopenPullRequest` and `PATCH state=open` fail with "the pull request cannot be reopened". Restoring a detached fork isn't possible through the API.

PR #9, opened 2026-07-29
Files pane: create folders, rename, move and delete entries

thomwolf:
The file browser could only read and receive files: you could walk the tree and upload into it, but making somewhere to upload *to*, renaming anything, or moving anything meant asking an agent to run `mkdir`/`mv`. This adds folder creation, and rename, move and delete for both files and folders.

Rebased onto current `main`.

## Server

Four endpoints beside the existing upload/download, all routed through the same symlink-aware `resolveSafe`:

- `POST /api/files/:id/mkdir?path=&name=`
- `POST /api/files/:id/rename?path=&name=`
- `POST /api/files/:id/move?path=&to=`
- `DELETE /api/files/:id/entry?path=`

Names must be a single path segment (no `/`, `\`, `..`, or NUL), the root itself can't be renamed, moved or deleted, and `mkdir` requires a parent that already exists — one folder, not a whole missing chain. Delete is recursive, which is why the UI asks first.

`move` keeps the entry's name and only changes where it lives. On top of rename's guards it wants a destination that is a folder, and refuses to put a folder inside itself — compared on **real** paths, so a symlink in the destination chain can't smuggle the subtree past the check.

Rename, move and delete refuse on a folder something else depends on — an agent's own workspace, or the shared skills dir — and say which: `that folder is worker's workspace`. From a Files pane at the workspaces root those folders sit right there in the listing, and moving one out from under a running agent breaks its cwd. Deleting the agent first frees its folder.

## Web

Files and folders now share one `EntryRow`, so the row actions are the same for both:

- **Click a folder to make it the destination.** `+ Folder` and `Upload` act on the folder you last clicked — marked with a rule down its edge, and named both in the two buttons' tooltips and in the hint bar. With nothing chosen they act on the folder the tree is rooted at, which the crumbs name. The header hands itself back to the root when you click the tree background, navigate, or rename, move or delete the chosen folder out from under it.
- **`+ Folder`** opens an inline name row inside the folder it will create into — at the top of that listing, opening it if it was collapsed; Enter creates, Escape cancels.
- Reaching for a folder opens it, and only clicking the one already chosen folds it up. The two gestures share the click that already disambiguates expand from open-as-root, so picking a destination can't close what was just picked.
- **Per-row hover actions**: the pencil renames in place with an inline field; the trash swaps the row into `Delete folder and everything in it?` / `Delete file?` with Delete/Cancel. While a row is editing or confirming, its click-to-expand and double-click-to-open gestures are suppressed.
- **Drag to move**: every row is draggable and every folder row is a drop target, as are the crumbs (rooted deep in the tree, they're the only way to send something back up without navigating there first) and the tree background, which stands for the folder the tree is rooted at. The folder under the drag lights up; the row being carried dims. Illegal targets — the entry's own parent, a folder's own subtree — never light up and never accept.
- **An OS file dropped on a folder row** now uploads into *that* folder; the background still uploads into the root, whatever is chosen — a drop names its own target, so it does not consult the selection.
- **Move on touch**, where there is no drag: each row's third action picks the entry up, the hint bar says what's in hand with a way to put it down, and every eligible folder offers `Move here` — the header offers it for the current root.
- After an edit only the affected listings re-read (a move re-reads two: the source parent and the destination), so the rest of the expanded tree stays where it was. Failures show in the hint bar with the server's reason.
- On touch, where there is no hover, the row actions stay visible — same as the sidebar already does.

Listings re-read through a small path-keyed subscription instead of a callback to the immediate parent, because a move changes two directories that aren't in a parent/child relationship, and only those two should re-read.

## Testing

`tsc --noEmit` clean, `vite build` clean. Every endpoint exercised against a throwaway server on a spare port — 47 checks, all passing:

| case | result |
| --- | --- |
| create → rename → delete round-trip, nested folder, non-empty recursive delete | ok |
| move a file up, down and sideways; move a non-empty folder; move to the root | ok |
| move into the folder it's already in (no-op) | ok, nothing changed |
| rename and delete a **file** | ok |
| duplicate name (mkdir, rename, move) | 409 `already exists` |
| `..` in the name, in the path, or in the move destination; blank name | 400 `bad name` / `bad path` |
| missing parent, missing entry, missing destination, the root itself | 400 `bad path` |
| a folder into itself, and into its own subfolder | 400 `can't move a folder into itself` |
| move destination that is a file | 400 `not a folder` |
| an agent's workspace, the shared skills dir | 409 with the naming message |
| a symlink planted in the workspace pointing outside it — mkdir, rename, move and delete through it | all 400, nothing outside the root created, moved or deleted |
| a symlink pointing back into the folder being moved | 400, subtree intact |

The destination behaviour was then driven through real Chromium against a throwaway server, asserting against the filesystem rather than against the UI's own account of itself — 9 checks, all passing:

| case | result |
| --- | --- |
| nothing chosen → both header actions target the root | ok |
| clicking a folder retargets both, marks the row, names it in the hint bar | ok |
| `+ Folder` created `alpha/made-in-alpha` | ok, and nothing at the root |
| the new folder appears inside the parent, which opened for it | ok |
| `Upload` wrote `alpha/uploaded.txt` | ok, and nothing at the root |
| a nested folder as the destination → `alpha/nested/deep` | ok |
| first click opens a folder, a second folds it up, selection survives | ok |
| clicking the tree background hands the header back to the root | ok |
| deleting the chosen folder falls the header back to the root | ok |

That last group caught a real bug, now fixed: the 200ms timer that tells click-to-expand from double-click-to-open stayed pending across a quick `+ Folder`, and fired afterwards to collapse the folder out from under the naming field. It only reproduces under fast interaction — pausing between the two clicks hides it.

Still not driven by hand: drag-and-drop and the touch move path.

Worth noting, and not something this PR changes: a newly created empty folder may not survive a Space restart, since the bucket doesn't persist empty directories (the repo notes this, and re-creates recorded workspace paths on boot).

### DIFF
--- a/server/src/index.js
+++ b/server/src/index.js
@@ -1137,6 +1137,35 @@ app.post('/api/files/:id/rename', (req, res) => {
   res.json({ ok: true });
 });
 
+// Move an entry into another folder under the same root, keeping its name —
+// the drag-and-drop half of rename. `to` is the destination FOLDER ('' = root).
+app.post('/api/files/:id/move', (req, res) => {
+  const s = store.get(req.params.id);
+  if (!s) return res.status(404).json({ error: 'not found' });
+  const { root, target: from } = browsePath(s, String(req.query.path || ''));
+  if (!from || from === root || !fs.existsSync(from)) return res.status(400).json({ error: 'bad path' });
+  const dep = dependedOnDir(from);
+  if (dep) return res.status(409).json({ error: `that folder is ${dep}` });
+  const destDir = resolveSafe(root, String(req.query.to || ''));
+  if (!destDir || !fs.existsSync(destDir)) return res.status(400).json({ error: 'bad path' });
+  try { if (!fs.statSync(destDir).isDirectory()) return res.status(400).json({ error: 'not a folder' }); }
+  catch { return res.status(400).json({ error: 'bad path' }); }
+  // A folder can't land inside itself — compared on real paths, so a symlink in
+  // the destination chain can't smuggle the subtree past the check.
+  try {
+    const realFrom = fs.realpathSync(from);
+    const realDest = fs.realpathSync(destDir);
+    if (realDest === realFrom || realDest.startsWith(realFrom + path.sep)) {
+      return res.status(400).json({ error: "can't move a folder into itself" });
+    }
+  } catch { return res.status(400).json({ error: 'bad path' }); }
+  const to = path.join(destDir, path.basename(from));
+  if (to === from) return res.json({ ok: true }); // already where it was dropped
+  if (fs.existsSync(to)) return res.status(409).json({ error: 'already exists' });
+  try { fs.renameSync(from, to); } catch (e) { return res.status(500).json({ error: e.message }); }
+  res.json({ ok: true });
+});
+
 // Recursive: deleting a folder takes what's inside it with it. The UI asks
 // twice before calling this.
 app.delete('/api/files/:id/entry', (req, res) => {
--- a/server/src/index.js
+++ b/server/src/index.js
@@ -1080,6 +1080,76 @@ app.post('/api/files/:id/upload', (req, res) => {
   req.pipe(out);
 });
 
+// One path segment, no traversal and no separator smuggling.
+const badSegment = (n) =>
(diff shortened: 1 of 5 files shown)


lvwerra:
Should have been addressed with #15 #21 and #22.

PR #20, opened 2026-07-31
Files pane: rename entries, and refuse to touch a folder an agent lives in

lvwerra:
Two pieces of PR #9 that #15 didn't carry over. Straight follow-up to the merged Files pane work.

## Rename

A pencil on row hover turns the name into an input, in place — the thing being renamed stays where it is instead of jumping to a dialog. Enter commits, Esc abandons, blur commits (leaving the field is an answer), and the stem is preselected, so renaming `draft.txt` doesn't mean retyping `.txt`. Works the same on folders.

The server takes a **name**, not a path, so a rename can never quietly move something. That stays `/move`'s job — still to come from #9.

## The guard #9 had and we didn't

`dependedOnDir()` refuses any folder something else is standing on: an agent's own workspace, or the shared skills dir. Renaming one breaks a running agent's cwd — and **DELETE would have removed it outright**, since the delete we merged in #15 only refused the workspace root. Both now refuse, with a reason worth reading:

```
that folder is worker's workspace
that folder is the shared skills folder
```

Credit to #9 — this is its function, carried across.

## Checks

Endpoints: rename a file 200 · slash in the name 400 · onto an existing name 409 · an agent's workspace 409 · the skills folder 409 · the workspace root 400. Delete of an agent's workspace and of the skills folder now 409, where both previously succeeded.

Browser: the editor opens with the stem selected, Enter renames on disk, Esc leaves it alone, a folder renames the same way, and a protected folder reports why it won't.

## Still open from #9

- **Move** (drag-and-drop into a folder, plus the `/move` route)
- **"The folder you click is where new folders and uploads land"** — create and upload currently target the breadcrumb folder, not the expanded one

Both are worth having; neither is in this PR.

### DIFF
--- a/server/src/index.js
+++ b/server/src/index.js
@@ -1414,6 +1414,20 @@ app.put('/api/files/:id/write', express.text({ limit: '8mb', type: '*/*' }), (re
   res.json({ ok: true, size: after.size, mtime: after.mtimeMs });
 });
 
+// Folders something else depends on: an agent's own workspace and the shared
+// skills dir. Renaming or deleting those out from under a running agent breaks
+// its cwd, so the pane refuses. Returns a label to say WHY, or null if it's fair
+// game. (Carried over from #9, which had this before we did.)
+function dependedOnDir(abs) {
+  const real = (p) => { try { return fs.realpathSync(p); } catch { return path.resolve(p); } };
+  const target = real(abs);
+  if (target === real(SKILLS_DIR)) return 'the shared skills folder';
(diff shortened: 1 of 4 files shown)


lvwerra:
Absorbed into #21, which was stacked on this branch and is now retargeted onto `main` — it carries this commit (`d65be1a`, rename + the `dependedOnDir` guard) plus move, folder targeting and the explicit-save work. Closing so there's one PR to review; the branch stays until #21 lands.

PR #30, opened 2026-08-04
Bucket backup: a skip list for folders that are slow and worth nothing

lvwerra:
## What this adds

A token field for folder names to keep out of the backup, on top of the mechanism
from #26. An env directory is thousands of files the Job has to hash on every run,
and not one of them is worth keeping.

```
Skip these folders — type a name and press Enter
[ node_modules × ] [ .venv × ] [ add another…            ]
```

Enter or comma commits, `×` removes, Backspace on an empty box takes the last one
back. Tokens are stored in `am-config.json` under `backup.exclude` and normalized
on both read and write.

**What it saves**, measured on an 805-file tree (800 of them a `.venv`):

| | files hashed | wall |
|---|---|---|
| no excludes | 805 | **7 s** |
| `.venv` skipped | 5 | **1 s** |

That is local disk. The Job reads a FUSE mount, which is slower, so the real
saving is larger.

## The glob rule, which is not obvious

`hf upload` takes repeatable `--exclude` globs, but **`**/env/**` does not match a
top-level `env/`** — verified against the Hub:

| patterns | `env/` at root | `sub/env/` |
|---|---|---|
| `**/env/**` | **kept** | excluded |
| `env/**` + `**/env/**` | excluded | excluded |

So a bare token expands to **both** patterns. With only the `**/` form, the copy at
the bucket root still goes up — exactly the case someone means when they type
`node_modules`. A token containing a slash anchors at the root (`state/claude/cache`
→ `state/claude/cache/**`); one already containing `*` or `?` is passed through
untouched.

The status API reports `exclude` (tokens) and `excludePatterns` (the expansion),
because a skip list nobody can see the expansion of is one nobody can debug.

## Scope: history, not the mirror

The skip list applies to the **dataset history only**. `hf cp` has no
`--include/--exclude` at all, and `hf sync` refuses remote→remote.

The mirror *could* be filtered — inside the Job the bucket is mounted at `/live`, so
`hf buckets sync /live hf://buckets/<mirror>/latest --exclude …` is a local→remote
sync and does support patterns. It would also throw away the property that makes
the mirror nearly free: no bytes move, hashes only, 13,482 files in 20 s. Filtering
it would mean reading and re-uploading 11.4 GB instead, which is a bad trade for
files nobody wanted. So excluded folders still occupy mirror storage and still come
back on a mirror restore, and the settings copy says "stay out of the history"
rather than implying more.

## Safety

Patterns cross into the Job as one space-joined env var and the shell splits them
back apart, so the token charset is what makes that split exact: no whitespace, no
shell metacharacters, 64 chars, 40 tokens max, deduped, leading/trailing slashes
stripped. Validated in the browser *and* server-side, since the server is the side
that must not trust its input.

The script builds the argv itself rather than having patterns interpolated:

```bash
read -ra AM_EXCLUDE_PATS <<< "$AM_EXCLUDE"
for p in "${AM_EXCLUDE_PATS[@]}"; do EXCLUDE_ARGS+=(--exclude "$p"); done
```

Checked in bash *inside a directory containing `env/` and `node_modules/`* to prove
the patterns are not glob-expanded against the container, and with the variable
empty and unset under `set -u` (both give `argc=0`).

Newly excluded folders drop out of the newest commit via the existing
`--delete "*"`, and stay recoverable in the commits before it.

## Verified

End to end with a real Job against the dev bucket, using the real script:

```
skipping: home/** **/home/** state/** **/state/**
Found 30 files to upload          ← bucket holds 61
✓ Uploaded
EXCLUDED FOLDERS LEAKED: none
```

Tokens `['home', 'state', '  /state/  ', 'state']` normalized to `['home','state']`
on the way in, so the dedupe and slash-stripping are exercised too.

The field itself was driven in a headless browser: `Enter` adds a chip, `two words`
is refused, `×` removes.

- 17 tests pass (4 new: the two-pattern rule, path/glob passthrough, normalization
  including shell-shaped input, and that patterns reach the Job as an env var the
  shell can split exactly — with an assertion that they are *not* baked into the
  script).
- `tsc --noEmit` and the web build are clean.
- `server/test/repin.test.mjs` fails in a fresh clone for want of `node-pty`,
  identically on `main` — a missing `npm install`, not this change.
- The probe dataset created while verifying was deleted.

## Not done

- **No presets.** Typing `node_modules` and `.venv` is two actions, and guessing a
  default skip list for someone else's bucket seemed worse than letting them say.
  Easy to add if you want them pre-filled on first run.
- **No effect on an existing history.** Excluded folders already committed stay in
  earlier commits — by design, since that is what history is for — so the dataset
  does not shrink retroactively. `super_squash_history` is the tool for that, and it
  remains undecided (§9).
- The skip field is only shown when a backup could actually run (token + mounted
  bucket), the same gate as the table and the button.

Design, measurements and the glob evidence: `docs/bucket-backup.md` §3.8, §3.9, §10.9.

### DIFF
--- a/docs/bucket-backup.md
+++ b/docs/bucket-backup.md
@@ -63,7 +63,17 @@ the bucket, ever.
    nothing. The dataset gets `--delete "*"` so its *newest* commit matches the
    bucket, while every earlier commit keeps the deleted file recoverable — the
    best of both.
-9. **Restore is a documented script, not a button** (§8).
+9. **The skip list applies to the history, not the mirror.** `hf cp` cannot
+   filter (§3.9), and making the mirror filterable would cost the server-side copy
+   that makes it nearly free. The dataset is where thousands of env files actually
+   hurt — every one of them gets hashed on every run — so that is where they are
+   skipped.
+10. **Skip tokens are folder names, expanded server-side.** A bare name becomes
+   both `name/**` and `**/name/**`, because the second does not imply the first
+   (§3.8). Tokens are validated to hold no whitespace or shell characters: they
+   cross into the Job as one space-joined env var and the shell splits them back
+   apart, so the charset is what makes that split exact.
+11. **Restore is a documented script, not a button** (§8).
 
 ## 2. What we already have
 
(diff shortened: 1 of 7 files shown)


lvwerra:

PR #31, opened 2026-08-05
Bucket backup: say so on the dashboard when it is failing

lvwerra:
## Why

The backups on this Space had been failing for about a day and nothing said so.
The dataset held exactly one commit — `initial commit` — while the settings row
showed a cheerful "Last updated" timestamp, because the *mirror* step kept
succeeding and only the dataset step was broken.

A backup that fails quietly is worse than no backup: you stop thinking about it.

## What was actually wrong

Two causes, from the Job logs (both recorded in `docs/bucket-backup.md` §3.8):

**1. The Hub refuses `.cache/` paths.**

```
✓ Copied
Error: Invalid value. Invalid `path_in_repo` in CommitOperation:
cannot update files under a '.cache/' folder
(path: 'home/.cache/claude-cli-nodejs/…/2026-07-09T08-43-35-094Z.jsonl')
```

Any bucket with an agent cache under `home/.cache/` can therefore *never* produce
a dataset commit.

**2. The walk does not fit the job.** Failures report `Job timeout` after
1h33m–2h7m — against the `--timeout 3000s` (50 min) we pass, so that flag is not
being honoured and something near 1h30m is. Runs against the 49-file dev bucket
finish in 14–17 s; this bucket is 102,691 files on a FUSE mount.

## What this PR does

It does **not** fix either cause. It makes them impossible to miss.

- **The timer records how each run ended**, once per job id, into
  `backup-state.json`: stage, the platform's message, and one line of reason
  lifted from the Job's own logs. The logs are where the useful part lives — the
  platform says `Job timeout`, the logs say the Hub won't commit a `.cache/` path,
  and only the second tells you what to do.
- **`/api/info` carries a health object**, or `null` when all is well — so the
  normal case shows nothing and costs nothing. Read from state, never the Hub,
  because every open tab polls that route every 15 s. Withheld while the Space is
  public, like `secrets`, since it names the operator's repos.
- **A strip above the stage**, on every view, **not dismissible** — a dismissed
  warning is silence again, and silence is the failure mode. It shows the reason,
  links to the label-filtered job list, and opens Settings.
- **Settings gains the detail**: a `Last failure` row with the timestamp, the
  platform message, the run count, and the reason underneath. `Last updated` now
  reads from the last *successful* run rather than the last attempt, which is what
  made the old row misleading.
- **Staleness counts as unwell.** Three intervals with no successful run raises the
  strip even when nothing errored. A timer that never fires, or a token gone bad,
  leaves you exactly as wrongly confident as a failing run — and looks perfectly
  healthy without the check.

## Verified

`BackupBanner` is its own component so it can be rendered in isolation; checked in
a headless browser inside a replica of the real shell (`.app` → `.main`), which is
what confirmed the layout under a long reason:

| state | result |
|---|---|
| failing, 7 in a row, real `.cache` error | one line, reason ellipsised, both links reachable |
| failing once, `Job timeout` | one line |
| stale, never succeeded | "It has never completed a run." |
| stale, succeeded 26 h ago | shows the last good run |
| healthy | renders nothing |

- 20 tests pass; 7 are new and all about *when* the strip appears: silence when
  off, silence when healthy, failing with reason and count, stale after three
  intervals, stale from arming when nothing ever succeeded, the window scaling
  with the interval, and an empty state file.
- `tsc --noEmit` and the web build are clean.
- `server/test/repin.test.mjs` fails in a fresh clone for want of `node-pty`,
  identically on `main`.

## Not done

- **Neither root cause is fixed here.** #30's skip list handles the `.cache`
  rejection if you add `.cache` as a token — and arguably `.cache` should be
  excluded unconditionally, since the Hub cannot accept those paths at all. Say the
  word and I'll add that to #30.
- **The walk may not be fixable with exclusions.** A prod-scale run with `.cache`,
  `node_modules`, `.venv` and `__pycache__` excluded was still walking after 40
  minutes, which suggests `hf upload` enumerates before it filters. If that holds,
  step 2 needs to change shape — a subtree at a time, or a narrower dataset scope —
  rather than a longer exclusion list. §10.10.
- **`--timeout` being ignored** is unexplained; worth a separate look, since it
  means a stuck run occupies whatever the platform's own limit is.

### DIFF
--- a/docs/bucket-backup.md
+++ b/docs/bucket-backup.md
@@ -162,6 +162,34 @@ The dataset received **no new commit**. This is the one control standing between
 the operator's saved logins and the public internet, so it is verified rather
 than assumed.
 
+### 3.8 What actually fails on a real bucket
+
+This Space's backups had been failing for a day before anyone noticed, which is
+the whole reason §7.1 exists. Two separate causes, from the Job logs:
+
+**The Hub refuses `.cache/` paths.** The mirror step succeeds, then:
+
+```
+✓ Copied
+Error: Invalid value. Invalid `path_in_repo` in CommitOperation:
+cannot update files under a '.cache/' folder
+(path: 'home/.cache/claude-cli-nodejs/…/2026-07-09T08-43-35-094Z.jsonl')
+```
+
+A bucket with an agent cache under `home/.cache/` can therefore *never* produce a
+dataset commit. `lvwerra/agent-manager-backup` held exactly one commit — "initial
+commit" — for a day of hourly runs.
+
+**The walk does not fit the job.** Failed runs show `Job timeout` after 1h33m to
+2h7m, against a `--timeout 3000s` (50 min) we asked for — so that flag is not
+being honoured either, and something around 1h30m is. Successful runs on a
+49-file bucket take 14–17 s; this bucket is 102,691 files on a FUSE mount.
+
+Excluding `.cache`, `node_modules`, `.venv` and `__pycache__` was **still running
+after 40 minutes** with no "Found N files" line, i.e. still walking. So exclusions
+fix the `.cache` rejection but may not fix the walk: `hf upload` appears to
+enumerate the tree before filtering it. Unresolved — §10.10.
+
 ### 3.7 Incidental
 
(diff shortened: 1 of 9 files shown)


lvwerra:

PR #32, opened 2026-08-05
Bucket backup: prepopulate the skip list with the caches nothing needs

lvwerra:
Stacked on #30 — base is `backup/exclude-folders`, not `main`. Merge #30 first.

#30 gave the operator a skip list. It starts empty, so a fresh install still
crawls every `node_modules` and `.venv` in the bucket and you only find out by
watching the first backup take an hour. This prepopulates it.

### The default

`node_modules`, `.venv`, `venv`, `__pycache__`, `.cache`, `.npm`, `.pnpm-store`,
`.pytest_cache`, `.mypy_cache`, `.ruff_cache`, `.ipynb_checkpoints`, `.next`,
`.turbo`

Picked by walking this Space's own bucket rather than from taste. Across 24
workspace folders: **11 `node_modules`, 6 `.venv`, 4 `.cache`**, plus a
`$HOME/.cache` holding `ms-playwright`, `google-chrome-for-testing-headless`,
`huggingface`, `uv`, `node-gyp` and the `claude-cli-nodejs` transcripts whose
paths a dataset commit refuses outright. Every name is something a package
manager or toolchain rebuilds on demand.

### What the survey found and the default deliberately omits

- **`.git` — 11 of them, the most common folder in the bucket.** Rank candidates
  by size or count and it sorts to the top; it is also the one thing you would
  most want back. A test pins it out of the default so nobody "optimizes" it in.
- **`dist`, `build`, `target`** — regenerable in most repos, hand-written source
  in enough of them. A default that silently drops work is worse than one that
  copies some junk, so they stay opt-in.

### `undefined` and `[]` now mean different things

`undefined` is "never asked" and takes the default; `[]` is a list the operator
emptied on purpose and stays empty. Without that split, clearing the field in
Settings would refill itself on the next read. Applied on both the read and write
paths, so a PUT that omits the key does not silently persist an empty list.

**Consequence worth a second look:** an existing install's config predates the
key, so it picks the default up and its next run skips those folders. Nothing
already backed up is lost — newly excluded files drop out of the newest commit but
remain in the dataset history.

### Verification

- `backup.test.mjs` 20/20 (17 existing + 3 new), full server chain green,
  `mobile` + `terminal-ui` green (SettingsView copy changed).
- End-to-end on a real server with a clean data dir: `GET /api/config` returns the
  13 defaults with no `am-config.json` on disk, so it is the default path and not a
  persisted value.
- Deployed to am-dev-2 (as part of #30+#31 combined). That Space has an explicitly
  set list (`.venv`, `.cache`) and correctly kept it — the prepopulation does not
  override a configured install.

### DIFF
--- a/docs/bucket-backup.md
+++ b/docs/bucket-backup.md
@@ -196,6 +196,35 @@ What it saves, measured on an 805-file tree (800 of them a `.venv`):
 That is local disk; the Job reads a FUSE mount, which is slower, so the real
 saving is larger than this.
 
+**The list ships prepopulated.** An install that has never set one gets
+`node_modules`, `.venv`, `venv`, `__pycache__`, `.cache`, `.npm`, `.pnpm-store`,
+`.pytest_cache`, `.mypy_cache`, `.ruff_cache`, `.ipynb_checkpoints`, `.next`,
+`.turbo` — every one of them rebuilt on demand by a package manager or toolchain.
+
+The names came from walking this Space's own bucket, not from taste. Across 24
+workspace folders: 11 `node_modules`, 6 `.venv`, 4 `.cache`, plus a `$HOME/.cache`
+holding `ms-playwright`, `google-chrome-for-testing-headless`, `huggingface`, `uv`
+and the `claude-cli-nodejs` transcripts whose paths a dataset commit refuses
+outright.
+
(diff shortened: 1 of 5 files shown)


lvwerra:

PR #37, opened 2026-08-05
One conversation renderer for the Overview card and the session pane

lvwerra:
One renderer for what used to be two: the Overview card and a session's trace, both built on the **exchange** — your prompt, the work, the answer.

Running on **am-dev-3** if you want to click it before reading the diff.

## Why

The card read well and showed too little; the trace viewer showed everything and read badly. Same content, two components, two data paths, two visual languages. The card couldn't show the middle at all: `digest.turnsLog` holds only assistant *text* from the current request — no tool calls, no thinking, nothing before the last prompt.

## What's here

**`web/src/components/conversation/`** — `exchanges.ts` (turns → exchanges → one line per step), `Exchange.tsx` (one exchange at any depth), `ToolCall.tsx` (an expanded call rendered as a command / an edit / a file, not as JSON), `ConversationView.tsx`, `web/src/conversation.css`.

The grammar: the prompt is a tinted band with its `❯` hanging in a 16px gutter, and it's the only thing outside the text column — no rail on the answer, none on the work. One meta line under it carries the work on the left (`▸ 14 steps · 9 tools · 42s · 18.4k tok`, and it *is* the fold control) and, where a surface shows more than one turn, `turn 6/13 01:32 PM` on the right. Step rows use one left column for two meanings: a tool's `✓`/`✗`, or a disclosure triangle greyed when there's nothing to open. Text steps expand in place, so nothing is said twice.

**Overview card** — reads the tail of the trace instead of the digest, so the work unfolds between prompt and answer and `↑ show previous turn` grows history a turn at a time. Opens on the latest turn; pins to the tail only while the agent works. Inline in the list it stays digest-driven — one transcript read per visible agent is not what the Overview is for. The `turnsLog` stepper is gone.

**Session pane** — a `terminal | reader` switch in the bottom bar, next to zoom. It's app-wide, like zoom: both modes show the same session, and what differs is the form, which is why the labels name the form. Reader mode draws *over* the terminal, which stays mounted and connected but loses the keyboard; a shell stays a terminal whatever the switch says. You can reply from it — the card's composer, the same `sendInput`, the same optimistic echo.

**Server** — `pageOf()` takes a negative offset, so "the last N turns" is one request instead of one to learn `total` and another to read.

## The subtle bit

What counts as *the answer*: **the trailing run of messages** — everything the agent said after its last action. Taking the last `final` looked right until a real session disagreed: asked for news, the agent searched, answered, and then replied "No response requested." to a notification — and the harness marked *that* `final`. The card showed the boilerplate and buried the news in the work. A message followed by more tool calls is thinking out loud, not a reply; mid-task there is no answer at all, and the running line carries the current step instead.

## Verification

- `npm test` in `server/` (incl. `trace-tail.test.mjs`) and in `web/` (`exchanges.test.mjs` — the answer rule, tool grouping, prompt splitting, a call cut mid-JSON), `tsc --noEmit`, production build.
- A local driving harness (`web/app-shots.mjs`, excluded from the repo like the design lab — see §9 of the doc) exercises a running instance and checks eleven behaviours: the list view makes zero trace reads, the window makes one, `full history ↗` flips an open pane, focus stays out of the covered terminal, reader mode has a reply line and the echo lands, a shell keeps its terminal, a conversation survives a failed refresh, and merging two agents switches to their group layout.

## Not in this PR (§10 of the doc)

`head.prompts[]`; the sidebar cleanup (`openTrace` and the per-row trace/share buttons); windowing reader mode by exchange — it reads the last 400 turns in one page today and says how many messages it's leaving out; the mobile card-sizing `@media` block.

`docs/conversation-view.md` carries the reasoning, the four depths, where each sidebar affordance goes, and what's left.

### DIFF
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -187,6 +187,9 @@ The pane header gets a two-state segmented control, next to the title:
 ● claude  agent-manager           [ TUI | RENDER ]   ⇪ share   ✕
 ```
 
+- **Only an agent gets the switch.** A shell has no conversation to render, and the files/trace
+  panels are not sessions at all — same rule the Overview uses to decide what is an agent
+  (`cli !== 'shell' && !isPassive(cli)`).
 - **TUI** — today's terminal, untouched.
 - **RENDER** — the same session as a conversation: `TraceView` (already extracted upstream,
   `TracePane.tsx:TraceView`, takes a `src` loader) over `/api/trace/:id`, at D3.
@@ -223,6 +226,12 @@ The pane header gets a two-state segmented control, next to the title:
   tool call is unusable while a task runs; one that never moves makes you chase it.
 - **The prompt band sticks to the top** while you read a long turn. What you want overhead deep
   in someone's 67-step answer is the question it is answering — not a row of numbers.
+- **RENDER can be replied to.** Reading a conversation and answering it are the same act — the
+  card has always known that, and a rendered session that could only be read would send you back
+  to the TUI to type. It is the card's own composer (`.ov-live`), the same `sendInput`, and the
+  same optimistic echo: your prompt appears at the bottom with a `working` line until the
+  transcript catches up. Only a trace with **no agent behind it** — a shared file, an import — is
+  read-only, which is what `readOnly` is for.
 - **Share** moves here from the sidebar. One session, one place.
 - **Handover** ("continue from this trace in a new agent") lives in the RENDER footer, beside
   the provenance line — the only place where its meaning is obvious.
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -143,6 +143,11 @@ Rules that keep this from becoming the ugly viewer in a small box:
   the scroll region, sized by the window (`flex: 1; min-height: 0; overflow-y: auto`) — **never**
   by a `vh` number, which is measured against the viewport and leaves a full-screen phone card
   two-thirds empty.
+- **Only the card you opened reads the trace.** Inline in the Overview list a card is a
+  summary — one prompt, one clamped answer — and the digest already has that. Reading the trace
+  per visible agent, and polling it every three seconds for the working ones, would turn the
+  Overview from one `/api/meta` poll into one transcript read per row. The window (and RENDER)
+  is where the middle gets fetched.
 - **A card says nothing the surface around it already says.** No turn number (there is one turn),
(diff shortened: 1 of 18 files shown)


lvwerra:
## Review round

I reviewed my own diff before asking anyone else to. Ten findings; seven are fixed in 85a08f0, three are follow-ups in §10 of the doc.

**Fixed**

| | |
|---|---|
| The Overview was about to read one transcript per visible agent | The card fetched 120 turns on mount and polled every 3s while working. Fine for the card you opened; a fleet-sized cost for the list view, which mounts a card per row. Inline, a card is a summary the digest already answers — only the **windowed** card reads the trace. |
| A failed refresh blanked the conversation | `ConversationView` returned the error *instead of* the page, so one `EIO` on the bucket mount wiped what you were reading. It's a strip above the turns now; the last good read stays. |
| A poll threw you back to the first search hit | The recount effect keyed on an array whose identity changed every refresh. Only a new query jumps now. |
| The covered terminal kept the keyboard | Blurring on the mode change wasn't enough — the pane becoming active refocuses on a timer, and the header and key bar refocus on click. Keys typed while reading went into the agent's TTY invisibly. Focus goes through one guarded helper. |
| "full history ↗" did nothing to an already-open pane | The card wrote `localStorage`; the pane only read it on mount. Both sides now use `web/src/lib/paneMode.ts`, whose write announces itself. Kills the duplicated `am:pane-mode:` literal too. |
| A session with no transcript 404'd every 3s | Backs off to 30s. |
| RENDER polled while its pane was off-screen | `TerminalPane` passes `visible` down; a paused view stops and catches up on return. |

Also: lab-only CSS moved out of the app stylesheet, and **`web/test/exchanges.test.mjs`** — the answer rule (what counts as the reply, what stays in the work) has been wrong twice and had no coverage. It covers mid-task vs finished turns, a superseded `final`, prompt splitting, tool grouping, and a call cut mid-JSON. `npm test` in `web/`; esbuild (already present for vite) does the transpile, so no test runner is added.

**Verified against a running instance**, not just built: list view makes 0 trace reads, opening the card makes 1, "full history ↗" flips an open pane to RENDER, focus stays out of the terminal under the overlay, and a conversation survives a failed refresh.

**Left as follow-ups** (§10): `head.prompts[]`, the sidebar cleanup, and windowing RENDER by exchange — it reads the last 400 turns in one page today and says so in the header.

PR #46, opened 2026-08-07
Agents: let a spawned peer join your group, and tell sessions the port

thomwolf:
Three things an agent could not do from inside the Space. All three were found while trying to start four sessions inside an existing group from another agent.

## A spawned peer could not join a group

`POST /api/agents` read `cli`, `name` and `path` from the query and never passed `groupId`. `createSession` has always accepted it, and the UI's `POST /api/sessions` passes it — only the agent-facing route didn't. So an agent spawning a peer got it into the right *folder* but loose in the sidebar, outside the group it was meant to work in.

The route now resolves a group the same way it already resolves a path, and the reasoning is the route's own: it defaults `path` to the caller's folder because "peers usually work on the same thing". By exactly that logic the group defaults to the caller's too — a peer spawned to work alongside you belongs beside you in the sidebar.

The rest follows from keeping `path` and `group` consistent:

- **Omitted → the caller's group.** An ungrouped caller spawns an ungrouped peer; nothing special-cased.
- **`?group=` names another one, by id or by display name.** The name matters: `GET /api/agents` reports `group` as a *name*, so a name is the only form an agent ever sees. Matching is case-insensitive and trimmed.
- **`?group=none` opts out.** `path` gets its opt-out for free (`path=.` is the workspaces root); a group needs an explicit sentinel, since "omitted" is already taken by inherit.
- **An unknown name is a 400, not a new group.** This is the one place I deliberately did *not* mirror `path`, which creates a missing folder. A folder is the caller's own business and cheap; the sidebar is the operator's, and group names come from a small set the caller can read off the roster first. A typo should be loud rather than quietly fragmenting the sidebar. The error lists the groups that do exist and points at `group=none`.

The 201 now reports the `group` it joined, so the caller can confirm without re-reading the roster.

## Sessions had no way to learn the API port

`PORT` is configurable (`config.js`, and `Dockerfile` sets `ENV PORT=7860`), but nothing exported it into session environments — sessions got `AM_SESSION`, `AM_NAME`, `AM_ID`, `AM_USER`, `AM_ROOT` and nothing else. The only other source was the number interpolated into the environment skill *at generation time*, and a generated-once document cannot stay correct across a port change.

Sessions now get `AM_PORT`, and the skill template reads `${AM_PORT:-<port>}` rather than baking in a constant. The fallback is the generation-time port, so the doc is still correct for a session started before this change, or by any path that misses the export.

## A wrong port fails silently, which is worse than it sounds

`curl -s` to a port nothing is listening on prints *nothing at all*, and under some agent sandboxes exits 0 rather than failing. The failure therefore reads as "the roster is empty, I have no peers" rather than "wrong port" — a genuinely expensive thing to debug. The skill now says so and points at `curl -sS --fail`.

## Tests

`server/test/spawn-group.test.mjs`, in the style of the existing `test/*.test.mjs` (plain node, no harness, `check(name, got, want)`, temp `DATA_DIR`), wired into `npm test`. 25 checks covering: inheriting the caller's group, the ungrouped-caller case, an unknown caller id, naming a group by id and by display name, the `none` opt-out, an unknown name being refused without creating anything, non-string `?group=` values (Express hands you an array for `?group=a&group=b`), and that the resolved id actually places the session and survives a reload.

Also verified by hand against a server booted on a throwaway `DATA_DIR`: an unknown group is rejected for both a grouped and an ungrouped caller, and group resolution correctly runs after the prompt check, so a promptless request still fails first. The existing suites pass (`repin`, `opencode-resume`, `terminal-modes`, `trace-tail`, and the `migration` boot suite, which confirms the server still boots with these changes).

### DIFF
--- a/server/migration.test.mjs
+++ b/server/migration.test.mjs
@@ -20,8 +20,15 @@ const check = (name, ok, detail = '') => {
 };
 const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
 
+// A test server must not publish skills. `SPACE_ID` is set when this runs inside
+// the Space itself, and it makes generateEnvSkill() fan the environment skill out
+// into every live agent's skills dir — from THIS checkout's template, over the
+// copies the running backend wrote. Those dirs are outside DATA_DIR, so a
+// throwaway DATA_DIR does not contain the damage. Strip both switches.
+const { SPACE_ID, AM_DISTRIBUTE_SKILLS, ...BASE_ENV } = process.env;
+
 const srv = spawn('node', ['src/index.js'], {
-  env: { ...process.env, PORT: String(PORT), DATA_DIR, AM_BASHRC: '/nonexistent' },
+  env: { ...BASE_ENV, PORT: String(PORT), DATA_DIR, AM_BASHRC: '/nonexistent' },
   stdio: ['ignore', 'pipe', 'pipe'],
 });
 let bootLog = '';
--- a/server/package.json
+++ b/server/package.json
@@ -14,7 +14,7 @@
     "start": "node src/index.js",
     "dev": "node --watch src/index.js",
     "test:ui": "node terminal-ui.test.mjs",
-    "test": "node test/repin.test.mjs && node test/opencode-resume.test.mjs && node test/terminal-modes.test.mjs && node test/trace-tail.test.mjs && node migration.test.mjs && node resize.test.mjs"
+    "test": "node test/spawn-group.test.mjs && node test/repin.test.mjs && node test/opencode-resume.test.mjs && node test/terminal-modes.test.mjs && node test/trace-tail.test.mjs && node migration.test.mjs && node resize.test.mjs"
   },
   "engines": {
     "node": ">=20.19"
--- a/server/resize.test.mjs
+++ b/server/resize.test.mjs
@@ -40,10 +40,17 @@ const waitFor = async (fn, timeout = 8000) => {
   return false;
 };
 
+// A test server must not publish skills. `SPACE_ID` is set when this runs inside
+// the Space itself, and it makes generateEnvSkill() fan the environment skill out
+// into every live agent's skills dir — from THIS checkout's template, over the
+// copies the running backend wrote. Those dirs are outside DATA_DIR, so a
+// throwaway DATA_DIR does not contain the damage. Strip both switches.
+const { SPACE_ID, AM_DISTRIBUTE_SKILLS, ...BASE_ENV } = process.env;
+
 const srv = spawn('node', ['src/index.js'], {
   cwd: HERE,
   env: {
-    ...process.env,
+    ...BASE_ENV,
     PORT: String(PORT),
     DATA_DIR,
     AM_BASHRC: '/nonexistent',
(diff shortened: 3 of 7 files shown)


lvwerra:
One small follow-up would make the intended natural-language workflow reliable: please document group creation in the generated environment skill.

Right now the skill explains that `group=<name>` selects an existing group and an unknown name is rejected. Although `POST /api/groups` already lets an agent create one explicitly, that endpoint is not mentioned, so an agent asked “create 4 agents in a group called taskforce” may try spawning directly into `group=taskforce`, get the unknown-group error, and stop.

Could the “Launch a new agent” section include the two-step workflow, for example:

```sh
curl -sS --fail -X POST "http://localhost:${AM_PORT}/api/groups?from=$AM_ID" \
  -H 'content-type: application/json' \
  -d '{"name":"taskforce"}'
```

Then spawn the agents with `group=taskforce` (or the returned group id). This preserves the deliberate typo protection in `resolveSpawnGroup` while making requests like “create 4 agents in a group called taskforce” work consistently.

thomwolf:
Good catch — agreed, and it's exactly the request that started this PR ("start four sessions in an existing group"), so the workflow being undocumented for the *new*-group case is a real gap. Fixed in a0c73ed.

The "Launch a new agent" section now shows both steps:

```sh
GID=$(curl -sS --fail -X POST "http://localhost:${AM_PORT:-7860}/api/groups" \
  -H 'content-type: application/json' -d '{"name":"taskforce"}' | jq -r .id)

curl -sS --fail -X POST "http://localhost:${AM_PORT:-7860}/api/agents?cli=claude&group=$GID&from=$AM_ID" \
  -H 'content-type: text/plain' --data-binary 'Draft the migration plan.'
```

Two small deviations from your snippet, both deliberate:

**It spawns with the returned id, not the name.** Nothing dedupes group names — `groups.create()` just pushes — so two groups can be called `taskforce`, and `resolveSpawnGroup` takes the first match. For the four-agents case the id is the unambiguous handle, so the doc says to prefer it and explains why. I added test coverage for this (`duplicate names resolve to the first match; ids stay exact`) so the guidance and the behaviour can't drift apart.

**It carries no `?from=`.** `POST /api/groups` never calls `sender()` — it's a manager endpoint, not an agent-to-agent call. Since the skill opens by telling agents that every call which changes something takes `?from=$AM_ID` so the operator can see who asked, showing it here would imply an attribution that isn't actually recorded. The doc notes the exception in one line rather than passing a parameter the route ignores.

Happy to instead make `POST /api/groups` record the sender, if you'd rather group creation were attributable — it'd have to stay optional, since the web UI calls it without a sender. Say the word and I'll do it as a separate change.

Verified the snippet verbatim against a server on a throwaway `DATA_DIR`: `jq -r .id` returns the id, the group lands in `/api/tree`, and a spawn with a bogus group still gives the unknown-group error, so the two failure modes stay distinguishable. Suite is at 28 checks, all passing.

lvwerra:
Thanks — I do want group creation attributable. More broadly, every state-changing call should carry an origin so we can reconstruct who performed each operation, whether it came from an agent or the operator UI.

I pushed 456762a to put `?from=$AM_ID` back on the group-creation example and remove the documented exception. I’ll handle enforcement and durable operation logging as a separate PR, including agent creation both with and without an initial prompt, plus group creation. The UI will use an explicit operator origin rather than silently omitting provenance.

PR #51, opened 2026-08-08
Reader on a phone: scroll the trace, and an instrument for the keyboard gap

thomwolf:
Three reader-mode problems reported from a phone: the trace could not be
scrolled with a finger, focusing the reply box opened a blank band above the
keyboard, and the prompt band would not scroll away.

**Now device-verified.** The branch was deployed to a private Space and
exercised on iOS Safari through the Hub page — see *Verification* for exactly
what that does and does not cover.

Reviewed twice. The first review found a regression in the touch fix and a
better hypothesis for the band (`699b938`); the second reviewed the keyboard
change and went and *measured* the layout, which changed what the code claims
(`8da137f`).

## The trace could not be scrolled back

Reader mode draws *inside* `.term-host`, and on a phone that element owns
terminal panning two ways:

- a **capture-phase, non-passive `touchmove`** listener that `preventDefault()`s
  and `stopImmediatePropagation()`s, spending the drag on xterm's scrollback;
- **`touch-action: none`** in the `max-width: 720px` rule — and touch-action is
  intersected down the ancestor chain, so `none` there forbids the browser from
  panning anything nested inside, including the reader's own `.cxv-body`.

Between them, every finger drag over the conversation scrolled the terminal
hidden underneath. (The ▲▼ turn nav still worked, so it was not literally
unreachable.)

The handlers bail when `modeRef.current === 'reader'`, and the frame carries a
`.reading` class that relaxes it to `auto` while covered — `auto`, not `pan-y`,
because `.markdown pre` and `.markdown table` inside a trace are
`overflow-x: auto`, and pinning the axis would have left wide output unreachable
sideways: the same bug one axis over. `.cxv-body` gains
`overscroll-behavior: contain`.

Two ways the terminal's momentum outlived the switch, also fixed: `onTouchEnd`
stands down in reader mode, and the mode effect stops a glide already coasting,
since the handler that would have caught it now stands down too. The flip is
broadcast app-wide over `am:pane-mode`, so neither path needs the hand that
started the gesture.

## The blank band: the app no longer estimates a keyboard

The app pins itself to `window.visualViewport`. Inside a cross-origin iframe —
how the Hub page serves a Space — mobile Safari leaves the child's viewport
unchanged when the keyboard opens, so there is no signal, and the code used to
*assume* a keyboard ate 46% and clamp itself to the remaining 54%.

That guess is the band. The abandoned strip does not stay hidden behind the
keyboard: the parent scroll-reveals the focused field and drags the strip into
view — `body`'s `--bg` against the reader's `--panel`, sized by a constant rather
than by a keyboard. It went unnoticed for as long as it did because in terminal
mode that strip is unused black.

**So the estimate is gone, not retuned.** Out: the `0.54`/`0.48` ratio,
`focusFallbackTimer`, `scheduleEmbeddedFallback`, `embeddedTouchLayout`, the
`focus-fallback` layout mode, and a `focusedInput` that had become write-only —
58 lines out, 24 in. Where real geometry exists nothing changes. `focusBaseline`
stays, because a keyboard is still *detected* as the shrink from it; the second
review's fix stops that baseline being re-taken while the keyboard is already up
(field → field), which would have made a keyboard plainly up measure as absent.

The second review measured the rest at 390×844: `html`, `body`, `#root`, `.app`,
`.main`, `.term-host`, `.pane-reader` and `.cxv` **all have a scroll range of
exactly 0**. A scroll-into-view inside this document moves nothing — the reveal
is wholly the parent page's, and what this document can contribute is to not
fight it by resizing against a keyboard it cannot see.

## The prompt band would not scroll away

`.cxv-col .cx-prompt` was `position: sticky; top: 0`, on the theory that deep
inside a long turn you want what you asked for overhead. In a pane-sized reader
it reads as the message refusing to scroll, and on a phone a wrapped prompt can
be a third of the viewport — most of the screen held back from what you are
scrolling toward.

The whole rule is gone, not just the `position`: the opaque background it also
set existed only so scrolling content would not show through a floating band,
and over `.cxv-body`'s `--panel` the base `color-mix(--accent 7%, transparent)`
composites to the same colour. The band looks identical and scrolls. Removed at
every width, not only under the phone breakpoint — the reason it reads badly is
the pane, not the screen.

## Verification

`tsc --noEmit` clean, `node test/exchanges.test.mjs` → `exchanges: ok`,
production `vite build` succeeds.

**On a device** (iOS Safari, Hub page, private Space running this branch):
reader-mode scrolling works — confirmed by use, since the sticky-prompt report
above could only come from scrolling a trace on a phone — and the prompt band now
scrolls away. Reported working after `9f10fe7`.

**Still not independently confirmed:** whether the composer now sits *above* the
keyboard rather than under it. Removing the clamp trades one failure mode for
another — if the parent page declines to pan far enough, the composer hides
instead of floating above a gap — and no one has reported either outcome
explicitly. Headless Chromium cannot emulate an on-screen keyboard or a
cross-origin iOS visual viewport, so this is not checkable here.

## Follow-ups this PR deliberately does not take

- **`.term-keybar` renders over the reader.** Load-bearing, not cosmetic:
  `.cxv-foot` (22px) and the key-bar (39px) sit below the composer, so the parent
  must pan keyboard **+75px**. esc/tab/arrows/`^C` are TUI controls with no
  meaning over a reader and gating them is one condition — the fix if the
  composer turns out to land under the keyboard. #49 owns that surface.
- **Focusing the composer resizes the agent's PTY**, via the mounted terminal's
  ResizeObserver. Reading a conversation should not perturb the agent. (Deleting
  the clamp incidentally fixes this for the embedded case, since nothing resizes
  any more.)
- **Chromium's `virtualKeyboard` branch is inert as wired** — `overlaysContent`
  is never set, so there is no `boundingRect` and no `geometrychange`. Predates
  this branch; either set it deliberately or drop the branch.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -117,10 +117,13 @@ export default function App() {
 
   // Track the visual viewport so the mobile layout can sit above the on-screen
   // keyboard (which shrinks visualViewport but not the layout viewport on iOS).
-  // The CSS variables pin the app to that viewport's exact rectangle. The Hub
-  // page embeds the app in a cross-origin iframe; mobile Safari leaves that
-  // child viewport unchanged when its keyboard opens. In that one no-signal
-  // case, fall back to a conservative focus-derived visible height.
+  // The CSS variables pin the app to that viewport's exact rectangle.
+  //
+  // Where there is no signal — the Hub page embeds the app in a cross-origin
+  // iframe, and mobile Safari leaves that child viewport unchanged when its
+  // keyboard opens — the app reports the viewport it can see and stops there.
+  // It does not estimate one. Nothing here knows how tall a keyboard is, and
+  // the browser that does already scrolls a focused field into view.
   useEffect(() => {
     const vv = window.visualViewport;
     type VirtualKeyboardLike = EventTarget & { boundingRect?: DOMRectReadOnly };
@@ -133,11 +136,10 @@ export default function App() {
     };
     const root = document.documentElement;
     const keyboardSignalThreshold = 80;
-    const embedded = window.self !== window.top;
-    let focusedInput: Element | null = null;
+    // The viewport as it was before a field took focus — the only thing left
+    // that needs remembering, because a keyboard is detected as the SHRINK from
+    // it (hasKeyboardGeometry), not as an absolute height.
     let focusBaseline: ViewportBaseline | null = null;
-    let focusFallback = false;
-    let focusFallbackTimer: ReturnType<typeof setTimeout> | null = null;
 
     const acceptsKeyboardInput = (target: Element | null): target is HTMLElement => {
       if (!(target instanceof HTMLElement)) return false;
@@ -147,9 +149,6 @@ export default function App() {
       return !['button', 'checkbox', 'color', 'file', 'hidden', 'image', 'radio', 'range', 'reset', 'submit']
         .includes(target.type);
     };
-    const embeddedTouchLayout = () => embedded
-      && window.matchMedia('(max-width: 720px)').matches
-      && (navigator.maxTouchPoints > 0 || window.matchMedia('(pointer: coarse)').matches);
     const captureViewport = (): ViewportBaseline => ({
       width: vv?.width ?? document.documentElement.clientWidth,
       height: vv?.height ?? window.innerHeight,
@@ -178,20 +177,18 @@ export default function App() {
       if (keyboardRect && keyboardRect.height > 0 && keyboardRect.top > top) {
         height = Math.min(height, keyboardRect.top - top);
       }
-      if (hasKeyboardGeometry()) focusFallback = false;
-      if (focusFallback && focusBaseline && acceptsKeyboardInput(document.activeElement)) {
-        // The parent page owns the real visual viewport, but cross-origin frame
-        // isolation prevents us from reading it. A phone keyboard typically
-        // consumes roughly the lower half; 54% visible keeps the xterm prompt
-        // above it without disturbing direct-app browsers with real geometry.
-        const visibleRatio = focusBaseline.width > focusBaseline.height ? 0.48 : 0.54;
-        height = Math.min(height, Math.round(focusBaseline.height * visibleRatio));
-        root.dataset.keyboardLayout = 'focus-fallback';
-      } else if (hasKeyboardGeometry()) {
-        root.dataset.keyboardLayout = 'browser-geometry';
-      } else {
-        delete root.dataset.keyboardLayout;
-      }
+      // When the browser reports no keyboard geometry — a cross-origin frame on
+      // mobile Safari, which the Hub page is — the app does NOT invent a height.
+      // It used to assume a keyboard ate 46% and shrink to fit, and a guess that
+      // is wrong in the safe direction is still wrong: the abandoned strip does
+      // not stay hidden behind the keyboard, because the browser scroll-reveals
+      // a focused field and drags it back into view. That strip is the blank
+      // band under the reader's composer. Leaving the viewport alone hands the
+      // job to the engine that can actually see the keyboard — the same
+      // scroll-into-view that already keeps xterm's pinned helper textarea
+      // above it.
+      if (hasKeyboardGeometry()) root.dataset.keyboardLayout = 'browser-geometry';
+      else delete root.dataset.keyboardLayout;
       root.style.setProperty('--vvw', `${Math.round(width)}px`);
       root.style.setProperty('--vvh', `${Math.round(height)}px`);
       root.style.setProperty('--vv-top', `${Math.round(top)}px`);
@@ -223,54 +220,24 @@ export default function App() {
         focusTimers.add(timer);
       }
     };
-    const scheduleEmbeddedFallback = () => {
-      if (focusFallbackTimer) clearTimeout(focusFallbackTimer);
-      focusFallbackTimer = null;
-      if (!embeddedTouchLayout() || !acceptsKeyboardInput(document.activeElement)) return;
-      focusFallbackTimer = setTimeout(() => {
-        focusFallbackTimer = null;
-        if (document.activeElement === focusedInput && !hasKeyboardGeometry()) {
-          focusFallback = true;
-          apply();
-        }
-      }, 500);
-    };
     const onFocusIn = (event: FocusEvent) => {
       const target = event.target instanceof Element ? event.target : null;
-      if (acceptsKeyboardInput(target)) {
-        focusedInput = target;
-        focusBaseline = captureViewport();
-        focusFallback = false;
-        stabilizeFocus();
-        scheduleEmbeddedFallback();
-        return;
-      }
+      if (acceptsKeyboardInput(target)) focusBaseline = captureViewport();
       stabilizeFocus();
     };
     const onFocusOut = () => {
-      if (focusFallbackTimer) clearTimeout(focusFallbackTimer);
-      focusFallbackTimer = null;
       const timer = setTimeout(() => {
         focusTimers.delete(timer);
         if (!acceptsKeyboardInput(document.activeElement)) {
-          focusedInput = null;
           focusBaseline = null;
-          focusFallback = false;
           apply();
         }
       }, 0);
       focusTimers.add(timer);
     };
     const onOrientationChange = () => {
-      if (focusFallbackTimer) clearTimeout(focusFallbackTimer);
-      focusFallbackTimer = null;
-      focusFallback = false;
-      if (acceptsKeyboardInput(document.activeElement)) {
-        focusedInput = document.activeElement;
-        focusBaseline = captureViewport();
-      }
+      if (acceptsKeyboardInput(document.activeElement)) focusBaseline = captureViewport();
       stabilizeFocus();
-      scheduleEmbeddedFallback();
     };
     apply();
     vv?.addEventListener('resize', onViewportChange);
@@ -284,7 +251,6 @@ export default function App() {
     return () => {
       for (const timer of settleTimers) clearTimeout(timer);
       for (const timer of focusTimers) clearTimeout(timer);
-      if (focusFallbackTimer) clearTimeout(focusFallbackTimer);
       vv?.removeEventListener('resize', onViewportChange);
       vv?.removeEventListener('scroll', onViewportChange);
       vv?.removeEventListener('scrollend', onViewportChange);
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -13,6 +13,11 @@ import Overview from './components/Overview';
 import Locked from './components/Locked';
 import BackupBanner from './components/BackupBanner';
 import Welcome from './components/Welcome';
+import ViewportDebug from './components/ViewportDebug';
+
+// `?vvdebug=1` — a phone has no devtools, and the keyboard layout is guessed
+// when the app is embedded cross-origin. Read once: this never changes mid-run.
+const VV_DEBUG = new URLSearchParams(location.search).has('vvdebug');
 import * as api from './api';
 import type { Cli, GridSpec, MoveTarget, OverviewFilter, Session, Tree } from './types';
 import { onPaneMode, readPaneMode, writePaneMode } from './lib/paneMode';
@@ -823,6 +828,9 @@ export default function App() {
         onToggleDemo={toggleDemo}
       />
       )}
+    {/* Outside .app on purpose: it reports where .app was put, so it must not
+        be inside the box it is measuring. */}
+    {VV_DEBUG && <ViewportDebug />}
     <div className={`app${settingsOpen ? ' app-suspended' : ''}${isMobile ? (mobileStage ? ' m-stage' : ' m-home') : ''}`}>
       {showWelcome && <Welcome onClose={dismissWelcome} />}
       {toast && <div className="toast mono" role="alert">{toast}</div>}
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { lazy, Suspense, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import Sidebar from './components/Sidebar';
 import TerminalPane from './components/TerminalPane';
 import FilesPane from './components/FilesPane';
@@ -13,16 +13,17 @@ import Overview from './components/Overview';
 import Locked from './components/Locked';
 import BackupBanner from './components/BackupBanner';
 import Welcome from './components/Welcome';
-import ViewportDebug from './components/ViewportDebug';
-
-// `?vvdebug=1` — a phone has no devtools, and the keyboard layout is guessed
-// when the app is embedded cross-origin. Read once: this never changes mid-run.
-const VV_DEBUG = new URLSearchParams(location.search).has('vvdebug');
 import * as api from './api';
 import type { Cli, GridSpec, MoveTarget, OverviewFilter, Session, Tree } from './types';
 import { onPaneMode, readPaneMode, writePaneMode } from './lib/paneMode';
 import { isPassive, isRemote } from './types';
 import { GridGlyph, ListGlyph } from './components/icons';
+
+// `?vvdebug=1` — a phone has no devtools, and the keyboard layout is a guess
+// when the app is embedded cross-origin. Read once: it never changes mid-run,
+// and lazily imported so a debug surface is not part of the shipped bundle.
+const VV_DEBUG = new URLSearchParams(location.search).has('vvdebug');
+const ViewportDebug = lazy(() => import('./components/ViewportDebug'));
 
 // Phone-sized viewport: the app becomes two full-screen views (list ⇄ pane).
 function useIsMobile() {
@@ -828,9 +829,10 @@ export default function App() {
         onToggleDemo={toggleDemo}
       />
       )}
-    {/* Outside .app on purpose: it reports where .app was put, so it must not
-        be inside the box it is measuring. */}
-    {VV_DEBUG && <ViewportDebug />}
+    {/* Outside .app: it reports where .app was put. That does not make it
+        immune — it is fixed too, so a displaced fixed subtree would carry it
+        along — but the history it keeps still shows the displacement happening. */}
+    {VV_DEBUG && <Suspense fallback={null}><ViewportDebug /></Suspense>}
     <div className={`app${settingsOpen ? ' app-suspended' : ''}${isMobile ? (mobileStage ? ' m-stage' : ' m-home') : ''}`}>
       {showWelcome && <Welcome onClose={dismissWelcome} />}
       {toast && <div className="toast mono" role="alert">{toast}</div>}
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -138,7 +138,9 @@ export default function App() {
     const keyboardSignalThreshold = 80;
     // The viewport as it was before a field took focus — the only thing left
     // that needs remembering, because a keyboard is detected as the SHRINK from
-    // it (hasKeyboardGeometry), not as an absolute height.
+    // it (hasKeyboardGeometry), not as an absolute height. Since the estimate
+    // was deleted this feeds no layout at all: its one consumer is the
+    // keyboardLayout label, which only ?vvdebug=1 reads.
     let focusBaseline: ViewportBaseline | null = null;
 
     const acceptsKeyboardInput = (target: Element | null): target is HTMLElement => {
@@ -183,10 +185,17 @@ export default function App() {
       // is wrong in the safe direction is still wrong: the abandoned strip does
       // not stay hidden behind the keyboard, because the browser scroll-reveals
       // a focused field and drags it back into view. That strip is the blank
-      // band under the reader's composer. Leaving the viewport alone hands the
-      // job to the engine that can actually see the keyboard — the same
-      // scroll-into-view that already keeps xterm's pinned helper textarea
-      // above it.
+      // band under the reader's composer.
+      //
+      // What replaces it is the PARENT page's scroll, not ours: measured at
+      // 390x844, every box in this document — html, body, #root, .app, .main,
+      // .term-host, .pane-reader, .cxv — has a scroll range of exactly 0, so a
+      // scroll-into-view in here moves nothing. The reveal is entirely the
+      // embedder's, and this document's job is to not fight it by resizing
+      // itself against a keyboard it cannot see.
+      //
+      // Nothing below sizes anything: with the estimate gone, keyboardLayout is
+      // a label for ?vvdebug=1 to read. No CSS matches it.
       if (hasKeyboardGeometry()) root.dataset.keyboardLayout = 'browser-geometry';
       else delete root.dataset.keyboardLayout;
       root.style.setProperty('--vvw', `${Math.round(width)}px`);
@@ -222,7 +231,12 @@ export default function App() {
     };
     const onFocusIn = (event: FocusEvent) => {
       const target = event.target instanceof Element ? event.target : null;
-      if (acceptsKeyboardInput(target)) focusBaseline = captureViewport();
+      // Only when there is no baseline yet: tapping from one field straight to
+      // another keeps the keyboard up, and re-reading here would take the
+      // shrunk viewport as the "before" — after which the shrink measures zero
+      // and a keyboard that is plainly up reads as absent. focusout clears it
+      // when focus really leaves, so this stays fresh without being re-taken.
+      if (acceptsKeyboardInput(target) && !focusBaseline) focusBaseline = captureViewport();
       stabilizeFocus();
     };
     const onFocusOut = () => {
--- a/web/src/components/TerminalPane.tsx
+++ b/web/src/components/TerminalPane.tsx
@@ -208,6 +208,9 @@ export default function TerminalPane({
   const resyncRef = useRef<() => void>(() => {});
   const reconcileScrollRef = useRef<() => void>(() => {});
   const claimRef = useRef<() => void>(() => {});
+  // Reachable from the mode switch: a flick can still be coasting through the
+  // terminal's scrollback when the reader covers it.
+  const stopGlideRef = useRef<() => void>(() => {});
   const reconnectRef = useRef<() => void>(() => {});
   const controllerRef = useRef(false);
   const previousZoomRef = useRef(zoom);
@@ -624,16 +627,19 @@ export default function TerminalPane({
     // gesture without moving xterm's viewport, leaving no reliable way back
     // through history on touch-only devices. xterm 5.5 registers no touch
     // listeners of its own (verified against the bundled lib) and .term-host
-    // sets touch-action:none on mobile, so neither xterm nor the browser will
-    // pan: this is the only touch scrolling a phone has, in every pane.
+    // sets touch-action:none on mobile while the terminal is what you see, so
+    // neither xterm nor the browser will pan: this is the only touch scrolling
+    // a phone has, in every pane. Reader mode is the exception at both ends —
+    // it hands touch-action back (.term-host.reading) and these handlers stand
+    // down — because there the scroller you mean to drag is its own.
     //
     // Convert the drag to whole rows and carry the remainder in `residual`.
     // Quantising each event to a fixed notch instead silently drops whatever
     // does not fill one — a 96px drag moved the view 68px — and that shortfall
     // is what reads as lag, because the text trails the finger by design.
     // scrollLines moves ydisp, the authority the viewport follows.
-    // The frame, not the inner measurement box: .term-host is what carries
-    // touch-action:none and what the user actually drags, and it stays the
+    // The frame, not the inner measurement box: .term-host is what carries the
+    // touch-action rule and what the user actually drags, and it stays the
     // gesture target however the box inside it is nested.
     const frame = frameRef.current ?? host;
     const viewport = host.querySelector<HTMLElement>('.xterm-viewport');
@@ -656,6 +662,7 @@ export default function TerminalPane({
       return rows;
     };
     const stopGlide = () => { if (glideFrame) { cancelAnimationFrame(glideFrame); glideFrame = 0; } };
+    stopGlideRef.current = stopGlide;
     // Reader mode covers this frame with its own scroller. These listeners sit
     // on .term-host in CAPTURE and preventDefault, so without this guard every
     // drag over the conversation was eaten here and spent on the hidden
@@ -698,6 +705,10 @@ export default function TerminalPane({
       touchY = null;
       const v0 = velocity;
       velocity = 0;
+      // A drag that began on the terminal and ended after the switch flipped
+      // must not launch anything: the mode is broadcast app-wide, so the flip
+      // can come from another pane rather than from this hand.
+      if (modeRef.current === 'reader') return;
       // Only a flick coasts. A slow, deliberate drag through history must land
       // exactly where the finger left it — drifting past the line someone was
       // reading is worse than having no momentum at all. 0.4px/ms is about
@@ -810,7 +821,10 @@ export default function TerminalPane({
   // with focus swallows every keystroke into the agent's TTY, invisibly. Hand
   // focus back when the terminal is on top again.
   useEffect(() => {
-    if (reading) termRef.current?.blur();
+    // The glide too: a flick left coasting under the reader keeps moving a
+    // viewport nobody can see, and no touch can catch it — the handler that
+    // would stop it now stands down in this mode.
+    if (reading) { termRef.current?.blur(); stopGlideRef.current(); }
     else if (focused) termRef.current?.focus();
   }, [reading, focused]);
 
--- a/web/src/components/TerminalPane.tsx
+++ b/web/src/components/TerminalPane.tsx
@@ -656,7 +656,13 @@ export default function TerminalPane({
       return rows;
     };
     const stopGlide = () => { if (glideFrame) { cancelAnimationFrame(glideFrame); glideFrame = 0; } };
+    // Reader mode covers this frame with its own scroller. These listeners sit
+    // on .term-host in CAPTURE and preventDefault, so without this guard every
+    // drag over the conversation was eaten here and spent on the hidden
+    // terminal's scrollback: the trace could not be scrolled back at all on a
+    // phone, which is the only place this gesture exists.
     const onTouchStart = (e: TouchEvent) => {
+      if (modeRef.current === 'reader') return;
       stopGlide();               // a new touch takes over from any coasting
       velocity = 0;
       residual = 0;
@@ -664,6 +670,7 @@ export default function TerminalPane({
       lastMoveAt = e.timeStamp;
     };
     const onTouchMove = (e: TouchEvent) => {
+      if (modeRef.current === 'reader') return;
       if (touchY == null || !e.touches.length) return;
       const y = e.touches[0].clientY;
       const deltaY = touchY - y;
@@ -849,7 +856,11 @@ export default function TerminalPane({
           <button className="mini-btn ph-close" title="Close" onClick={(e) => { e.stopPropagation(); onClose(); }}><CloseGlyph /></button>
         </div>
       </div>
-      <div className="term-host" ref={frameRef}>
+      {/* `reading` releases the frame's touch-action: the phone rule pins it to
+          `none` so the drag handler above owns terminal panning, and that also
+          forbids the browser from panning anything nested inside — including
+          the reader's own scroller. */}
+      <div className={`term-host${reading ? ' reading' : ''}`} ref={frameRef}>
         <div className="term-fill" ref={hostRef} />
         {/* Reader mode draws OVER the terminal rather than replacing it: xterm needs
             layout to fit, and detaching tmux costs a repaint and can trip the
--- a/web/src/components/ViewportDebug.tsx
+++ b/web/src/components/ViewportDebug.tsx
@@ -1,0 +1,49 @@
+import { useEffect, useState } from 'react';
+
+/**
+ * A readout of the numbers the mobile keyboard layout is computed from
+ * (App.tsx: the visualViewport effect). iOS Safari has no devtools and the Hub
+ * page embeds the app cross-origin, so when the layout lands in the wrong place
+ * on a phone there is otherwise nothing to read. Opt in with `?vvdebug=1`.
+ *
+ * Sampled on a frame loop rather than on events: the interesting moment is the
+ * keyboard animation, where the values that matter are the ones that DISAGREE
+ * mid-flight.
+ */
+export default function ViewportDebug() {
+  const [lines, setLines] = useState<string[]>([]);
+
+  useEffect(() => {
+    let frame = 0;
+    const read = () => {
+      const vv = window.visualViewport;
+      const root = document.documentElement;
+      const css = (name: string) => root.style.getPropertyValue(name).trim() || '–';
+      const rect = (el: Element | null) => {
+        if (!el) return '–';
+        const r = el.getBoundingClientRect();
+        return `${Math.round(r.top)}→${Math.round(r.bottom)}`;
+      };
+      const active = document.activeElement;
+      const name = active instanceof HTMLElement
+        ? `${active.tagName.toLowerCase()}.${(active.className || '').split(/\s+/)[0] || '-'}`
+        : '–';
+      setLines([
+        `screen ${window.screen.height}  inner ${window.innerHeight}  embed ${window.self !== window.top ? 'y' : 'n'}`,
+        `vv     h ${Math.round(vv?.height ?? -1)}  top ${Math.round(vv?.offsetTop ?? -1)}  scale ${(vv?.scale ?? 1).toFixed(2)}`,
+        `css    vvh ${css('--vvh')}  vv-top ${css('--vv-top')}`,
+        `mode   ${root.dataset.keyboardLayout || 'none'}`,
+        `scroll doc ${Math.round(document.scrollingElement?.scrollTop ?? 0)}  win ${Math.round(window.scrollY)}`,
+        `app    ${rect(document.querySelector('.app'))}`,
+        `reader ${rect(document.querySelector('.pane-reader'))}`,
+        `box    ${rect(document.querySelector('.cxv-live textarea'))}`,
+        `focus  ${name}`,
+      ]);
+      frame = requestAnimationFrame(read);
+    };
+    frame = requestAnimationFrame(read);
+    return () => cancelAnimationFrame(frame);
+  }, []);
+
+  return <pre className="vvdebug">{lines.join('\n')}</pre>;
+}
--- a/web/src/components/ViewportDebug.tsx
+++ b/web/src/components/ViewportDebug.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useRef } from 'react';
 
 /**
  * A readout of the numbers the mobile keyboard layout is computed from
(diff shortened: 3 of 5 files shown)


thomwolf (commented):
Read the whole touch path, the `visualViewport` effect, and the touch-action chain
end to end. **The diagnosis in part 1 is correct and the mechanism is described
accurately** — I checked each claim below. The touch fix is the right shape but is
one axis short and two handlers short. The keyboard section is honest about not
knowing, which I'd rather have than a tuned constant; I think the arithmetic
*does* close, just not the way the PR body assumes, and I've written that up as
the main thing in here.

Inline comments carry the specifics. Summary, ranked.

---

## Must fix

**1. `pan-y` re-breaks the same bug one axis over.** `styles.css:1109`.
`.markdown pre` and `.markdown table` are `overflow-x: auto` (`styles.css:1032`,
`:1036`) and render inside `.cxv-body` as `.markdown.cx-md` (`Exchange.tsx:221`).
touch-action intersects down the chain, so `pan-y` on `.term-host.reading`
forbids the browser from panning them sideways — a wide code block or a table in
a trace is unscrollable on a phone, for exactly the reason the trace itself was.
`auto` is what you want (or `pan-x pan-y` if pinch-zoom over the reader is
unwanted). Confidence: high — this is a desk-checkable CSS chain, and it was
checkable here.

**2. Two of the four touch handlers are guarded.** `TerminalPane.tsx:697`,
`:725`. `onTouchEnd` launches the glide and is not guarded; `onTouchStart`
returning early also skips its `stopGlide()`. So: flick the terminal, flip the
mode bar while the coast is still running, and the hidden terminal keeps gliding
under the reader with no touch able to stop it; and a flip *mid-drag* leaves
`velocity` at its last sampled value, so the unguarded `onTouchEnd` starts a
glide on a terminal nobody is looking at. The mode is app-wide and broadcast
(`am:pane-mode`, `lib/paneMode.ts`), so this is not only a "who taps mid-flick"
path. Two lines: guard `onTouchEnd`, and add `stopGlide()` to the effect at
`:812` that already blurs the terminal on entering reader mode. Confidence: high
that it's reachable; medium that a user notices (the symptom is "my scrollback
moved while I was reading").

**3. The comment that states the invariant is now false.**
`TerminalPane.tsx:626-631` and `:636` both assert that `.term-host` sets
`touch-action: none` on mobile, full stop — ":627 … so neither xterm nor the
browser will pan: this is the only touch scrolling a phone has, in every pane."
That is the premise the new guard 30 lines below depends on *not* holding
unconditionally. Given how much this file invests in comments that explain why,
leaving the old one to contradict the new one is the expensive kind of drift.

**4. `var(--mono, monospace)` — the token is `--font-mono`.** `styles.css:530`.
Falls back silently to plain `monospace`, so the readout is not in Geist Mono and
nothing tells you. (`:1175` already has the same typo, so it's
consistent-with-a-bug rather than novel.) Trivial, and one grep here would have
caught it.

---

## The keyboard bug: the hypothesis I'd put first

The PR says the arithmetic predicts ~6% of the screen and the report says a
keyboard's worth. I think both halves of that estimate are off, and the gap
closes without touching the ratio:

**(a) 54% is applied to the wrong denominator.** `App.tsx:186-187` multiplies
`focusBaseline.height`, which is `vv.height` — inside a cross-origin iframe, the
*iframe's* height. The comment's heuristic ("a phone keyboard typically consumes
roughly the lower half") is about the *screen*. On the Hub the iframe is the
viewport minus the page's own chrome, so `H_iframe < S`, and the shortfall grows
as the iframe gets smaller relative to the screen. The ~6% estimate assumed
`H_iframe ≈ S`.

**(b) The abandoned 46% is not necessarily behind the keyboard.** This is the
part I think the PR body assumes away. The estimate `100% − 54% − keyboard`
treats the height `.app` gives up as landing under the keyboard, hidden. But the
composer is a real focusable `<textarea>`, so the embedder does its own
scroll-to-reveal: the Hub page scrolls the iframe up until the focused field is
comfortably clear of the keyboard. `.app` is now only `0.54 · H_iframe` tall and
pinned to the iframe's top, so what that scroll reveals below `.app` is the
abandoned `0.46 · H_iframe` — and because `html, body, #root` are
`overflow: hidden` with nothing in them, that strip paints `body`'s `--bg`
(`#eef1f3` light, `#0b0f13` dark) against the reader's `--panel` (`#ffffff` /
`#11181e`). A visibly different strip between the app and the keyboard, sized
`0.46 · H_iframe` ≈ 350-400px on a phone. That is the reported symptom, at the
reported size, with no re-tuning: the bug is the denominator plus the assumption
that giving up height hides it.

If that's right, the fix is not a new constant. Two candidates, both smaller:
- Derive the clamp from a screen-relative quantity rather than the frame's own
  height (`window.screen.height` is available and the overlay already prints it).
- Better: **skip the fallback clamp when the focused element is a real input.**
  The clamp exists for xterm, whose only input is a deliberately-pinned 1×1
  helper textarea (`styles.css:1112-1117`) — nothing the browser can usefully
  reveal, hence the guess. The reader's composer is a real field, and WebKit and
  the embedder will reveal it natively. The clamp is fighting a mechanism that
  already works in exactly the case it's breaking.

**A discriminator that needs no instrument at all, and I'd add it to the PR
body.** `acceptsKeyboardInput` doesn't care which field it is, so the clamp also
fires for the reader's **search box** (`ConversationView.tsx:196`, `.cxv-search`,
at the *top* of the reader) and for the pane-header rename input. Ask the
reporter to tap search instead of the composer:
- band appears there too → it's the clamp (nothing needs revealing; the field is
  already at the top, so scroll-to-reveal can't be the mover);
- band only on the composer at the bottom → it's scroll-to-reveal moving a fixed
  subtree.

That's the entire fork, from one tap, with no build. It also stands on its own as
a bug: focusing a search box at the top of the pane halves the results list under
it for no benefit.

**Question zero, which the readout answers in one glance and the body should
state as an assumption rather than a fact:** the whole story rests on mobile
Safari leaving a child frame's visual viewport unchanged. If `mode` reads
`browser-geometry` on the phone, `focusFallback` never engaged and the cause is
elsewhere entirely. Worth naming, because it's the cheapest thing on the list and
it can invalidate everything above it.

**Two more things visible in the code:**

- `apply()` feeds `.app`'s `top` and `height` from different clocks in the
  fallback branch: `top` is live `vv.offsetTop`, `height` is derived from a
  `focusBaseline` captured before the keyboard. That is the shape a
  double-subtraction takes, even if `offsetTop` happens to be 0 in an iframe
  today. The branch should use the baseline's `top`, or force `0`.
- **The composer is not at `.app`'s bottom edge.** `.term-keybar`
  (`TerminalPane.tsx:874`) is gated on `isMobile && conn === 'connected'` only,
  not on `reading` — so ~41px (`styles.css:603-604`) of terminal control keys sit
  below the reader, for a terminal the user can't see, still wired to the PTY. I
  checked #49: it gates the covers and the paste sheet, not this. So the PR body's
  "the composer is pinned to the app's bottom edge" isn't true today, and there's
  a real strip there that a reporter could easily fold into "the gap".
- Adjacent, but you'll trip over it during the investigation: the terminal stays
  mounted and its `ResizeObserver → requestSize()` is live in reader mode
  (`TerminalPane.tsx:586`), so focusing the composer clamps `.app`, shrinks
  `.term-host`, and **resizes the agent's PTY**. Typing a reply in reader mode
  reflows the agent's TUI. Same 54%, different victim.

---

## The overlay: keep the idea, change the sampling

I'm fine with a debug surface in the product for this — no devtools, cross-origin
embed, and the failure is a layout you can only see. The justification holds. Two
problems with the implementation, both inline:

- **The instrument perturbs what it measures, in the frames that matter.** 3 ×
  `getBoundingClientRect()` forces a layout flush and `setState` triggers a
  reconcile, 60×/s, on the slowest device the app runs on, during the keyboard
  animation. And a live readout answers the wrong question: one screenshot
  catches one arbitrary frame of a 250-400ms animation, when the PR's stated
  purpose is to see which values *disagree mid-flight*. Sampling into a plain
  array, keeping the last N frames where anything changed, and painting via
  `ref.textContent` is both cheaper and strictly more informative — one
  screenshot then shows the timeline. `querySelector` per frame is not the cost;
  the rects are. (Confidence: high on the mechanism, unmeasured on the magnitude.)
- **`.vvdebug` is `position: fixed`, so it shares `.app`'s fate.** If the failure
  really is "WebKit displaced a fixed subtree", the instrument is displaced with
  it and leaves the screen precisely in the case it exists for. And the comment's
  reason for living outside `.app` isn't a reason: `getBoundingClientRect()` on an
  ancestor works fine, and a fixed child escapes `.app`'s `overflow: hidden`
  regardless — so the placement buys nothing and costs that failure mode.

Smaller: it's a static import, so it ships in the main chunk (`lazy()` behind
`VV_DEBUG` fixes that and makes the answer to "is there a debug surface in the
product" be "not unless you ask"); the `VV_DEBUG` const is wedged between two
import statements; and the readout is missing the two numbers that decide the
question — `focusBaseline.height` (what 54% multiplies) and `vv.width` (which
picks 0.48 vs 0.54). Stashing the baseline on `root.dataset` in `apply()` would
keep the overlay a pure reader.

---

## Checked and fine — no work needed

- **The geometry claim holds.** `.pane-reader` is a direct child of `.term-host`
  (`TerminalPane.tsx:863` → `:869`), and `.slot .term-host { position: relative }`
  (`conversation.css:200`) is what makes the overlay land. Both blockers are real
  as described: capture-phase non-passive `touchmove` at `:727` that
  `preventDefault`s and `stopImmediatePropagation`s, and `touch-action: none` at
  `styles.css:1104`.
- **Both guards are needed; neither is redundant.** touch-action alone would
  still let the JS handler scroll the hidden terminal on every touchmove; the JS
  guard alone would still leave the browser forbidden from panning. The PR does
  not overstate this.
- **Nothing else eats the gesture.** The only other capture listener is
  `pointerdown`/`paste` on `.term-fill` (`:406`) — a *sibling* of `.pane-reader`,
  and it doesn't `preventDefault`. xterm 5.5 registers no touch handlers (the
  existing comment's claim, and consistent with the bundled lib). Nothing in
  `ConversationView`/`Exchange` prevents a touch default. The listener
  add/remove option flags match, so no leak.
- **`overscroll-behavior: contain` is the right value.** `none` would kill the
  end-of-trace bounce, which is the only "you've reached the end" signal a phone
  gives, and buys nothing — `html, body, #root` are already
  `overscroll-behavior: none` in the same media query. `.cxv-body` is rendered
  only by the reader, so the un-scoped placement has no desktop fallout. (The
  stated *reason* overstates what it adds, but the property is correct.)
- **`modeRef` is reliable here.** Assigned during render and read from a
  once-registered listener, but `mode` is a plain prop off top-level state, the
  assignment is idempotent under StrictMode's double render, and the only window
  is render→commit — sub-frame, for a listener that can't fire in it. The pattern
  already exists at `:204-205`. Not new debt.
- **Mode flip mid-gesture, on the browser's side:** touch-action is latched at
  `touchstart`, so flipping *into* reader with a finger down leaves the browser
  still refusing to pan while the JS handler has gone inert — the drag does
  nothing until the finger lifts. Mildly odd, correct, not worth code.

## Two notes on the write-up

- "the trace could not be scrolled back at all" is true for a finger, but the
  ▲▼ turn-nav in `.cxv-bar` calls `scrollTo` and did work. That's the difference
  between "unusable" and "unusable the way anyone would actually try", and the
  commit message is the place it matters.
- The note for #49 is stale: it's already based on `main`.

Four of the findings above (the `pan-y`/`overflow-x` chain, the `--mono` typo,
the keybar rendering in reader mode, and #49's retargeting) were desk-checkable
in this environment and none needed a device. The touch fix resting on reading
rather than a repro is disclosed and fine.

web/src/styles.css
```diff
@@ -1094,6 +1102,11 @@ a.btn-ghost { text-decoration: none; }
     overflow: hidden;
   }
   .term-host { touch-action: none; overscroll-behavior: contain; }
+  /* touch-action is intersected down the ancestor chain, so `none` here would
+     also forbid panning the reader's scroller nested inside. Reader mode hands
+     the vertical gesture back to the browser; the terminal's drag handler is
+     inert in that mode anyway. */
+  .term-host.reading { touch-action: pan-y; }
```
thomwolf: **Must fix · high confidence.** `pan-y` re-creates the bug you just fixed, one axis over.

`touch-action` is intersected down the chain — the insight this PR is built on — so `pan-y` here also forbids the browser from panning anything nested inside *horizontally*. And the reader has horizontal scrollers:

- `styles.css:1032` — `.markdown pre { overflow-x: auto }`
- `styles.css:1036` — `.markdown table { display: block; overflow-x: auto }`

Both render inside `.cxv-body`, via `.markdown cx-md` at `Exchange.tsx:221`. So a trace containing a wide code block or a table — i.e. most agent traces — is unscrollable sideways on a phone, for exactly the reason the trace itself was.

`touch-action: auto` is what you want here: the terminal's handler is inert in this mode, so there is nothing left to protect and no reason to enumerate axes. If you specifically want to keep pinch/double-tap zoom off the reader, `pan-x pan-y` is the narrower version.

(`.cs-pre` is fine — `white-space: pre-wrap; word-break: break-word` means it never overflows horizontally, only vertically past its `max-height: 20rem`.)

web/src/components/TerminalPane.tsx
```diff
@@ -656,14 +656,21 @@ export default function TerminalPane({
       return rows;
     };
     const stopGlide = () => { if (glideFrame) { cancelAnimationFrame(glideFrame); glideFrame = 0; } };
+    // Reader mode covers this frame with its own scroller. These listeners sit
+    // on .term-host in CAPTURE and preventDefault, so without this guard every
+    // drag over the conversation was eaten here and spent on the hidden
+    // terminal's scrollback: the trace could not be scrolled back at all on a
+    // phone, which is the only place this gesture exists.
     const onTouchStart = (e: TouchEvent) => {
+      if (modeRef.current === 'reader') return;
```
thomwolf, line 672: **Must fix · high confidence it's reachable, medium that anyone notices.** Two of the four handlers are guarded; the two that manage the coast are not.

`onTouchEnd` (`:697`) is what launches the glide, and it has no mode check. `onTouchCancel` (`:725`) is unguarded too, and this early return also skips `stopGlide()`.

Two reachable states:

1. **Flick, then flip.** A flick in terminal mode leaves a glide running for up to ~1s. Flip the mode bar during it and the hidden terminal keeps coasting under the reader — and because `onTouchStart` now returns before `stopGlide()`, no touch in reader mode can stop it. `residual` keeps accumulating too.
2. **Flip mid-drag.** `velocity` is only zeroed in `onTouchStart`/`onTouchEnd`/`onTouchCancel`. Flip to reader with a finger down and `onTouchMove` stops updating it but leaves it at its last sample, so the unguarded `onTouchEnd` fires a glide on a terminal nobody is looking at.

Neither needs a contrived user: the mode is app-wide and broadcast over `am:pane-mode` (`lib/paneMode.ts`), so it can change from another surface entirely while a finger is down. The symptom is mild — "my scrollback moved while I was reading" — but that is the kind of thing that gets filed later as a separate mystery.

Two lines: add the same guard to `onTouchEnd` before it reads `velocity`, and call `stopGlide()` in the effect at `:812` that already blurs the terminal when `reading` goes true. That effect is the natural home for "the terminal is no longer being driven".

web/src/components/TerminalPane.tsx
```diff
@@ -656,14 +656,21 @@ export default function TerminalPane({
       return rows;
     };
     const stopGlide = () => { if (glideFrame) { cancelAnimationFrame(glideFrame); glideFrame = 0; } };
+    // Reader mode covers this frame with its own scroller. These listeners sit
```
thomwolf, line 666: **Must fix · nit, but this codebase cares.** This new comment is accurate; the two above it are now wrong, and they are the ones a future reader hits first.

`:627` — "*`.term-host` sets touch-action:none on mobile, so neither xterm nor the browser will pan: this is the only touch scrolling a phone has, in every pane*" — and `:636` — "*`.term-host` is what carries touch-action:none*". Both state unconditionally the thing this PR just made conditional, and it is the exact premise the guard six lines below depends on.

Given how much this file invests in comments that explain *why*, having the old explanation contradict the new one thirty lines apart is the expensive kind of drift. "in every pane" → "in every pane the terminal is on top of".

web/src/styles.css
```diff
@@ -523,6 +523,14 @@ body {
 /* transient error toast */
 .toast { position: fixed; bottom: 16px; left: 50%; transform: translateX(-50%); z-index: 300; max-width: 90vw; padding: 9px 14px; background: color-mix(in srgb, var(--danger) 92%, #000); color: #fff; border-radius: var(--r-md); font-size: 12.5px; box-shadow: 0 8px 28px rgb(0 0 0 / 0.3); animation: rise-in 0.16s ease-out; }
 
+/* ?vvdebug=1 — the mobile viewport readout. Above everything, transparent to
+   touch so it never becomes the reason a gesture misbehaves. */
+.vvdebug {
+  position: fixed; top: 0; left: 0; z-index: 999; margin: 0; padding: 4px 6px;
+  pointer-events: none; font-family: var(--mono, monospace); font-size: 9px; line-height: 1.35;
```
thomwolf: **Nit · high confidence, and one grep here would have caught it.** `--mono` is not a token in this codebase; it is `--font-mono` (`styles.css:6`). This falls through silently to plain `monospace`, so the readout renders in the platform default and nothing tells you.

`:1175` has the identical typo, so this is consistent-with-an-existing-bug rather than novel — but the fix is `var(--font-mono, monospace)`, or just add `mono` to the class list and let `.mono` do it.

web/src/components/ViewportDebug.tsx
```diff
@@ -0,0 +1,49 @@
...
+      const rect = (el: Element | null) => {
+        if (!el) return '–';
+        const r = el.getBoundingClientRect();
+        return `${Math.round(r.top)}→${Math.round(r.bottom)}`;
+      };
+      const active = document.activeElement;
+      const name = active instanceof HTMLElement
+        ? `${active.tagName.toLowerCase()}.${(active.className || '').split(/\s+/)[0] || '-'}`
+        : '–';
+      setLines([
```
thomwolf: **Consider · the biggest thing about this file.** An unbounded rAF is defensible for a diagnostic; a `setState` plus three forced layouts per frame is what I would push back on — not on principle, but because it degrades the specific measurement this exists to take.

Per frame, forever, on the slowest device the app runs on: three `getBoundingClientRect()` calls, each a forced layout flush — and `apply()` is mutating `.app`'s inline size on timers at 0/50/150/300/500/800ms, so there is always dirty layout to flush — plus a fresh 9-element array into `setState`, which never compares equal, so React reconciles every frame. The interesting window is the ~250-400ms keyboard animation. That is exactly where this is heaviest, and where a dropped frame changes which values you observe disagreeing.

`querySelector` is not the cost here; the rects are. Don't bother caching the elements — they mount and unmount.

There is a design that is both cheaper and strictly more informative. The description says the point is to catch "*the values that disagree mid-flight*" — but a live-updating readout cannot deliver that: one screenshot catches one arbitrary frame of the animation and you cannot tell whether you caught the interesting one. Instead:

- sample into a plain array in the rAF (no state, no reconcile);
- keep the last N frames in which any value *changed*;
- paint with `ref.current.textContent = …` — no React in the loop at all.

Then one screenshot from the phone shows the timeline rather than an instant, which is what actually separates "the 54% clamp overshoots" from "scroll-to-reveal moved a fixed subtree". Same instrument, less perturbation, more answer.

web/src/components/ViewportDebug.tsx
```diff
@@ -0,0 +1,49 @@
...
+        `screen ${window.screen.height}  inner ${window.innerHeight}  embed ${window.self !== window.top ? 'y' : 'n'}`,
+        `vv     h ${Math.round(vv?.height ?? -1)}  top ${Math.round(vv?.offsetTop ?? -1)}  scale ${(vv?.scale ?? 1).toFixed(2)}`,
```
thomwolf: **Consider.** The readout is missing the two numbers that actually decide the question it was built for:

- **`focusBaseline.height`** — the value `0.54` is multiplied by (`App.tsx:187`). `vv h` here is the *live* reading, which is the same number only if the premise holds that a child frame's visual viewport does not move for the keyboard. If that premise is wrong you can no longer recover what the clamp was computed from — and testing that premise is the whole reason this file exists.
- **`vv.width`** — `App.tsx:186` picks `0.48` vs `0.54` from `focusBaseline.width > focusBaseline.height`. `mode` says `focus-fallback` but not which ratio won, so a landscape reading is ambiguous.

Cheapest way to get both without coupling this component to App's internals: have `apply()` stash the baseline on `root.dataset` (it already writes `keyboardLayout` there) and keep this file a pure reader of `documentElement`. That is a property this component is nicer for having.

web/src/App.tsx
```diff
@@ -823,6 +828,9 @@ export default function App() {
         onToggleDemo={toggleDemo}
       />
       )}
+    {/* Outside .app on purpose: it reports where .app was put, so it must not
+        be inside the box it is measuring. */}
+    {VV_DEBUG && <ViewportDebug />}
```
thomwolf: **Consider · the placement argument does not hold, and the placement it chose has a failure mode.**

`getBoundingClientRect()` on an ancestor works fine, and a `position: fixed` child escapes `.app`'s `overflow: hidden` regardless — so being outside `.app` in the DOM buys nothing measurable. "It must not be inside the box it is measuring" is intuitive but is not a real constraint here.

What it does cost: `.vvdebug` is itself `position: fixed` (`styles.css:529`), so it is anchored to the same layout viewport as `.app` and shares its fate. If the bug turns out to be the second hypothesis in your description — WebKit displacing a `position: fixed` subtree on focus — the instrument is displaced with it and leaves the screen in precisely the case it was written for. You would get a blank band and no readout, which reads like "the overlay didn't work".

There is no clean fix from inside the app (`html, body, #root` are `overflow: hidden`, so `position: absolute` is no better), and that is itself worth knowing: it means this readout can only be trusted for the *clamp* hypothesis, and the `app` rect line is doing double duty as the tell for the other one. Better to say that in the comment than to claim a constraint that isn't one.

Separately: this is a static import, so `ViewportDebug` ships in the production main chunk. `const ViewportDebug = lazy(() => import('./components/ViewportDebug'))` behind `VV_DEBUG` keeps it out, and makes the answer to "is there a debug surface in the shipped product" be "not unless you ask for one" — a better answer to the question you raise at the end of the description.

web/src/App.tsx
```diff
@@ -13,6 +13,11 @@ import Overview from './components/Overview';
 import Locked from './components/Locked';
 import BackupBanner from './components/BackupBanner';
 import Welcome from './components/Welcome';
+import ViewportDebug from './components/ViewportDebug';
+
+// `?vvdebug=1` — a phone has no devtools, and the keyboard layout is guessed
+// when the app is embedded cross-origin. Read once: this never changes mid-run.
+const VV_DEBUG = new URLSearchParams(location.search).has('vvdebug');
```
thomwolf: **Nit.** This wedges a statement between two `import` declarations — `import ViewportDebug` above, `import * as api` below. Legal (imports hoist) but it splits the import block in a file where every other module reads top to bottom, and any import-ordering lint will eventually trip on it. Move it under the last import.

The comment itself is good: "read once: this never changes mid-run" is exactly the *why* for a module-level const rather than a hook.

web/src/conversation.css
```diff
@@ -163,7 +163,7 @@ mark.cx-hit.on { background: var(--accent); color: var(--panel); }
    you scan for — still spanned the full width, so the two disagreed about where
    the conversation began. The pane is the measure: make it narrow and the
    conversation is narrow. */
-.cxv-body { flex: 1; min-height: 0; overflow-y: auto; background: var(--panel); padding: 4px 14px 30px; }
+.cxv-body { flex: 1; min-height: 0; overflow-y: auto; overscroll-behavior: contain; background: var(--panel); padding: 4px 14px 30px; }
```
thomwolf, line 166: **Fine as-is** — `contain` is the right value, not `none`. `none` would additionally suppress the end-of-trace bounce, which on a phone is the only signal that you have reached the end of what was loaded, and it buys nothing `contain` doesn't already give.

One correction to the rationale in the description ("*so reaching the end of a trace does not start dragging whatever is behind the app*"): that was already true — `html, body, #root { overflow: hidden; overscroll-behavior: none }` sits in the same `max-width: 720px` block (`styles.css:1097`). What this line actually adds is stopping the chain from reaching `.term-host`, which matters more now that `.term-host.reading` no longer says `touch-action: none`. Same change, more accurate reason.

Also checked: `.cxv-body` is rendered only by `ConversationView`, which only `TerminalPane` mounts, so putting this in the base rule rather than the mobile block has no desktop fallout.

web/src/components/TerminalPane.tsx
```diff
@@ -656,14 +656,21 @@ export default function TerminalPane({
...
     const onTouchStart = (e: TouchEvent) => {
+      if (modeRef.current === 'reader') return;
       stopGlide();               // a new touch takes over from any coasting
       velocity = 0;
       residual = 0;
       touchY = e.touches[0].clientY;
       lastMoveAt = e.timeStamp;
     };
     const onTouchMove = (e: TouchEvent) => {
+      if (modeRef.current === 'reader') return;
```
thomwolf, line 680: **Fine — and worth recording that it is not redundant with the CSS.** I went looking for whether one of the two mechanisms was enough on its own. Neither is:

- `touch-action: pan-y` alone: `touchmove` still fires, so this handler would still call `scrollByPixels` on every event and the reader would pan *while* the hidden terminal scrolled.
- This guard alone: `touch-action: none` would still forbid the browser from panning `.cxv-body` at all.

Also checked `modeRef`, since it is assigned during render and read from a listener registered once: fine here. `mode` is a plain prop off top-level state, the assignment is idempotent under StrictMode's double render, and the only window where the ref leads the DOM is render→commit — sub-frame, and no touch event can land in it. Same pattern already at `:204-205`.

And the mid-gesture flip, from the browser's side: `touch-action` is latched at `touchstart`, so flipping *into* reader with a finger already down leaves the browser still refusing to pan while this handler has gone inert — the drag does nothing until the finger lifts. Slightly odd, correct, not worth code.

thomwolf:
Thanks — this caught a real regression and a real leak. All four must-fixes are in `699b938`.

**1. `pan-y` → `auto`.** You're right, and it's the same bug one axis over: `.markdown pre` and `.markdown table` are `overflow-x: auto` (styles.css:1032, :1036) and a trace is full of wide output. Fixed, and the comment now says why the axis is not pinned, so the next person does not "tighten" it back.

**2. The glide.** Both paths were real. `onTouchEnd` now bails in reader mode, and the `[reading, focused]` effect stops a glide already in flight via a `stopGlideRef` — your point that the flip is broadcast over `am:pane-mode`, so it need not be the same hand, is what makes it worth more than a theoretical guard. I left `onTouchCancel` unguarded on purpose: it only clears `touchY`/`velocity` and stops the glide, so it is the one handler that should still run in reader mode.

**3. Stale comments.** Fixed both. They asserted the exact premise the new guard depends on being false.

**4. `var(--mono)`.** Real leak, `--font-mono` it is. No excuse — the grep was one line.

**5–7. The readout.** Took all of it, and the history point improved the instrument rather than just cheapening it: a live line cannot show values disagreeing mid-flight, because by the time a phone can be screenshotted the animation is over. It now keeps the last ten *changed* samples, so the tail of the list is the transition; writes through `textContent` on the frame loop instead of setState, so measuring the layout animation stops perturbing it; adds the unfocused `vv.height` baseline and `vv.width`; and is `lazy()`, confirmed splitting to its own `ViewportDebug-*.js` chunk instead of riding in `index-*.js`. On the "outside `.app`" justification — you're right that fixed shares `.app`'s fate, so the comment no longer claims immunity, only that the history still records the displacement.

**8.** Moved below the imports.

**9. `overscroll-behavior: contain`.** Agreed and kept; also agreed the original commit message oversold what it adds.

**On the keyboard bug — your hypothesis closes the arithmetic and mine did not.** `focusBaseline.height` being the *iframe's* height rather than the screen's, plus the Hub page doing its own scroll-to-reveal and dragging the abandoned strip into view — that produces a keyboard-sized band of `body`'s `--bg` against the reader's `--panel`, which is exactly what was reported. That is a better explanation than "the 54% constant is slightly off".

And the free discriminator is the most useful thing in this review: `acceptsKeyboardInput` does not care *which* field has focus, so the clamp fires for the reader's search box at the top of the pane too. Band on the search box as well ⇒ the clamp. Band only on the bottom composer ⇒ scroll-to-reveal. No instrument required, and it is being tested on the reporter's phone now. I am deliberately not shipping a fix for this until that comes back — the `apply()` mixing a live `top` with a baseline-derived `height` is the shape a double-subtraction takes, and I'd rather know which one it is than patch both.

Two things you raised that I am leaving as follow-ups rather than folding in here, so this PR stays about the reported bugs — flag if you disagree:

- **`.term-keybar` renders in reader mode**, so the composer is *not* at `.app`'s bottom edge. You're right that the PR body says otherwise; that sentence is wrong. The bar is esc/tab/arrows/`^C` for a TUI and has no business over a reader, but removing it is a UX change, and #49 owns that surface.
- **Focusing the composer resizes the agent's PTY**, via the mounted terminal's ResizeObserver. That is a genuine "reading a conversation perturbs the agent" bug and it deserves its own issue, not a hunk here.

Also: the #49 note in the description is stale, as you say — it has been retargeted to `main` already.

thomwolf:
Owner's call, and it collapses the open question rather than tuning it: **the app no longer estimates a keyboard height at all** (`07ad947`).

When there is no geometry to read — a cross-origin frame on mobile Safari, which is how the Hub page serves a Space — `apply()` now reports the viewport it can see and stops. Where real geometry exists (a direct browser's `visualViewport`, Chromium's `virtualKeyboard`) nothing changes.

The argument for it, beyond "one less constant": a guess in the safe direction is still a guess, and the review established that this one's abandoned strip does not stay hidden behind the keyboard — the browser scroll-reveals a focused field and drags it back into view, painting `body`'s `--bg` against the reader's `--panel` at a size set by a constant instead of by the keyboard. And the job the clamp was doing is already done by the engine that can actually see the keyboard: the same scroll-into-view that keeps xterm's pinned 1px helper textarea above it today.

Out: the 0.54/0.48 ratio, `focusFallbackTimer`, `scheduleEmbeddedFallback`, `embeddedTouchLayout`, the `focus-fallback` layout mode, and a `focusedInput` that had become write-only. 58 lines out, 24 in. `focusBaseline` stays — a keyboard is still *detected* as the shrink from it, which is a measurement, not an estimate.

This also makes the discriminator moot by running the experiment: with no clamp left to cause a band, a band that survives means WebKit is displacing a fixed subtree on scroll-to-reveal — a different fix in a different place, and now unambiguous. `?vvdebug=1` stays for exactly that case.

What to watch for on a device, since this trades a guess for the browser's judgement: if the parent page declines to scroll, the composer sits *under* the keyboard instead of above a gap. That is the failure mode this replaces the old one with, and it is the thing to check first — in reader mode and in a terminal pane both.

tsc clean, `exchanges: ok`, production build succeeds.

thomwolf (commented):
**Reviewing `07ad947` only** ("Mobile keyboard: stop estimating a height nobody can
measure"). `fb47763`/`6e42244`/`699b938` were reviewed above and I have not re-litigated
them.

Short version: **the deletion is right, and it is clean but for one comment.** I checked
the "strictly a no-op outside the iframe" claim against the pre-image rather than the
message, and it holds. The three findings below are all small; the section after them is
the thing I'd actually spend the review on, because the commit bets on a mechanism I was
able to measure here, and the measurement is not quite what the comment says.

---

## Must fix

**1. One orphaned comment survived the deletion — in the instrument.**
`ViewportDebug.tsx:27-29` still reads "what App.tsx's `focusFallback` later multiplies by
its constant … the discrepancy that makes the guess wrong." There is no `focusFallback`
and no constant. The *variable* is still worth keeping — `baseline` is the shrink
reference the readout's `keyboard-layout` column now turns on — but the reason it exists
changed completely in this commit. Confidence: high.

That is the **only** survivor. I grepped the whole `web/` tree for `focus-fallback`,
`focusFallback`, `embeddedTouchLayout`, `focusedInput`, `scheduleEmbeddedFallback`,
`visibleRatio`, `0.54`/`0.48` and `window.self !== window.top`: nothing else. In
particular **no CSS selector keys off `[data-keyboard-layout="focus-fallback"]` — nothing
keys off the attribute at all**, in any stylesheet; the sole reader in the codebase is
`ViewportDebug.tsx:60`. So there is no orphaned styling to clean up, which is worth
stating because it is the first thing a reader of this commit will wonder about.

---

## Consider

**2. `focusBaseline` goes stale on an input→input focus move, and that now corrupts the
only output the machinery has left.** `onFocusIn` (`App.tsx:225`) re-captures
unconditionally. Focus the composer while a keyboard is already up — from the search box,
or from the terminal, where reader mode blurs xterm (`TerminalPane.tsx:827`) but the OSK
does not reliably drop between that blur and the next tap — and the baseline is captured
*post*-shrink. `visualShrink` and `layoutShrink` are then ~0, `hasKeyboardGeometry()`
returns false, and `dataset.keyboardLayout` is deleted while a keyboard is demonstrably
up.

Today the only consequence is that `?vvdebug=1` prints `none`. But that readout is
precisely the instrument shipped to answer "did the browser report geometry or not", and
tapping from one field to another is exactly the sequence the previous review's
discriminator asks the reporter to perform. An instrument that mislabels the case it was
built for is worth two words:

```js
if (acceptsKeyboardInput(target) && !focusBaseline) focusBaseline = captureViewport();
```

Safe, because `focusout` fires before `focusin` on a move and its 0ms timer only clears
the baseline when focus has left *every* accepting element — so an input→input hop keeps
the genuine pre-keyboard baseline, and a real defocus→refocus still re-captures. Leave
`onOrientationChange` re-capturing unconditionally; a baseline from the old orientation is
meaningless. Confidence: high on the mechanism, low on user impact, high on instrument
impact. The shape predates this commit — but this commit makes `dataset.keyboardLayout`
the *sole* product of that bookkeeping, so a wrong value is now the whole signal.

**3. Say that the retained machinery is diagnostic, or the next person will misjudge it in
one direction or the other.** After this commit `focusBaseline`, `captureViewport()` and
`hasKeyboardGeometry()` **do not affect layout at all** — their only consumer is
`root.dataset.keyboardLayout` (`:190-191`), which nothing styles and only the lazy debug
overlay reads. The `keyboardRect` clamp above at `:176-179` is independent of
`hasKeyboardGeometry()`. That is a fine thing to keep — it is the discriminator for the
open investigation — but "`focusBaseline` stays — a keyboard is still detected as the
SHRINK from it" reads as though detection still steers something, and the new comment at
`:139-141` says it is "the only thing left that needs remembering" without naming what
needs it. One clause fixes both. Confidence: high.

**4. The commit message names Chromium's `virtualKeyboard` as one of the two places "real
geometry exists", and as wired it is not one.** `navigator.virtualKeyboard.overlaysContent`
is never set anywhere in `web/` (grepped). Per the API, `boundingRect` is a zero `DOMRect`
and `geometrychange` does not fire unless `overlaysContent = true`; with the default
`false` the browser resizes the visual viewport instead, which the `vv` path already
handles. So the clamp at `:176-179`, the `geometrychange` listener at `:246` and
`keyboardHeight` in `hasKeyboardGeometry()` are all inert today. **This commit does not
make it worse and I would not fix it here** — but it is one of the two survivors the
message leans on, so either opt in (a real behaviour change: content then sits under the
keyboard and you own the inset) or note in the comment that the branch is for a mode the
app does not currently enable. Confidence: high on the grep, high on the spec.

**5. The PR body still describes the deleted code as present** — "**No fix here yet, on
purpose**", and it quotes `visibleRatio` as current. Your follow-up comment explains
`07ad947` well, but the body is what a merger reads.

---

## The bet, measured — and I could measure more of it than "needs a phone" implies

The commit trades the clamp for "the browser already scrolls a focused field into view."
The keyboard half of that is genuinely untestable here. The **structural** half needs no
keyboard, and it is the half that decides whether the bet can pay: *is there anything in
the app's own layout for a scroll-into-view to move?*

`playwright@1.62.1` is a devDependency (`web/package.json:57`) and there is a working
Chromium at `/opt/pw-browsers/chromium-1234/`. I extracted the mobile chain — `html, body,
#root { overflow: hidden }`, `.app { position: fixed; height: var(--vvh) }`, `.term-host`,
`.pane-reader { position: absolute; inset: 0 }`, `.cxv` column, `.cxv-body`, `.cxv-live`,
`.cxv-foot`, `.term-keybar` — into a standalone page and measured it at 390x844 in
standards mode:

```
                         scroll range   scrollTop
  html / body / #root         0             0
  .app                        0             0
  .main / .term-host          0             0
  .pane-reader / .cxv         0             0
  .cxv-body                1399             0     <- the only scroller
```

Unchanged after `composer.focus()`, and unchanged after an explicit
`scrollIntoView({block:'center'})` on the composer — and the same for the search box.

**So: every scroll container between the composer and the document has zero scroll range,
and the one container that can scroll (`.cxv-body`) does not contain the composer — it is a
sibling above it.** A scroll-into-view inside this document is a provable no-op. The reveal
is therefore delegated *entirely* to the cross-origin parent, with nothing behind it. This
does not mean the commit is wrong — WebKit does propagate a focused element's rect across a
frame boundary, and that is how any form in an embedded Space works on iOS today — but it
means there is no second line of defence, which is worth knowing before merge.

Two concrete things fall out that I'd add to the PR body:

- **The required travel is keyboard + 75px, not keyboard.** Measured at 390x844: composer
  bottom 769, then `.cxv-foot` 775-797 (22px), then `.term-keybar` 797-836 (39px). The
  composer's bottom edge sits 75px above the frame's. Under the old clamp the composer was
  at ~381 with room to spare; now it is as deep under the keyboard as the layout allows, so
  this is the commit that makes the `.term-keybar`-in-reader-mode follow-up (#49) cost
  something rather than merely look wrong.
- **The search-box experiment got sharper, not moot.** The search box sits at top=38 and
  needs no reveal at all. With the clamp gone, a band there now means neither the clamp nor
  scroll-to-reveal — it is the parent page, which is a third answer the previous
  discriminator could not produce.

And the failure mode you named in the thread is the right one to check first. I'd add its
*success* mode to the watch list too, because a reporter will file it as a new bug: when
the parent does pan, the app's header and mode bar go off the top of the screen. Nothing
shrinks any more, so revealing the bottom necessarily costs the top — in terminal mode
that is the whole pane header for a prompt line.

---

## Checked and fine — no work needed

- **"Strictly a no-op outside the iframe" holds, verified against the pre-image.**
  `focusFallback` could only become true inside `scheduleEmbeddedFallback`'s 500ms timer,
  which early-returned unless `embeddedTouchLayout()` — and that tested
  `window.self !== window.top` *before* it looked at width or pointer type. Off the Hub,
  `focusFallback` was permanently false and `apply()` took the `else if
  (hasKeyboardGeometry())` arm, which is what the new two-line form does verbatim. Desktop
  fails all three conditions besides. Direct mobile Safari and Chromium are untouched.
- **`dataset.keyboardLayout` is set or cleared on every path.** Every listener funnels into
  `apply()`, which now always takes one of the two branches, and the effect's cleanup still
  deletes it (`:266`). No path can leave a stale value.
- **The remaining `focusBaseline` lifecycle is correct** apart from finding 2: cleared on
  `focusout`'s 0ms timer only when focus truly left every accepting element, which is the
  right moment and the right test.
- **No leaked timers.** `focusFallbackTimer`'s clear was removed from the cleanup together
  with the timer itself, both remaining `Set`s are still drained, and `stabilizeFocus()` is
  still called on every path it used to be — including the early-return branch that used to
  duplicate it.
- **A pleasant side effect worth recording on the follow-up issue:** deleting the clamp
  also fixes "focusing the composer resizes the agent's PTY" **for the embedded case** —
  nothing shrinks, so `.term-host` does not resize and the ResizeObserver never fires. It
  still happens on a direct mobile browser, where `--vvh` genuinely shrinks.
- **Comment quality.** The header comment at `:118-126` and the inline one at `:180-189`
  overlap fairly heavily — both explain "we do not estimate, the browser reveals". Not
  worth a round trip on its own, but if you touch `:180-189` for finding 3 anyway, the
  inline one is the one that can shrink.

## On the stated verification

`tsc`, `exchanges`, and a production build are the right checks for a pure deletion and I
have no reason to doubt them. The one thing that was checkable and was not checked is the
structural probe above — no keyboard, no device, and the tooling is already in
`devDependencies`. To be plain about the limit: headless Chromium **cannot** emulate an
on-screen keyboard, cannot shrink a visual viewport independently of the layout viewport,
and cannot reproduce mobile Safari's cross-origin visual-viewport behaviour, so the actual
keyboard question is not answerable here by any means I can find. "Needs a phone" is true —
it is just narrower than the whole change.

*(Also true and not held against it: the caveat in my probe is that it is a faithful
extraction of the CSS chain, not the built app — building would have written into the
worktree.)*

web/src/components/ViewportDebug.tsx
```diff
@@ -0,0 +1,84 @@
...
+export default function ViewportDebug() {
+  const out = useRef<HTMLPreElement | null>(null);
+
+  useEffect(() => {
+    let frame = 0;
+    let last = '';
+    const history: string[] = [];
+    // vv.height while nothing is focused: what App.tsx's focusFallback later
+    // multiplies by its constant. In an iframe this is the FRAME's height, not
+    // the screen's — the discrepancy that makes the guess wrong.
```
thomwolf: **Must fix — the one leftover the deletion missed** (confidence: high).

`focusFallback` and its constant no longer exist as of `07ad947`; this comment is the only
reference to them left anywhere in `web/` (grepped `focus-fallback`, `focusFallback`,
`embeddedTouchLayout`, `focusedInput`, `scheduleEmbeddedFallback`, `visibleRatio`,
`0.54`/`0.48` — everything else is clean, and no CSS selector keys off
`[data-keyboard-layout]` at all).

The variable earns its keep, so this is a rewrite and not a deletion: `baseline` is still
the pre-focus reference that `hasKeyboardGeometry()`'s shrink test turns on, which is now
what decides the `keyboard-layout` column on line 60. What changed is that nothing
multiplies it any more — it is a comparand, not a multiplicand. The iframe-vs-screen
observation in the last two lines was the reason the *guess* was wrong and no longer
explains anything; what is worth saying in its place is that in an iframe this number is
expected to stay flat when the keyboard opens, because that flatness IS the no-signal case
the readout exists to confirm.

web/src/App.tsx
```diff
@@ -217,54 +220,24 @@ export default function App() {
...
       const target = event.target instanceof Element ? event.target : null;
-      if (acceptsKeyboardInput(target)) {
-        focusedInput = target;
-        focusBaseline = captureViewport();
-        focusFallback = false;
-        stabilizeFocus();
-        scheduleEmbeddedFallback();
-        return;
-      }
+      if (acceptsKeyboardInput(target)) focusBaseline = captureViewport();
```
thomwolf: **Consider — stale baseline on an input→input focus move** (confidence: high on the
mechanism, low on user impact, high on instrument impact).

This re-captures unconditionally, so the baseline is only a *pre-keyboard* viewport if the
keyboard was down when the field took focus. Two reachable ways it is not:

- reader mode: tap the search box, then tap the composer — `focusin` fires again with the
  keyboard already up;
- terminal → reader: the mode effect blurs xterm (`TerminalPane.tsx:827`), but the OSK does
  not reliably drop between that blur and the next tap on the composer.

In both, the baseline is captured *post*-shrink, so `visualShrink` and `layoutShrink` are
~0 and `hasKeyboardGeometry()` returns false with a keyboard demonstrably up — and
`apply()` then deletes `dataset.keyboardLayout`.

The only consequence today is that `?vvdebug=1` prints `none`. That is exactly why it is
worth fixing rather than noting: that readout is the instrument shipped to answer "did the
browser report geometry or not", and field-to-field tapping is the sequence the previous
review's discriminator asks the reporter to perform. Two words:

```js
if (acceptsKeyboardInput(target) && !focusBaseline) focusBaseline = captureViewport();
```

Safe because `focusout` fires before `focusin` on a move and its 0ms timer only clears the
baseline when focus has left *every* accepting element — so a hop keeps the genuine
pre-keyboard baseline while a real defocus→refocus still re-captures. `onOrientationChange`
should keep re-capturing unconditionally; a baseline from the old orientation is
meaningless.

The shape predates `07ad947`. I am raising it here because this commit makes
`dataset.keyboardLayout` the sole product of this bookkeeping, so a wrong value is now the
whole signal rather than one input among several.

web/src/App.tsx
```diff
@@ -127,11 +136,10 @@ export default function App() {
     };
     const root = document.documentElement;
     const keyboardSignalThreshold = 80;
-    const embedded = window.self !== window.top;
-    let focusedInput: Element | null = null;
+    // The viewport as it was before a field took focus — the only thing left
+    // that needs remembering, because a keyboard is detected as the SHRINK from
+    // it (hasKeyboardGeometry), not as an absolute height.
```
thomwolf: **Consider — name the consumer** (confidence: high).

"The only thing left that needs remembering" is accurate but stops one clause short of the
thing a reader needs, which is *what* needs it. After `07ad947`, `focusBaseline`,
`captureViewport()` and `hasKeyboardGeometry()` have **no effect on layout at all**: their
sole consumer is `root.dataset.keyboardLayout` at `:190-191`, nothing in any stylesheet
matches that attribute, and the only reader in the codebase is `ViewportDebug.tsx:60`. The
`keyboardRect` clamp at `:176-179` reaches `keyboard.boundingRect` directly and does not
call `hasKeyboardGeometry()`.

So this is now diagnostic machinery, and that is a defensible thing to keep — it is the
discriminator for the open investigation, and the commit is explicitly an experiment. But
left unsaid it fails in both directions: the next person either treats ~30 lines of
lifecycle bookkeeping as load-bearing and works around it, or notices nothing consumes it
and deletes the instrument's only signal. "…and the only consumer is the `keyboard-layout`
marker `?vvdebug=1` prints" closes it.

Same note for the commit message: "`focusBaseline` stays — a keyboard is still detected as
the SHRINK from it" is true, but reads as though the detection still steers the layout.

web/src/App.tsx
```diff
@@ -172,20 +177,18 @@ export default function App() {
...
+      // When the browser reports no keyboard geometry — a cross-origin frame on
+      // mobile Safari, which the Hub page is — the app does NOT invent a height.
+      // It used to assume a keyboard ate 46% and shrink to fit, and a guess that
+      // is wrong in the safe direction is still wrong: the abandoned strip does
+      // not stay hidden behind the keyboard, because the browser scroll-reveals
+      // a focused field and drags it back into view. That strip is the blank
+      // band under the reader's composer. Leaving the viewport alone hands the
+      // job to the engine that can actually see the keyboard — the same
+      // scroll-into-view that already keeps xterm's pinned helper textarea
+      // above it.
```
thomwolf: **Consider — the last three lines name a mechanism that is a no-op in this document**
(confidence: high; measured, not reasoned).

The premise is right and I would keep the change. But "hands the job to the engine that can
actually see the keyboard — the same scroll-into-view that already keeps xterm's pinned
helper textarea above it" quietly covers two different mechanisms, and only one of them is
scroll-into-view:

- **direct mobile browser:** `vv.height` shrinks → `--vvh` shrinks → `.app` shrinks and the
  focused field comes up with it. No reveal required.
- **cross-origin frame (the case this comment is about):** nothing shrinks, and the child
  document has nothing to scroll.

I measured the second one. `playwright@1.62.1` is already in `devDependencies` and there is
a Chromium at `/opt/pw-browsers/`; I extracted the mobile chain (`html, body, #root
{overflow:hidden}` → `.app {position:fixed; height:var(--vvh)}` → `.term-host` →
`.pane-reader {position:absolute; inset:0}` → `.cxv` column) into a standalone page at
390x844, standards mode:

```
                         scroll range   scrollTop
  html / body / #root         0             0
  .app                        0             0
  .main / .term-host          0             0
  .pane-reader / .cxv         0             0
  .cxv-body                1399             0     <- the only scroller
```

Unchanged after `composer.focus()`; unchanged after an explicit
`scrollIntoView({block:"center"})`; same for the search box. Every scroll container between
the composer and the document has **zero** range, and the one container that can scroll does
not contain the composer — `.cxv-body` is a sibling above it.

So in the embedded case the reveal is delegated *entirely* to the parent process, with
nothing behind it. That is still very likely to work — WebKit propagates a focused
element's rect across a frame boundary, which is how any form in an embedded Space works on
iOS — but the comment reads as if the local engine will do something, and locally it
provably cannot. Worth one line, because it is also the thing to check first on the device.

Two numbers from the same run that belong in the PR body rather than here:

- the required travel is keyboard **+ 75px**, not keyboard: composer bottom 769, then
  `.cxv-foot` 775-797 and `.term-keybar` 797-836. Under the old clamp the composer sat at
  ~381; it is now as deep under the keyboard as the layout allows, which is what turns the
  `.term-keybar`-in-reader-mode follow-up (#49) from cosmetic into load-bearing.
- the search-box experiment got *sharper*, not moot: it sits at top=38 and needs no reveal,
  so a band there now implicates neither the clamp nor scroll-to-reveal but the parent page
  — a third answer the earlier discriminator could not produce.

thomwolf:
Addressed in `8da137f`. The measurement is the most useful thing anyone has contributed to this bug — thank you for going and taking it rather than reasoning about it.

**Must fix — the orphaned comment.** Gone. It described multiplying by a constant that no longer exists.

**The baseline re-capture.** Taken, and you were right that it matters more than its blast radius suggests: the tap sequence that breaks it (field → field with the keyboard already up) is *exactly* the search-box-then-composer sequence the discriminator asks for, so the instrument would have lied on the one run it exists for. `focusin` now captures only when there is no baseline; `focusout` already clears it when focus genuinely leaves.

**"Names the mechanism that is a no-op in the child document."** Correct, and the comment now says the opposite of what it did: measured at 390×844, every box in this document has a scroll range of exactly 0, so a scroll-into-view in here moves nothing and the reveal is wholly the parent page's. What this document can contribute is *not fighting it* by resizing against a keyboard it cannot see. That is a better statement of why the change is right than the one I wrote.

**`focusBaseline` sizes nothing now.** Said in both places — at the declaration and at the one place it is consumed — so it is not mistaken for load-bearing or for dead.

**Chromium's `virtualKeyboard` is inert as wired** (`overlaysContent` is never set, so no `boundingRect` and no `geometrychange`). You are right, and right that it predates this branch. The commit message should not have listed it as a surviving path; a follow-up either sets `overlaysContent` deliberately or drops the branch. Not smuggling it in here.

**The PR body quoted deleted code as current** — fixed.

**On the +75px.** This is the part I want to flag rather than quietly act on: the composer is not the bottom of the pane. `.cxv-foot` (22px) and `.term-keybar` (39px) sit below it, so the parent must pan keyboard + 75px, not keyboard. That turns the "key-bar renders over the reader" follow-up from cosmetic into the difference between a visible composer and a hidden one — esc/tab/arrows/`^C` are TUI controls with no meaning over a reader, and gating them behind `!reading` is one condition. I have left it out because it changes what the pane looks like and #49 owns that surface, but it is now the first thing to do if the device test shows the composer landing under the keyboard.

**On what was checkable and wasn't:** fair hit. The structural half was measurable here and I did not measure it; you did. What remains genuinely un-checkable in this environment is the OSK and the cross-origin iOS visual viewport, which is the half the change bets on.

thomwolf:
**Confirmed on a device.** The branch has been running on a private Space and exercised from iOS Safari through the Hub page — the surface none of this could be tested against here. Reported working at `9f10fe7`.

What that covers, precisely:

- **Reader scrolling works.** Confirmed by use rather than by inspection: the sticky-prompt report that produced `9f10fe7` could only have come from scrolling a trace with a finger on a phone, which is the thing that was impossible before `fb47763`.
- **The prompt band scrolls away** — the report `9f10fe7` answers.

What is still open, and I would rather leave it labelled than quietly bank it:

- **Whether the composer sits above the keyboard or under it.** Deleting the clamp trades one failure mode for another, and nobody has reported either outcome explicitly — "working" came in the context of the prompt band. If it turns out to hide under the keyboard, the fix is the `.term-keybar` gate in the follow-ups: the second review measured the required pan at keyboard **+75px**, and 39 of those px are a TUI key-bar with no business over a reader.

The description is updated with the third fix and with this verification status. Six commits; `tsc` clean, `exchanges: ok`, production build green.

PR #55, opened 2026-08-09
Trace reader: open on the last exchange, page backwards by byte window

thomwolf:
Opening a long trace meant normalizing the whole transcript before anything could be shown, and then landing on its **oldest** turn. On the longest session on this Space (`edbfc11f…`, 19 MB, ~1,420 turns) that is 676 KB on the wire and ~160 ms of server time for the first read after the file changes — which, while an agent is writing, is *every* read, since the parse memo is keyed on mtime.

The reader now opens at the end and pages backwards.

## How

A **window** is a byte range of the `.jsonl` run through the same normalizers, with line-aligned cursors handed back to the caller. Consecutive windows abut exactly, so the conversation stitches together with no gaps and no repeats — and the part of the trace nobody scrolled to is never read.

| request | answers with |
|---|---|
| `?tail=1&bytes=` | the last window of the transcript |
| `?before=<cursor>` | the window in front of one you already hold |
| `?after=<cursor>` | whatever has been appended since |
| `?summary=1` | whole-trace facts (`total`, `userTurns`, `usage`, `firstTs`), no turns |
| `?offset=&limit=` | **unchanged** — index paging, still used by the Overview cards and RENDER mode |

`total`, `userTurns`, `firstTs` and the session's token total are whole-trace facts a window cannot know, so a window does not claim them. They come from `?summary=1`, which the pane asks for 400 ms **after** its first paint; until it lands the header says `N turns loaded` rather than inventing a total. The SQLite harnesses (opencode, hermes) have no byte offsets to seek, so they answer the same shape with message indices as cursors — the reader treats a cursor as opaque.

In the pane: the scroller is anchored on the row you are reading, using that row's **real geometry**; exactly one window request is ever in flight; reaching byte 0 stops the paging; and a growing trace is followed at the end — 3 s while it moves, 10 s once it has been quiet, which is also how it notices movement resuming — without dragging a reader who has scrolled up.

## Numbers

Against `upstream/main` built and served from the same files, on two real traces: a 19 MB Claude transcript and a 4.8 MB codex rollout. Server time is `curl` against a local server; "first turn on screen" is Playwright, median of 5–9 opens.

| | Claude 19 MB | | codex 4.8 MB | |
|---|---|---|---|---|
| | before | after | before | after |
| bytes, first request | 676 KB | **82 KB** | 513 KB | **151 KB** |
| server time, transcript changed since last request | 160 ms | **6 ms** | 46 ms | **6 ms** |
| server time, transcript unchanged (memo hit) | 6 ms | 6 ms | 5 ms | 6 ms |
| the `?summary=1` whole-file read, 400 ms after the paint | — | 176 ms | — | 46 ms |
| first turn on screen, localhost | 181 ms | 167 ms | 172 ms | 189 ms |
| first turn on screen, 10 Mbps / 40 ms | 866 ms | **267 ms** | 868 ms | **355 ms** |
| opens on | oldest turn | **newest turn** | oldest turn | **newest turn** |

**The whole-file parse is deferred, not removed** — the summary row above is the correction. The reader still asks for `?summary=1` once per pane open, and since `viewMemo` is keyed on mtime, a transcript a working agent is appending to pays that 160–176 ms parse on every open, after the first paint instead of in front of it. An earlier version of this table implied that work was gone; it moved. What the window buys is the first view and every page after it (a `before=` window is ~6 ms and never touches the memo); what it does not buy is a server that never reads the whole file.

Read honestly: **on loopback this is a wash** — both are ~170–190 ms, dominated by render and app work, and the codex case is if anything marginally slower. The win is the bytes, and it shows the moment there is a real link between browser and Space. Main was never slow at serving a memoized parse — only at building one, and at putting 676 KB on the wire to show you the wrong end of the conversation.

The 384 KB default window was chosen by sweeping 128 KB – 1 MB in a browser: paint time is flat (155–171 ms, the list is virtualized either way) and only the payload moves.

## What deploying it found

The last three commits are fixes from running this on the testing Space (`thomwolf/agent-manager`, now on this branch on top of #51), which put it in front of a **codex rollout** — turns far denser than Claude's — for the first time:

- **`734ecb7`** — the "at least 30 turns" growth floor grew that rollout's window to 1.5 MB and its response to 591 KB, within sight of the whole-file page this exists to avoid. The floor is a guard against opening on a stub, not a target: 12 turns, doubling instead of quadrupling.
- **`5300305`** — scroll anchoring restored from the prefix-sum height model, which is estimates until a row is measured; on rows several times `ROW_EST` a prepend moved the text under the reader ~280 px on the Space (~30 px locally, small enough to look like nothing). The anchor now carries the row's real `getBoundingClientRect()` position. 0 px drift on both traces.
- **`4fcb868`** — the same estimates made a 19-turn window model as 800 px against a 774 px pane, so the reader thought it was at the top and fetched an older window on every open. Paging now waits for a measurement, not just a placement. Opening a pane is back to exactly two requests.

**And one the operator caught while looking at it (`bd5aab9`), same principle one pane over:** switching a pane from terminal to **reader** mode landed at the *top* of the last 400 turns, because that view only placed itself when the agent was `working` — which most agents, most of the time, are not. On a 1,420-turn session that is 5,033 px above the end, hundreds of rows behind what you were just watching in the terminal. The placement is no longer conditional; `stick` still goes false the instant you scroll up, verified by parking the view and watching it stay across a poll tick.

## Review round (`d49cd6d`)

A review of the branch found six real defects, every one the same mistake in a different place — a window answering a question only the whole file can. All six now have tests that fail on the previous commit:

1. **The codex guardian guard never fired on a real rollout.** It read a fixed 16 KB and parsed the first line; a real `session_meta` carries `base_instructions.text` and runs 18–44 KB here (measured 18,596 B and 43,865 B), so the parse always threw and `catch { return }` turned "I could not tell" into "not a subagent". Any guardian rollout bigger than one window rendered as the operator's conversation. The read now grows until the line is whole; an unreadable head falls back to the whole-file reader, which throws on the marker itself. `/api/files/:id/trace` also mapped that refusal to a 500 — fixed.
2. **A load could outlive its source.** Switching files in the Files pane while a window was in flight prepended the old file's turns to the new file's list, with the old file's byte cursors and header. Every load now carries a generation stamp.
3. **A window's `note`** (codex's count of encrypted reasoning steps) was rendered as a fact about the session, and changed as you scrolled — 40, then 23, 25, 23. It now travels with `usage`/`firstTs` as summary-only.
4. **A line larger than the window read as the beginning of the conversation.** A 9 MB file-history blob cannot be paged past; the client turned "no progress" into `atStart`. The server now says `blocked` and the reader says what actually happened.
5. **A torn last line cost the final answer its accent forever** — `atEnd` meant "consumed every byte" rather than "read to EOF".
6. **`lastTs` went backwards when paging back**, dropping the live tail from 3 s to 10 s, and an older window's `model` replaced the current one. Both are monotonic in the merge now.

Plus: index-mode windows (opencode/hermes) page 100 turns rather than 12; the harness sniff memo gets a 60 s TTL so a re-imported bundle ref is not pinned to its old format; a prompt jump that has to fetch waits for the prepend to commit.

## What I verified, and how

**In a real browser, locally** (Playwright/Chromium, this branch on :7999 vs `upstream/main` on :7998, same transcripts, same seeded sessions). Every check ran on both builds; the control fails 7 of them:

- opens on the newest turn, with one tail request and nothing older fetched until asked
- scrolling into the top band prepends older turns and **the row under the reader stays at the same pixel** — asserted over an emulated 400 ms link so the before/after samples straddle the prepend, with the content above growing 2,800 px (Claude) and 4,171 px (codex)
- 25 scroll events at the top → 20 requests, never more than one window request in flight
- paging back reaches the first turn and then stops — 0 further requests while held at the top
- a transcript appended to while open: new turns appear in 1.8 s, exactly once, follows the end while pinned there and does not jump when the reader has scrolled up
- the same reader in the Files pane renders a `.jsonl` opened from the file tree, opens at the end, reports its turn count

**Reader mode**, timeline sampled every 200 ms rather than once at the end (a single late sample gave a false pass first time round): before, `scrollTop` sits at 0 for the whole 6 s after entering reader mode; after, it is at the end from the first frame the reader exists, on the same session, on the same Space.

**On the deployed Space** (private, driven through `https://thomwolf-agent-manager.hf.space` with a bearer token), against a codex **session** trace and an imported **bundle**: opens at the end in ~360 ms, two requests, anchor holds at 0 px drift on both, paging back reaches the beginning in 7 requests and stops, one window request in flight.

**Server tests:** new `server/test/trace-window.test.mjs` — stitching every window reproduces a 600-turn conversation exactly, paging past byte 0 returns nothing and stays put, an appended turn arrives once, a torn last line is not a turn until it is whole, a transcript with no trailing newline still ends in one, a 300 KB turn bigger than the window still arrives. `npm test` in `server/` passes in full (including both boot suites); `tsc --noEmit` and `web/test/exchanges.test.mjs` clean.

**Not verified:** mobile/touch. OpenClaw, STS and the SQLite harnesses go through the shared code paths and the unit tests, but I only opened Claude and codex traces in a browser. The codex subagent guard is re-checked from the head of the rollout, since a tail window never sees `session_meta`.

**One behavioural difference, on purpose:** a harness message whose lines straddle a window boundary is split into two turns instead of merged into one. Stitching every window of the 19 MB transcript gives 1,433 turns against the full parse's 1,420 — character-for-character identical content. Documented at the window reader and in `docs/trace-panel-spec.md` §13.

### DIFF
--- a/docs/trace-panel-spec.md
+++ b/docs/trace-panel-spec.md
@@ -450,6 +450,22 @@ and one older window was fetched on every open. Paging is therefore gated on a f
 waits for both the first measurement and the first placement of the scroller. Opening a
 pane now costs exactly two requests, the tail and the summary, on both traces.
 
+**Things a review caught, all of them about a window claiming more than it knows.** The
+codex guardian guard read a fixed 16 KB and parsed the first line; a real `session_meta`
+carries `base_instructions` and runs 18–44 KB, so the parse always threw and the `catch`
+turned "I could not tell" into "not a subagent" — the guard never fired on a real rollout.
+It now grows the read until the line is whole, and a head it genuinely cannot read falls
+back to the whole-file reader rather than guessing. A window's `note` (codex's count of
+encrypted reasoning steps) is a per-parse number and was being rendered as a statement
+about the session, changing as you scrolled — it now travels with `usage`/`firstTs` as a
+summary-only fact. A single line larger than `WINDOW_MAX_BYTES` (a file-history blob)
+cannot be paged past; the reader used to conclude it had reached the beginning of the
+conversation, and now says what actually happened. A last line that never becomes valid
+JSON no longer costs the final answer its accent: having READ to EOF is what makes a
+window the end, not having consumed every byte. And in the pane, every load now carries a
+generation stamp, so a window in flight when you switch files cannot be spliced into the
+next transcript with the previous one's byte cursors.
+
 **The one behavioural difference:** a harness message whose lines straddle a window
 boundary is split into two turns instead of merged into one. Stitching every window of
 that transcript gives 1,433 turns against the full parse's 1,420, with character-for-
--- a/docs/trace-panel-spec.md
+++ b/docs/trace-panel-spec.md
@@ -434,6 +434,14 @@ turn floor that makes a window grow (`WINDOW_MIN_TURNS`) is deliberately low for
 reason: at 30 it grew the codex rollout's window to 1.5 MB and its response to 591 KB,
 which is the whole-file cost this exists to avoid.
 
+**Scroll anchoring is measured, not modelled.** The first version restored the reading
+position from the prefix-sum height model, which is only ever approximately right for rows
+it has not measured yet: on a codex rollout — whose collapsed tool groups are several
+times `ROW_EST` — a prepend moved the text under the reader by ~280 px on the Space and
+~30 px locally. The reader now records the anchor row's real `getBoundingClientRect()`
+position and re-applies *that* after the layout, falling back to the model only for a row
+that has scrolled out of the rendered window. Drift on both traces is 0 px.
+
 **The one behavioural difference:** a harness message whose lines straddle a window
 boundary is split into two turns instead of merged into one. Stitching every window of
 that transcript gives 1,433 turns against the full parse's 1,420, with character-for-
--- a/docs/trace-panel-spec.md
+++ b/docs/trace-panel-spec.md
@@ -442,6 +442,14 @@ times `ROW_EST` — a prepend moved the text under the reader by ~280 px on the 
 position and re-applies *that* after the layout, falling back to the model only for a row
 that has scrolled out of the rendered window. Drift on both traces is 0 px.
 
+**"Am I at the top?" is only answerable after measurement.** The reader fetches the window
+above when `scrollTop` enters a 400 px band, and `scrollTop` means nothing until the rows
+have been measured: before the first measuring pass every row is `ROW_EST` (44 px), so a
+window of 19 dense codex turns models as 800 px when it is really 3,600 — inside the band,
+and one older window was fetched on every open. Paging is therefore gated on a flag that
+waits for both the first measurement and the first placement of the scroller. Opening a
+pane now costs exactly two requests, the tail and the summary, on both traces.
+
 **The one behavioural difference:** a harness message whose lines straddle a window
 boundary is split into two turns instead of merged into one. Stitching every window of
 that transcript gives 1,433 turns against the full parse's 1,420, with character-for-
--- a/docs/trace-panel-spec.md
+++ b/docs/trace-panel-spec.md
@@ -414,9 +414,18 @@ on screen" is Playwright, median of 5–9 opens of the pane.
 | bytes, first request | 676 KB | **82 KB** | 513 KB | **151 KB** |
 | server time, transcript changed since the last request | 160 ms | **6 ms** | 46 ms | **6 ms** |
 | server time, transcript unchanged (parse memo hit) | 6 ms | 6 ms | 5 ms | 6 ms |
+| the `?summary=1` read, 400 ms after the paint | — | 176 ms | — | 46 ms |
 | first turn on screen, localhost | 181 ms | **167 ms** | 172 ms | 189 ms |
 | first turn on screen, 10 Mbps / 40 ms link | 866 ms | **267 ms** | 868 ms | **355 ms** |
 | opens on | oldest turn | **newest turn** | oldest turn | **newest turn** |
+
+Read that table with the summary row in view: the whole-file parse is **deferred, not
+removed**. The reader still asks for `?summary=1` once per pane open, and on a transcript
+being appended to (the memo is keyed on mtime) that is the same 160–176 ms parse as before,
+just after the first paint instead of in front of it. What the window buys is the first
+view and every subsequent page; what it does not buy is a server that never reads the whole
+file. The honest summary of the server work is: one bounded read on the critical path, one
+whole-file read off it, and none at all for the pages after the first.
 
 Read honestly: **on loopback this is a wash** — both are ~170–190 ms, dominated by render
 and app work, and the codex case is if anything marginally slower. The win is the bytes,
@@ -466,10 +475,34 @@ window the end, not having consumed every byte. And in the pane, every load now 
 generation stamp, so a window in flight when you switch files cannot be spliced into the
 next transcript with the previous one's byte cursors.
 
-**The one behavioural difference:** a harness message whose lines straddle a window
-boundary is split into two turns instead of merged into one. Stitching every window of
+**codex windows are cut at task boundaries, not at arbitrary bytes.** A byte window is
+sound for a format whose lines stand alone. Codex's do not: `event_msg/task_complete`
+points BACK at the assistant message it marks, and a dangling pointer does not degrade —
+`normalizeCodex` cannot find the answer, so it pushes a verbatim second copy of it.
+Measured on a real 4.8 MB rollout at 128 KB windows: 955 stitched turns against the full
+parse's 953, two answers shown twice. A rollout describes its own unit
+(`task_started` … `task_complete`) and every backward reference lives inside one task, so a
+codex window now begins on a `task_started` line. Measured over every rollout on this
+Space: 339 `task_started`, 333 `task_complete`, none outside a started task, largest task
+768 KB — inside the 8 MB growth ceiling with room to spare. A task bigger than the ceiling
+cannot be aligned; there the pointer is treated as a marker only, since the answer it names
+is rendered (and marked) in the window that holds it. Verified: a synthetic 10 MB
+single-task rollout yields exactly one copy of its answer.
+
+With that, stitching every window of a codex rollout reproduces the full parse **exactly** —
+same turns, same blocks, same `final` accents — at 64 KB, 128 KB and 384 KB windows, on all
+six rollouts on this Space.
+
+**The one behavioural difference, and it is Claude's:** a harness message whose lines
+straddle a window boundary is split into two turns instead of merged into one. Stitching every window of
 that transcript gives 1,433 turns against the full parse's 1,420, with character-for-
-character identical content — thirteen turns show as two rows each. Covered by
+character identical content — thirteen turns show as two rows each. Checked as a multiset
+of blocks against the full parse across the six largest Claude transcripts and the six
+largest codex rollouts here, at three window sizes: **0 blocks missing and 0 duplicated in
+all 36 runs**. The `final` accent is withheld (never wrongly added) from a window that
+cannot see whether an answer is the last one before the next prompt: at the shipped 384 KB
+window the accents match the full parse exactly on both formats; at an unusually small
+64 KB window a 19 MB Claude transcript withholds 3 of 56. Covered by
 `server/test/trace-window.test.mjs`, which also pins the edges: paging back terminates at
 byte 0, an appended turn arrives exactly once, a torn last line is not a turn until it is
 whole, and a transcript with no trailing newline still ends in one.
--- a/docs/trace-panel-spec.md
+++ b/docs/trace-panel-spec.md
@@ -415,6 +415,7 @@ on screen" is Playwright, median of 5–9 opens of the pane.
 | server time, transcript changed since the last request | 160 ms | **6 ms** | 46 ms | **6 ms** |
 | server time, transcript unchanged (parse memo hit) | 6 ms | 6 ms | 5 ms | 6 ms |
 | the `?summary=1` read, 400 ms after the paint | — | 176 ms | — | 46 ms |
+| bytes `pread` to page the whole trace back | — | 20.8 MB / 35 windows | — | 5.7 MB / 13 windows |
 | first turn on screen, localhost | 181 ms | **167 ms** | 172 ms | 189 ms |
 | first turn on screen, 10 Mbps / 40 ms link | 866 ms | **267 ms** | 868 ms | **355 ms** |
 | opens on | oldest turn | **newest turn** | oldest turn | **newest turn** |
@@ -489,9 +490,16 @@ cannot be aligned; there the pointer is treated as a marker only, since the answ
 is rendered (and marked) in the window that holds it. Verified: a synthetic 10 MB
 single-task rollout yields exactly one copy of its answer.
 
-With that, stitching every window of a codex rollout reproduces the full parse **exactly** —
+Be precise about which half does what, because the first version of this section claimed
+more than the code delivers. Alignment is an *optimization*: when a window happens to hold a
+`task_started` it begins there and the task's grouping is not split at the seam. Most windows
+hold none (at 64 KB, with tasks up to 768 KB, almost none do) and are cut mid-task anyway —
+what keeps those correct is the second rule, that a `task_complete` whose answer is not in
+this window is treated as a marker and never emitted. **That fallback is load-bearing.**
+
+With both, stitching every window of a codex rollout reproduces the full parse **exactly** —
 same turns, same blocks, same `final` accents — at 64 KB, 128 KB and 384 KB windows, on all
-six rollouts on this Space.
+eight rollouts on this Space (24 runs, independently reproduced in review).
 
 **The one behavioural difference, and it is Claude's:** a harness message whose lines
 straddle a window boundary is split into two turns instead of merged into one. Stitching every window of
--- a/docs/trace-panel-spec.md
+++ b/docs/trace-panel-spec.md
@@ -403,19 +403,41 @@ indices as cursors (`window.mode`), and the reader treats a cursor as opaque eit
 which the reader asks for 400 ms after its first paint. Until it lands the header says
 `N turns loaded` rather than inventing a total.
 
-**Cost, measured on the 19 MB / 1,395-turn transcript** (`edbfc11f…`, warm page cache):
+**Cost.** Measured against `upstream/main` built and served from the same files, on
+two real traces on this Space: a 19 MB / ~1,420-turn **Claude** transcript and a 4.8 MB /
+~1,200-turn **codex** rollout. Server time is `curl` against a local server; "first turn
+on screen" is Playwright, median of 5–9 opens of the pane.
 
-| | before | after |
-|---|---|---|
-| server time for the first request | 165 ms | 5 ms |
-| bytes on the wire | 676 KB | 96 KB |
-| first turn on screen (localhost) | 180 ms | 155 ms |
-| first turn on screen (10 Mbps / 40 ms) | 865 ms | 267 ms |
+| | Claude 19 MB | | codex 4.8 MB | |
+|---|---|---|---|---|
+| | before | after | before | after |
+| bytes, first request | 676 KB | **82 KB** | 513 KB | **151 KB** |
+| server time, transcript changed since the last request | 160 ms | **6 ms** | 46 ms | **6 ms** |
+| server time, transcript unchanged (parse memo hit) | 6 ms | 6 ms | 5 ms | 6 ms |
+| first turn on screen, localhost | 181 ms | **167 ms** | 172 ms | 189 ms |
+| first turn on screen, 10 Mbps / 40 ms link | 866 ms | **267 ms** | 868 ms | **355 ms** |
+| opens on | oldest turn | **newest turn** | oldest turn | **newest turn** |
+
+Read honestly: **on loopback this is a wash** — both are ~170–190 ms, dominated by render
+and app work, and the codex case is if anything marginally slower. The win is the bytes,
+and it shows the moment there is a real link between the browser and the Space (3.2× and
+2.4×), plus the server work that vanishes: the "changed since the last request" row is the
+one a *working* agent hits, because the parse memo is keyed on mtime and every append
+invalidates it. The memo-hit row is what an idle trace costs, and there the two are the
+same — main was never slow at *serving* a memoized parse, only at building one and at
+putting 676 KB on the wire to show you the wrong end of the conversation.
+
+The 384 KB default window was chosen by sweeping 128 KB – 1 MB in a browser: paint time is
+flat (155–171 ms median, since the list is virtualized either way) and only the payload
+moves, so the smallest window that still opens on a full-looking conversation wins. The
+turn floor that makes a window grow (`WINDOW_MIN_TURNS`) is deliberately low for the same
+reason: at 30 it grew the codex rollout's window to 1.5 MB and its response to 591 KB,
+which is the whole-file cost this exists to avoid.
 
 **The one behavioural difference:** a harness message whose lines straddle a window
 boundary is split into two turns instead of merged into one. Stitching every window of
-that transcript gives 1,401 turns against the full parse's 1,395, with character-for-
-character identical content — six turns show as two rows each. Covered by
+that transcript gives 1,433 turns against the full parse's 1,420, with character-for-
+character identical content — thirteen turns show as two rows each. Covered by
 `server/test/trace-window.test.mjs`, which also pins the edges: paging back terminates at
 byte 0, an appended turn arrives exactly once, a torn last line is not a turn until it is
 whole, and a transcript with no trailing newline still ends in one.
--- a/docs/trace-panel-spec.md
+++ b/docs/trace-panel-spec.md
@@ -369,3 +369,53 @@ dedupe per ENTRY, not per joined string; and web searches exist ONLY as `web_sea
 
 **Encrypted reasoning is normal.** 49 of 56 reasoning items carried only
 `encrypted_content`. Say so in the UI; do not leave the impression the model didn't think.
+
+## 13. Tail-first reading (2026-08-09)
+
+§5 shipped one way of asking for turns: `offset`/`limit` over a whole-file parse. It is
+the right shape for the surfaces that want a fixed slice (the Overview card, RENDER mode)
+and the wrong one for the reader, because it can only answer *any* question after
+normalizing the entire transcript — 19 MB and 1,395 turns for the longest session on this
+Space — and the pane then opened on the OLDEST turn, which is never the one you came for.
+
+The reader now opens at the end and pages backwards.
+
+**Wire.** Same endpoints, three modes (`server/src/index.js` `traceOpts`):
+
+| Request | Answers with |
+|---|---|
+| `?tail=1&bytes=` | the last window of the transcript |
+| `?before=<cursor>` | the window in front of one you hold |
+| `?after=<cursor>` | whatever has been appended since |
+| `?summary=1` | whole-trace facts (`total`, `userTurns`, `usage`, `firstTs`), no turns |
+| `?offset=&limit=` | unchanged — index paging over the full parse |
+
+A window is a byte range of the .jsonl run through the same normalizers
+(`rangeJsonLines`), and the cursors it hands back are line-aligned, so consecutive
+windows abut exactly: no line is read twice and none is skipped. A window that starts
+mid-line drops that fragment (the window before it owns the line); a window that ends at
+a half-written line stops before it, so a live tail picks it up once it is whole. The
+SQLite harnesses have no byte offsets to seek — they answer the same shape with message
+indices as cursors (`window.mode`), and the reader treats a cursor as opaque either way.
+
+**What a window cannot know**, and therefore does not claim: `total`, `userTurns`,
+`firstTs` and the session's `usage` are whole-trace facts. They arrive from `?summary=1`,
+which the reader asks for 400 ms after its first paint. Until it lands the header says
+`N turns loaded` rather than inventing a total.
+
+**Cost, measured on the 19 MB / 1,395-turn transcript** (`edbfc11f…`, warm page cache):
+
+| | before | after |
+|---|---|---|
+| server time for the first request | 165 ms | 5 ms |
+| bytes on the wire | 676 KB | 96 KB |
+| first turn on screen (localhost) | 180 ms | 155 ms |
+| first turn on screen (10 Mbps / 40 ms) | 865 ms | 267 ms |
+
+**The one behavioural difference:** a harness message whose lines straddle a window
+boundary is split into two turns instead of merged into one. Stitching every window of
+that transcript gives 1,401 turns against the full parse's 1,395, with character-for-
+character identical content — six turns show as two rows each. Covered by
+`server/test/trace-window.test.mjs`, which also pins the edges: paging back terminates at
+byte 0, an appended turn arrives exactly once, a torn last line is not a turn until it is
+whole, and a transcript with no trailing newline still ends in one.
--- a/server/node_modules
+++ b/server/node_modules
@@ -1,0 +1,1 @@
+/app/server/node_modules
--- a/server/package.json
+++ b/server/package.json
@@ -14,7 +14,7 @@
     "start": "node src/index.js",
     "dev": "node --watch src/index.js",
     "test:ui": "node terminal-ui.test.mjs",
-    "test": "node test/spawn-group.test.mjs && node test/repin.test.mjs && node test/opencode-resume.test.mjs && node test/terminal-modes.test.mjs && node test/trace-tail.test.mjs && node migration.test.mjs && node resize.test.mjs"
+    "test": "node test/spawn-group.test.mjs && node test/repin.test.mjs && node test/opencode-resume.test.mjs && node test/terminal-modes.test.mjs && node test/trace-tail.test.mjs && node test/trace-window.test.mjs && node migration.test.mjs && node resize.test.mjs"
   },
   "engines": {
     "node": ">=20.19"
--- a/server/src/index.js
+++ b/server/src/index.js
@@ -1663,16 +1663,35 @@ app.delete('/api/files/:id/entry', (req, res) => {
 // One page of an on-disk transcript, rendered by the same reader the Trace pane
 // uses. Paged rather than whole: these files reach tens of MB, and the viewer
 // only ever has a window of them on screen.
+// Which slice of a trace the caller is asking for. Three modes, one endpoint:
+//
+//   ?tail=1 | ?before=<cursor> | ?after=<cursor>   a window (the reader's path):
+//       the last stretch of the conversation, the stretch in front of one it
+//       already holds, or whatever has been written since — each answered from a
+//       byte range of the transcript rather than a parse of all of it.
+//   ?summary=1                                     whole-trace facts, no turns.
+//   ?offset=&limit=                                index paging, as before.
+function traceOpts(q) {
+  if (q.summary !== undefined) return { summary: true };
+  const at = q.tail !== undefined ? 'tail' : q.before !== undefined ? 'before' : q.after !== undefined ? 'after' : null;
+  if (!at) return { offset: Number(q.offset) || 0, limit: Number(q.limit) || 200 };
+  return {
+    window: {
+      at,
+      cursor: Number(at === 'before' ? q.before : q.after) || 0,
+      bytes: Number(q.bytes) || 0,
+      min: Number(q.min) || 0,
+    },
+  };
+}
+
 app.get('/api/files/:id/trace', async (req, res) => {
   const s = store.get(req.params.id);
   if (!s) return res.status(404).json({ error: 'not found' });
   const f = resolveSafe(folderPathOf(s), req.query.path);
   if (!f || !fs.existsSync(f) || !fs.statSync(f).isFile()) return res.status(404).json({ error: 'not found' });
   try {
-    res.json(await readTraceByPath(f, {
-      offset: Number(req.query.offset) || 0,
-      limit: Number(req.query.limit) || 200,
-    }));
+    res.json(await readTraceByPath(f, traceOpts(req.query)));
   } catch (e) {
     // "not a transcript" is an ordinary answer here, not a failure: the pane
     // falls back to showing the file as text.
@@ -2007,17 +2026,16 @@ app.get('/api/trace/:id', async (req, res) => {
   if (!pane) return res.status(404).json({ error: 'not found' });
 
   const source = pane.traceSource || { kind: 'session', ref: pane.id };
-  const offset = Number(req.query.offset) || 0;
-  const limit = Number(req.query.limit) || 200;
+  const opts = traceOpts(req.query);
 
   try {
     if (source.kind === 'bundle') {
       if (!/^[\w.-]+$/.test(String(source.ref))) return res.status(400).json({ error: 'bad bundle ref' });
-      return res.json(await readTraceBundle(path.join(DATA_DIR, 'traces', source.ref), { offset, limit }));
+      return res.json(await readTraceBundle(path.join(DATA_DIR, 'traces', source.ref), opts));
     }
     const target = store.get(source.ref);
     if (!target) return res.status(404).json({ error: 'source session is gone', code: 'no-trace' });
-    res.json(await readTrace(target, { offset, limit }));
+    res.json(await readTrace(target, opts));
   } catch (e) {
     // These are expected states, not failures: no transcript yet, an
     // unsupported CLI, or a codex guardian rollout. The pane renders the reason.
--- a/server/src/index.js
+++ b/server/src/index.js
@@ -1694,8 +1694,10 @@ app.get('/api/files/:id/trace', async (req, res) => {
     res.json(await readTraceByPath(f, traceOpts(req.query)));
   } catch (e) {
     // "not a transcript" is an ordinary answer here, not a failure: the pane
-    // falls back to showing the file as text.
-    if (['no-trace', 'unsupported-harness'].includes(e && e.code)) {
+    // falls back to showing the file as text. A codex guardian rollout is the
+    // same kind of answer — the session route already says so, and a 500 here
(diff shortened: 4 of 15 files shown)


thomwolf:
Reviewed by reading the branch at `perf/trace-tail-first` plus the code around it, and by running `server/test/trace-window.test.mjs` (passes) and three probes of my own against `readTraceByPath`. Six things, most severe first.

## 1. codex: the final answer is duplicated at every window seam

`server/src/traces.js:1224` — `normalizeCodex`'s `task_complete` branch walks **`out.messages`** backwards for the assistant text the event points at. For a window, `out.messages` holds only the turns in *that window*. When a boundary falls between an assistant `response_item` and the `event_msg/task_complete` that marks it, the new window contains no such message, `marked` is null, and the `else` at line ~1240 pushes `{ role: 'assistant', kind: 'final', … }` — a verbatim second copy of the answer the previous window already rendered.

Reproduced. A synthetic 60-exchange rollout (`user` / `assistant` / `task_complete`, ~600 chars each):

```
full parse:        120 turns
stitched windows:  122 turns, 2 adjacent identical assistant texts
  (bytes=8K, 16K and 32K all give the same 2)
DUPE at stitched index 6  "ANSWER 2 …"  roles assistant/assistant  kinds undefined/final
DUPE at stitched index 93 "ANSWER 45 …" roles assistant/assistant  kinds undefined/final
```

Both duplicates are the *first* turn of a window, which is the fingerprint of exactly this path. This is not the split §13 admits to: the content is doubled, not divided, and the `final` accent lands on the copy rather than the original. §13's "character-for-character identical content" claim was presumably measured on the Claude transcript — my Claude probe (200 exchanges, 4 lines per streamed assistant message, stitched at 32 KB) does come back identical at 400 turns, so that half of the claim holds and the codex half does not.

Cheapest fix: don't take the fallback push when the window did not begin at byte 0 (`range.start > 0`) — a `task_complete` whose answer is not in this window belongs to the window before it. The same per-window-state class of problem exists for `seenThinking` and Claude's `byMsgId`, but those only cost you a repeated thinking block or the documented split, not a duplicated answer.

## 2. a prompt jump near the top loads windows and lands on the wrong turn

`web/src/components/TracePane.tsx:573` — `goToTurn` writes an **absolute turn index** into `wanted.current`, sets `scrollTop`, and calls `recompute()` synchronously. `recompute` ends (line 520) with `if (positioned.current && el.scrollTop < NEAR_TOP_PX) loadOlder();`. So jumping to a prompt sitting in the top 400 px of what is loaded starts a page load.

`loadOlder` (line 390) prepends `got.length` turns and shifts `range` by the same amount, but nothing adjusts `wanted.current`. The layout effect at line 530 tests `wanted.current` **first**, ahead of both `stick` and `anchor`, so the anchor `loadOlder` just captured is discarded and `el.scrollTop = offsets[wanted.current]` now addresses turn `wanted` of the *newly prepended* window — a whole window (~20–60 turns) before the prompt the reader asked for.

It then repeats: the restored `scrollTop` is again below `NEAR_TOP_PX`, `offsets` changed so `recompute` is a fresh callback, and `useEffect(() => { recompute(); }, [recompute, meta])` runs it again → another `loadOlder`. Only `if (++tries.current > 8) wanted.current = null` stops it. One ▲ click can therefore pull several 384 KB windows and leave the view nowhere near the target. `onWheel`/`onTouchStart` clear `wanted`, but a toolbar button click does not.

Secondary: the `wanted` branch never clears `anchor.current`, so the anchor captured during the prepend survives into a later layout pass with an `i` that has since shifted again.

Fix: shift `wanted.current` (and `anchor.current.i`, if set) by `got.length` inside `loadOlder`, and consider putting the `anchor` branch ahead of `wanted` since the anchor is the measured truth and `wanted` is the model estimate.

## 3. prepending remounts every visible row, so open folds snap shut

`TracePane.tsx:656` keys rows by absolute index (`<div key={i} …>`); `:698` does the same for search matches. On `main` that was safe: the array was pre-sized to `total` and pages filled in place, so an index never changed meaning. With `loadOlder` prepending, every loaded turn moves by `got.length` and `range` moves with it, so the rendered children go from keys `k…k+39` to `k+got.length…k+got.length+39`. No key matches, so React unmounts and remounts all of them.

`Collapsible` holds `open` in `useState` (line 72). So **every expanded tool group, thinking block and system blob in the viewport collapses the moment an older window lands** — which is precisely the reading gesture this feature exists for (expand a tool call, scroll up for the context that produced it). It also discards all the markdown memoized in `TextBlock` and forces a full ResizeObserver re-measure of the viewport.

The anchoring itself survives this — the layout effect reads the anchor row's real `getBoundingClientRect()` after the commit, so that row stays put — but everything below it shifts as the folds close, and the reader loses what they had opened.

Fix: key by something that doesn't move. A `base` ref decremented by `got.length` on each prepend, with `key={base.current + i}`, is enough.

## 4. the "reasoning was encrypted" disclosure never reaches the reader

`traces.js:1589` — `windowOf` sets `note: cur.atStart && cur.atEnd ? parsed.note : null`, which is right; a window cannot count the session's encrypted reasoning steps. §13 says the note "now travels with `usage`/`firstTs` as a summary-only fact", and the server does deliver it there. Verified on a 60-exchange codex file with encrypted reasoning:

```
tail window : note = null
?summary=1  : note = "60 reasoning steps were encrypted by the model and carry no readable text"
```

The client never picks it up. `head` (TracePane.tsx:665–673) takes `total`, `userTurns`, `usage`, `firstTs` and `truncated` out of the summary and nothing else, and the render at line 716 uses `meta.note`, not `head.note`. So for every codex trace longer than one window — i.e. every real one — the pane silently stops saying the model's reasoning was withheld, which §12 of the spec calls out as the one thing that must not be left implied. On `main` it was always shown.

The same discard costs the rest of `headOf`. Verified on a bare STS `.jsonl` opened through the Files pane, where `title`/`harness`/`id` live only on the single `{type:'session'}` first line that `normalizeSts` reads:

```
tail window : title=""                     harnessLabel="sts"      sessionId=null
?summary=1  : title="My important session" harnessLabel="opencode" sessionId="sess-1"
```

Bundles are rescued by `decorate` (which runs on every window parse); a bare STS file is not, so its header shows the wrong harness and no title indefinitely. Fix both by folding the summary's head fields into `head` (`title: meta.title || summary?.title`, and so on) and rendering `head.note`.

## 5. the whole-file read is deferred by 400 ms, not removed

The window read itself is genuinely bounded, and I want to be clear about that: `rangeBuf` (`traces.js:956`) is an `fsp.open` plus a positional `read` of at most `WINDOW_MAX_BYTES`, and `rangeJsonLines` (`:979`) splits the buffer rather than streaming the file. No whole-file read hides behind the slice.

But `serveTrace:1747` answers `?summary=1` from `full()` — an unranged `parseTraceFile` over the entire transcript — and TracePane fires exactly that 400 ms after every open (line 494). `viewMemo` is keyed on `mtime:size`, so for the working agent the cost table is written about, the one whose appends invalidate the memo on every request, the 160 ms / 19 MB parse still happens on every pane open. It has moved off the first paint, not off the server. The table's "server work that vanishes" row reads as though it were gone; it would be more honest to add the summary's cost as its own line.

## 6. the new 3 s poll re-parses the whole conversation for opencode/hermes

`TracePane.tsx:502`'s `setInterval(loadNewer, …)` is new — `main`'s TracePane never polled. For the SQLite harnesses `serveTrace:1749` takes `if (!WINDOWABLE.has(harness)) return windowIndex(await full(), opts.window)`, a full `parseTraceFile` of the DB. While the agent works, `mtime` changes constantly so `viewMemo` misses every time, so an open Trace pane on an opencode session now costs a full re-parse of the conversation every 3 seconds. The response stays small; the work does not. `viewMemo` is a single slot, so this also evicts whatever the Overview cards had memoized. A cheap mitigation is to skip `loadNewer` entirely when `cursor.mode === 'index'` and `atEnd`, or to poll index-mode sources at `IDLE_MS`.

---

## What I checked and found sound

- **UTF-8 at the boundary.** Splitting is done on the Buffer (`buf.indexOf(0x0a)`, `buf.toString('utf8', i, nl)`), so a `from` that lands mid-character is harmless. I stitched a 300-turn Claude transcript whose every line is `é🙂漢`-heavy at 32 KB windows: 300 turns, character-identical to the full parse.
- **Partial records.** A window that starts mid-line drops the fragment (the window before owns it); a window that ends mid-line defers it (`if (last && !range.eof) break`, and `range.end` only advances past an unterminated *final* line if it parses). `?after=` picks it up once whole. The new test pins this and it passes.
- **Concurrency.** `loading.current` is set synchronously before the first `await` in all three loaders and cleared in `finally`. The `mine === gen.current` guard correctly declines to clear a *newer* generation's flag, and the `srcKey` effect clears it on the way in. I could not construct a path that leaves it stuck true after an error — the `catch` in `loadOlder`/`loadNewer` falls through to the `finally`.
- **Termination.** `loadOlder` returns early on `cur.atStart || cur.blocked`; the server sets `atStart` from `start <= 0`; the client additionally treats a non-advancing cursor as the start. Paging back over a 600-turn file terminates at byte 0, and `before: 0` returns empty with `atStart` rather than looping.
- **`blocked`.** A window that consumes nothing (`start >= to && to > 0`) is correctly distinguished from the start of the trace, and the client stops paging on it.
- **Claude cross-window merging.** The split §13 admits to is real but lossless: 200 exchanges with 4 lines per streamed assistant message stitched back to the same 400 turns with identical text.
- **Non-paged callers.** `runner.js:651`, `Overview.tsx` and `ConversationView.tsx` all still go through `pageOf`, whose response shape is unchanged (`headOf` reproduces every field it used to spread inline). The `ConversationView` `live` removal looks right — `stick` is maintained by its own `onScroll`.

## What I did not check

I did not build or type-check the client, did not run it in a browser, and did not boot the server, so findings 2 and 3 and the client half of 4 are read out of the code rather than observed — in particular I have not watched the runaway in 2 actually fire, only traced that nothing prevents it. My codex duplication repro (1) is synthetic; I had no real rollout to hand, though the mechanism is plain in `normalizeCodex`. I did not exercise the opencode/hermes index-window path at all, and I did not test a trace being appended to concurrently by a real agent (only the test's simulated torn line).

thomwolf:
Worked through all six. Five fixed, one rejected on measurement, and your #1 changed the design rather than a detail of it. Pushed as `bf4fb36` (codex) and `b5d584e` (reader).

## 1. codex duplication — you were right, and the repair is structural

Reproduced on a **real** rollout, not just synthetically: 4.8 MB, 128 KB windows → **955 stitched turns against the full parse's 953**, +175 characters, two answers rendered twice. The raw lines at one seam are exactly the mechanism you named:

```
 last lines BEFORE the boundary          first lines AFTER
 event_msg/agent_message  "The Cowrite…" event_msg/task_complete "The Cowrite…"
 response_item/message:assistant "The…"  event_msg/task_started
```

I did not dedupe it. `task_complete` is a pointer, and a scheme that cuts pointers is not made sound by cleaning up after itself. But a codex rollout **describes its own unit** — `task_started` … `task_complete` — and every backward reference in the format lives inside one task. So the windowing unit for codex is now the task: a window begins on a `task_started` line, and each task lands whole in one window with its answer, its reasoning and its token count.

I measured the format before committing to that, across every rollout on this Space:

```
339 task_started   333 task_complete   0 task_complete outside a started task
largest task 768 KB   (growth ceiling 8 MB)
```

The outcome is stronger than "no duplicates": **stitching now reproduces the full parse exactly** for codex — same turns, same blocks, same `final` accents — at 64 KB, 128 KB and 384 KB on all six rollouts here. Across the twelve largest traces of both formats at three window sizes, checked as a multiset of blocks: **0 missing, 0 duplicated, 36/36 runs**. A task bigger than the ceiling cannot be aligned; there the pointer is treated as a marker only, since the answer it names is rendered *and marked* in the window that holds it — verified on a synthetic 10 MB single-task rollout, which yields exactly one copy.

Two things fell out that your note predicted in spirit: `seenThinking` and the turn grouping stop straddling seams too, because tasks no longer straddle them. And `unmarkTrailingFinal` was withholding the accent from windows that could perfectly well know — for codex that is precisely a window whose last task is still open at its edge, now tracked in the parse. Accents went 267→266 before, 267→267 after.

**One correction to your suggested fix, with evidence.** I first tried deriving the accents client-side over the stitched list. That destroys them: measured on a rollout, 267 of the full parse's accents come from `task_complete`, and the derivation rule finds **12**. On Claude the same derivation *over-marks* (72 against 56) because split turns give it more candidates. The accents have to stay server-side.

## 2. the prompt-jump stampede — right, fixed, with numbers

I could not get it to fire from a single click in isolation, so I walked ▲ back twelve times through the 19 MB transcript and counted, previous build vs this one:

| | windows fetched | worst single click | landing offsets from the prompt |
|---|---|---|---|
| before | 19 | 4 | 19, 18, 19, 18, **117**, 18, **−100**, 19, 18, **1045**, 19 |
| after | 5 | 1 | 19, 18, 19, 19, 18, 19, 19, 18, 19, 19, 19, 19 |

`wanted` now moves with the prepend, a settling jump suppresses the automatic fetch (it does its own paging), and the jump branch clears the anchor it is deliberately leaving, as you suggested. I left `wanted` ahead of `anchor` in the layout effect: a jump is a request to *leave* the anchored position, so the anchor should lose there — the bug was the stale index, not the precedence.

While fixing it I introduced and caught a regression worth recording: shifting `anchor.i` by the prepend *as well as* `wanted` double-counts, because `captureAnchor` is already given the shift — it threw the view a whole window forward. Invisible to the anchor test; it surfaced as folds "not surviving" a prepend that had in fact scrolled elsewhere.

## 3. rows remounting — right, fixed

Confirmed by tagging the DOM nodes with an expando before a prepend: **0 of 50 survived**. Keys are now `keyBase + index`, base moving against the prepend. Of five folds opened and still on screen after a prepend: previous build kept **0**, this one keeps **5** (and 2 of 3 on a codex trace, where one fold left the rendered window by itself).

## 4. the disclosure never reaching the reader — right, fixed

`head` now falls back to the summary for `note`, `title`, `harnessLabel`, `sessionId`, `model`, `cwd` and provenance, and the pane renders `head.note`. On a codex trace: before, nothing; after, "382 reasoning steps were encrypted by the model and carry no readable text".

## 5. the deferred whole-file parse — right, and the PR body was overstating it

Corrected in the body rather than argued away. The table now carries the summary read as its own row (176 ms Claude / 46 ms codex) and says plainly that the parse moved off the first paint, not off the server, and that a working agent's mtime churn means it is paid on every pane open. Thank you for being precise about `rangeBuf` being a real positional pread — that part I did want on the record.

## 6. the 3 s poll on opencode/hermes — right, fixed

Those sources are no longer polled at all. Their "window" is a whole-DB parse, so a tick would re-read the conversation and evict the Overview's memo; before this feature the pane polled nothing, so this is a return to that rather than a new limitation.

## Verified by running vs reasoned about

**Run:** the codex repro and its fix on six real rollouts at three window sizes; the 36-run multiset comparison; the task-structure census; the 10 MB single-task ceiling case; the ▲ walk and the fold/expando measurements, each A/B against a build of the previous commit served by the same server; the encrypted-reasoning note on a codex pane; `npm test` in `server/` (including the new codex-alignment tests, which fail on the previous commit with "no answer is shown twice"), `tsc --noEmit`, `web/test/exchanges.test.mjs`, and the existing anchor / open-cost / live-append / reader-mode browser checks.

**Not run:** the opencode/hermes path — there is no such session on this box, so #6 is a code change I have type-checked and reasoned about but not exercised. I also have not re-deployed to the shared testing Space: another agent is actively deploying PR #56 there, and I am not going to replace their build to check mine.

thomwolf:
## What this PR is for, restated — because I had it pointed at the wrong surface

The request was: *the trace reader is slow on a long conversation; load the end first and page older messages in as you scroll back*. I built that in the **Trace pane** and shipped it. That was a misread. The Trace pane is not where anyone reads a conversation — **reader mode** is, and reader mode was the surface with the problem:

- it fetched the last 400 turns in one request (**702 KB** on a 19 MB session) and stopped there, showing `1,020 earlier messages are not shown` with no way to reach them;
- while an agent was working it re-fetched all 400 **every 3 seconds**, each one forcing a full re-parse of the transcript on the server.

`c145ecd` fixes that. Reader mode now opens on a window of the end, pages backwards when you scroll to the top, and follows the end with `after=`.

| on the 19 MB / 1,420-turn session | before | after |
|---|---|---|
| opening the reader | 702 KB | **84 KB** |
| poll while the agent works | 702 KB + whole-file parse | **359 bytes** + a bounded read |
| history you can reach | last 400 turns | all of it (34 windows to the first turn) |

Verified on the deployed Space just now: opens at the end, scrolls back to "beginning of the conversation" in 13.5 s.

### So the PR is now three things

1. **A range API for traces** (`?tail=`, `?before=`, `?after=`, `?summary=`), with `?offset=&limit=` untouched for the Overview cards. For codex, windows are cut at `task_started` boundaries — an arbitrary cut made the reader *invent* a duplicate answer, because `task_complete` points backwards.
2. **Reader mode pages back** — the actual ask.
3. **The Trace pane does too**, sharing the same paging via `web/src/lib/traceWindows.ts`. This is the part nobody asked for. It is a real fix (that pane opened on the *oldest* turn of a 19 MB file, 676 KB to show you the wrong end) and it is what the server API was reviewed against, but I am happy to split it into its own PR if the reviewer would rather see #55 as just "the reader pages".

### Anchoring differs between the two, deliberately

The Trace pane pins a row by its own geometry. Reader mode cannot: its unit is an exchange, and an exchange at the top of the list is a *fragment* whose prompt was in the window not yet read — when that window arrives, the two become one exchange and the element the reader was looking at ceases to exist. It restores the distance to the bottom instead, which regrouping cannot invalidate. Checked by what sits under the middle of the pane, not by element identity.

### No virtualization in reader mode, on purpose

Measured rather than assumed: the whole 1,420-turn conversation paged in is **2,719 DOM nodes at a 16.6 ms median frame** (worst 19.3 ms). Windowing by exchange (spec §10.7) remains the right next step for a session an order of magnitude longer — it is not needed at this one.

A review of the previous round is at #55 (comment above) and its six findings are fixed in `d49cd6d`/`bf4fb36`/`b5d584e`. The newest commit has not been reviewed by anyone but me.

thomwolf:
Reviewed at `c145ecd`, with the weight on that commit since it hasn't been looked at. Everything below was run against a clone of the branch: `server/test/trace-window.test.mjs` (passes), `npm test` in `server/` (all checks passed), `tsc --noEmit` (clean), a `vite build`, a throwaway server on :7995 over the `tracebench` fixtures driven with Playwright, and five scratch harnesses against `readTraceByPath` on the real transcripts on this box (19.2 MB / 15.8 MB claude, 4.8 MB codex, eight rollouts).

Seven findings, worst first. Where I A/B'd against a patched build I say so; where I only read the code I say that too.

---

## 1. An open Trace pane burns a CPU core, forever — and it is new in `c145ecd`

`web/src/lib/traceWindows.ts:252` with `web/src/components/TracePane.tsx:510`.

`head` is now built inline on every render of whoever calls the hook. It was behind a `useMemo` in every earlier commit of this PR — `34acd80:585`, `d49cd6d:664`, `b5d584e:700` — and moving it into the shared hook dropped the memo. `TraceView` still hands it up with `useEffect(() => { onHead?.(head); }, [head, onHead])` (`TracePane.tsx:510`) and the host puts it in state (`onHead={setHead}`, `TracePane.tsx:638`). New object every render → the effect fires every render → the parent's `setState` → re-render → new object → round and round. It never trips React's "maximum update depth" because the loop goes through `useEffect`, so it just spins.

Measured with Chrome's `Performance.getMetrics`, pane open on `Trace: big` and **not touched**:

```
nothing open   script 0.01s of 4.00s wall    0% busy
Trace pane     script 3.91s of 4.00s wall   98% busy      (14 rows rendered)
reader mode    script 0.02s of 4.00s wall    1% busy
```

I then wrapped the `head` object in `useMemo(..., [meta, summary, version, atStart, blocked])` inside the hook, rebuilt, and re-measured against the same server:

```
Trace pane     script 0.01s of 4.00s wall    0% busy
```

That is the whole fix. Reader mode escapes only because `ConversationView` consumes `head` itself instead of lifting it into a parent's state — it is not immune by design, just by shape.

`FilesPane.tsx:807` has the identical shape (`onHead={setTraceHead}` into a `useState` in `FileView`), so a `.jsonl` opened in the Files pane should do the same thing. **I did not demonstrate that one** — I could not get the Files pane to open the fixture file in this harness — so treat it as traced through the code rather than observed.

## 2. `onAppend` is written but never wired, so every live append corrupts the Trace pane's height model

`TracePane.tsx:299` defines `onAppend`; `TracePane.tsx:313` passes only `{ onPrepend, onReset }`. The callback is dead, `cb.current.onAppend?.()` at `traceWindows.ts:190` resolves to nothing, and the render-time repair at `TracePane.tsx:331-336` then does exactly the wrong thing — it sees `heights` shorter than `turns` and **prepends** the missing slots:

```js
heights.current = grew > 0 && heights.current.length
  ? [...new Array(grew), ...heights.current]   // older turns arrived on top
  : new Array(turns.current.length);
```

which is verbatim the bug the commit message says this commit fixed ("A single 'the array grew' path prepended height slots for an append too, which silently misaligned the Trace pane's height model on every live poll"). The mechanism was fixed; the wiring was not.

Demonstrated on `Trace: live` (341 turns). Sweep the list so every row is measured, note which row sits at a fixed `scrollTop`, append 12 turns **at the end**, come back to the same `scrollTop`. Nothing was added above, so the same row must still be there:

```
as shipped:
  measured scrollHeight 22717;  at scrollTop=9087 -> row 138 ("…6:13:35 PM row …")
  after 6 appends (12 turns, all at the END)      -> row 142 ("…6:13:39 PM row …")
  MISALIGNED — 4 rows of drift at the same scroll position

with `{ onPrepend, onAppend, onReset }` (one-word patch, rebuilt, same server):
  measured scrollHeight 23509;  at scrollTop=9404 -> row 142
  after the same 6 appends                        -> row 142   drift 0px
```

`scrollHeight` grew by exactly `12 × ROW_EST = 528px` rather than by the real height of the appended rows, which is the fingerprint: the whole `heights` array shifted and the last twelve real measurements are now attributed to different rows.

The anchor hides this at the moment of the append (the ResizeObserver captures a row and puts it back, so nothing visibly jumps) — what it does not hide is that afterwards the scroll model, the virtualizer's range, and `goToTurn`'s `offsets[i]` are all wrong for every row outside the rendered window, and it happens again on every 3-second poll of a working agent.

## 3. Reader mode can be permanently unable to page back

`ConversationView.tsx:229` — `loadOlder()` is called from `onScroll` and from nowhere else. If the first window does not overflow the pane, no scroll event can ever fire, and there is no other caller. `TracePane`'s `recompute` has precisely this guard, and the comment at `TracePane.tsx:349-351` names the case: *"Also covers a tail window too short to fill the pane — there is no scrolling to do, so nothing else would ask."* Reader mode has nothing equivalent.

Demonstrated on `big` (19.2 MB, 1,420 turns), same build, four viewport heights, 30 real `mouse.wheel` up-scrolls at the top each time:

```
viewport  900px: content 3087px, pane  707px, scrollable=true,  6 exchanges, 2 before= requests
viewport 1300px: content 3087px, pane 1107px, scrollable=true,  6 exchanges, 2 before=
viewport 1800px: content 3087px, pane 1607px, scrollable=true,  6 exchanges, 2 before=
viewport 2200px: content 2007px, pane 2007px, scrollable=FALSE, 4 exchanges, 0 before=
```

At 2200 px the reader is parked on the last four exchanges of a 19 MB conversation, with `earlier turns load as you scroll up…` sitting above them, and no gesture reaches the rest. A tall or portrait display gets you there directly; so does any tail window whose turns are short, since the growth floor is 12 turns and twelve one-line turns are a few hundred pixels. `loadTail` already has the "nothing to render, walk back" recovery for the *zero*-turn case (`traceWindows.ts:117`); this is the same problem one step later.

Fix: after every load and reset, if `scrollHeight <= clientHeight && !atStart && !blocked`, call `loadOlder()` — the unconditional check the Trace pane already makes.

## 4. One "window" can `pread` 5.76 MB, and paging a trace reads more bytes than the file has

`server/src/traces.js:1736-1751`. The growth loop doubles `span` and calls `oneWindow` again, and `oneWindow` reads the whole new span from scratch; nothing carries the previous buffer forward.

Measured with `fs.promises.open` instrumented to count bytes actually `pread`, 384 KB windows, paging from the tail to the first turn:

```
19.2 MB claude   35 windows, 163ms, 28.3 MB pread   worst window: 5760 KB in 4 preads, 26ms -> 58 turns
15.8 MB claude   17 windows, 116ms, 31.3 MB pread   worst window: 5760 KB in 4 preads, 14ms -> 19 turns
 4.8 MB codex    13 windows,  38ms,  5.2 MB pread   worst window:  448 KB in 2 preads,  3ms -> 92 turns
```

The *open* is honest and I want that on the record: the tail of the 19.2 MB file is **one** 384 KB `pread`, 60 turns, 81 KB of JSON, 98 ms cold. It is the growth path that costs — 384K + 768K + 1.5M + 3M for a single response. Cheap on local disk; not cheap on the bucket, which is the whole reason this PR exists. Reading `to - span` once at the final span, or extending the buffer rather than re-reading it, removes it.

**And in the same loop, `unaligned` (`traces.js:1741`) can never be true.** For any `before`/`tail` window with `from > 0` the range is built with `aligned: from === 0` → false, so `range.start` is always at least `from + 1` — either the byte after the first newline, or the `task_started` found at or after it. `w.cur.start <= from` therefore never holds; the "grow rather than serve a window whose references point outside it" branch never fires; a task larger than the window is cut anyway. Evidence: 64 KB windows on the 4.8 MB rollout, whose largest task is 768 KB, take **64 hops** — i.e. no window ever grew.

The *output* is still correct, because the `danglingTaskComplete: 'ignore'` fallback carries it (verified below). But the code comment and the PR body both say each task "lands whole in exactly one window", and that is not what the code does — what makes codex windows correct is the fallback, not the alignment. Either fix the condition (it should compare against the pre-alignment `from`, which `oneWindow` overwrites) or delete the dead guard and describe the fallback as the load-bearing part, because that is what a future reader will rely on.

## 5. Reader-mode search now searches one window and reports as if it searched the conversation

`ConversationView.tsx:115-120` filters `exchanges`, which is only what is loaded; `:243` says `N of M turns match`, where `M` is the loaded count. The Trace pane's equivalent line is careful about exactly this — `{matches.length} matches in the {n} turns loaded so far{atStart ? '' : ' — scroll up to load more'}` (`TracePane.tsx:530-531`).

Observed on `big`, immediately after opening the reader (10 exchanges loaded), typing `worktree`:

```
6 of 6 turns match “worktree” · 28 highlighted, ▲▼ walks them
```

Before this PR the same box searched the last 400 turns. The count isn't wrong; it is about something narrower than a reader will assume, and nothing on screen says so. One clause fixes it. (Paging while filtered does still work — I checked: scrolling the filtered list to the top fetched another window.)

## 6. Both surfaces now poll a stopped session forever, and the Trace pane cannot be paused

`traceWindows.ts:237-241` polls whenever the source is seekable and `opts.paused` is falsy — there is no notion of "this session is not running". `TraceView` has no `paused` prop at all (`TracePane.tsx:203-211`), and neither host passes one, while reader mode gets `paused={visible === false}` from `TerminalPane.tsx:859`.

Measured, session state `stopped`, pane open and untouched, 30 s:

```
reader mode: 3 trace requests {"after":3}
Trace pane : 3 trace requests {"after":3}
```

Every 10 s, indefinitely, per open pane. Before this PR reader mode polled only while `live` (`working && !paused`) and the Trace pane polled nothing at all. Each poll is small, so this is a cost note rather than a correctness one — but a Trace pane behind another tab has no way to stop, which is a step back from where the pane was. Two cheap things: give `TraceView` a `paused`, and skip the tick entirely when `lastTs` is older than `FRESH_MS` **and** the caller says the session isn't running.

## 7. Smaller things, read out of the code rather than observed

- **`ConversationView.tsx:109-110`** — `send()` sets `stick.current = true` and calls `loadNewer()`, but the layout effect that acts on `stick` runs on `[version, live]` (`:194`), and `loadNewer` bumps `version` only if turns actually arrived. The optimistic `sent` echo (`:263`) renders below without bumping anything. So sending from a scrolled-up position no longer scrolls to your own message; on `main` the deps were `[turns, live]` and every `load()` produced a fresh array, so it always did. Adding `sent` to the deps restores it. I could not exercise this — the fixture sessions have no agent behind them to accept input.
- **`traceWindows.ts:64`** — the comment says `model` is "the one the session is running now", but `out[k] &&` with no further test means the *first* model ever seen wins and is never replaced. Tail-first makes that usually the current one, so it is benign; the comment states the opposite rule from the code.
- **`traceWindows.ts:57-67`** — "Identity is preserved when nothing changed" holds for scalars only: `out[k] !== v` is always true for an object-valued field. `usage` is `null` on any window that isn't both `atStart` and `atEnd`, so it doesn't bite today, but the invariant in the comment isn't the one the code has.
- **`traceWindows.ts:204-216`** — the reset effect depends on `loadTail`, which depends on `src`. All three call sites memoize `src`; one that forgets would refetch the tail on every render, silently. Worth one line in the file's doc comment, since the point of that comment is to stop the next caller getting it wrong.

---

## What I checked and found sound

**Codex `task_started` alignment — I tried to break it and could not.** Stitched every window backwards through the API and compared against a full parse (walking `pageOf` in 500-turn pages, since it caps there), block by block, on **all eight real rollouts on this box at 64 KB / 128 KB / 384 KB — 24 runs, all IDENTICAL**, including turn count, block contents and the `final` accents:

```
rollout-2026-08-01T07-45-07 (4.8 MB) @64K   full=953 stitched=953 IDENTICAL final 267->267 hops=64
                                     @128K  full=953 stitched=953 IDENTICAL final 267->267 hops=39
                                     @384K  full=953 stitched=953 IDENTICAL final 267->267 hops=13
rollout-2026-08-03T12-47-51 (3.3 MB) @64/128/384K  84 -> 84  IDENTICAL  final 14->14
rollout-2026-08-03T13-12-12 (3.1 MB) @64/128/384K  94 -> 94  IDENTICAL  final 22->22
rollout-2026-07-29T23-28-03 (2.0 MB) @64/128/384K  95 -> 95  IDENTICAL  final 19->19
… (four smaller rollouts, all identical)
```

Note that the 64 KB runs are the interesting ones: with tasks up to 768 KB, most of those windows are cut *inside* a task, and the result is still exact. That is the `ignore` fallback doing the work, which is why I want finding 4's wording changed rather than the behaviour. Census across the same files: 329 `task_started` / 325 `task_complete`, no rollout with a `task_complete` and no `task_started`.

**Claude window seams lose nothing.** Same stitch on three claude transcripts, then a character-level comparison of every block's text concatenated in order:

```
5d654a93 (7.4 MB)   turns 1150 -> 1156   text 1,076,925 vs 1,076,925 chars   CONTENT IDENTICAL
345473eb (7.0 MB)   turns 1313 -> 1318   text 1,194,397 vs 1,194,397         CONTENT IDENTICAL
edbfc11f (19.2 MB)  turns 1420 -> 1433   text 2,799,661 vs 2,799,661         CONTENT IDENTICAL
```

The extra turns are exactly the §13 split (13 seams over 35 windows on the 19 MB file; each `missing 1, extra 2`), nothing is duplicated or reordered, and the `final` accents match (`56 -> 56`). §13's claim holds as stated.

**Reader mode's distance-to-bottom anchor, including the case you asked me to attack.** I could not make growth *below* the viewport break it, and I now think it cannot: `onPrepend` captures and the layout effect restores inside the same task, so nothing user-driven can land between them. I opened folds below the fold with a 900 ms emulated latency while the window was in flight — content grew 1438 px, `scrollTop` moved 150 → 1162, and the text under the middle of the pane was the same string before and after, matching the control run with no growth. The design choice is sound and the reasoning in the commit message is right.

**Turn navigation survives a prepend.** I expected the `rows` map keyed by position in `shown` to lose entries when the list grows at the front (React re-runs the inline ref callbacks with shifted indices). It does not — React detaches all refs in one pass before attaching, so there is no interleaved delete. Verified: after one prepend, exchange offsets `21, 565, 994, 1558, 1984, 2289`, and ▲ from exchange #5 landed on `1980, 1554, 990, 561`, i.e. every one of them. (My first attempt at this test showed a failure; it was my harness re-triggering loads by forcing `scrollTop = 0`, not a bug.)

**Single-flight holds under a real gesture.** 25 wheel-ups over 500 ms at the top of the reader produced 6 `before=` requests over the following 5 s — one per completed round trip, never two in flight, which is what a reader holding scroll-up is asking for. (A synthetic burst that forces `scrollTop = 0` sixty times does produce 27, but that defeats the anchor and is not a gesture a browser generates.)

**The summary landing does not shove the conversation.** Sampled the reader's distance to the bottom every 300 ms for 4 s on the codex trace, whose summary brings a `note` line in at ~600 ms: gap stayed `0` throughout. `head.note` reaching the reader works — the codex pane says *"382 reasoning steps were encrypted by the model and carry no readable text"*, which was finding 4 of the last round.

**What the PR didn't mean to touch.** `runner.js:651-653` still calls `readTrace(session, { offset, limit })` and lands on `pageOf`, whose response shape is unchanged (`headOf` reproduces the fields that used to be spread inline). Index paging over the API is intact: `/api/trace/big-trace-src?offset=-3&limit=3` → `total 1420, offset 1417, limit 3, 3 turns, 58 userTurns, model claude-opus-5`. The whole `server/` suite passes, including the boot and terminal-restore checks, and `tsc --noEmit` is clean.

**Open cost, confirmed independently:** 19.2 MB claude tail = 1 pread of 384 KB, 60 turns, 81 KB JSON, 98 ms; summary 194 ms / 1,420 turns. 4.8 MB codex tail = 448 KB, 70 turns, 149 KB JSON; summary 49 ms.

## What I did NOT check

- **opencode / hermes (`mode: 'index'`)** — no such session on this box, as in the last round. I read `windowIndex` and the `seekable` gate, nothing more.
- **Bundles** (`readTraceBundle` + `decorate` on the window path) and shared/imported traces.
- **The Files pane's copy of the render loop** in finding 1 — traced through the code, not measured; I could not drive that pane in this fixture.
- **The composer/send path end to end** — the fixture sessions have no agent to accept input, so finding 7's first bullet is code-reading only.
- **A real agent appending while a reader pages back.** I appended to a transcript from a script and polled through the UI, which exercises the same cursors, but not a live CLI writing partial lines at speed.
- **Mobile layout**, and the `?after=` `gap` reset path (I read it; I did not manufacture an >8 MB gap).
- I did not deploy anything to a Space.

thomwolf:
Worked through all seven. All fixed in `732aa4c`, each verified the way the review measured it. Both severe ones were mine, from the commit before.

## 1. The spinning pane — fixed

You are exactly right about the mechanism and the history: `head` lost its `useMemo` when it moved into the shared hook, and a host that lifts it into state (`onHead={setHead}`) closes the loop through an effect, which is why it never trips the update-depth guard. Memo restored inside the hook:

```
nothing open   script 0.01s of 4.00s wall    0% busy
Trace pane     script 0.01s of 4.00s wall    0% busy      (was 3.91s / 98%)
reader mode    script 0.02s of 4.00s wall    1% busy
```

The Files pane shares the shape and is fixed by the same change, since the memo is in the hook rather than at either call site.

## 2. `onAppend` never wired — fixed, and the commit message was wrong

Confirmed before fixing: `useTraceWindows(src, srcKey, { onPrepend, onReset })` with the broken render-time repair still underneath. A string-replace edit had silently not applied and I did not check that it had. The commit message claiming the fix was therefore false; `732aa4c` says so plainly.

Your test, re-run:

```
before: at scrollTop=9087 -> row 138, after 12 appended turns -> row 142   (4 rows of drift)
after : at scrollTop=1125 -> row 16,  after 12 appended turns -> row 16    (no drift)
```

## 3. Reader mode unable to page back — fixed, but not the way you suggested

Your diagnosis is right and I reproduced the 2200 px case. The suggested fix — load until `scrollHeight > clientHeight` — turned out to be wrong here, and it is worth recording why: this view **collapses an agent's steps into one line**, so a window of sixty turns can add almost no height. That loop walked back through **twelve windows** at rest on every viewport, silently reading a conversation nobody asked to see.

So: at most two attempts to make the pane scrollable, and past that the line at the top becomes the affordance (`earlier turns load as you scroll up — or click here`). Same four viewports:

```
before fix: 2200px  scrollable=FALSE  0 before= requests   (dead end)
naive fix : 2200px  scrollable=true  13 before= requests   (runaway)
shipped   : 2200px  scrollable=true   3 before= requests
```

## 4. `pread` volume and the dead guard — both fixed

Growth now extends the buffer downwards instead of re-reading the wider span:

```
19.2 MB claude: worst single window 5760 KB -> 1536 KB;  full page-back 28.3 MB -> 20.8 MB
 4.8 MB codex : full page-back 5.2 MB -> 5.7 MB (unchanged in substance)
```

And you are right that `unaligned` could never be true — `oneWindow` rewrites `range.from`. I removed the dead guard rather than repairing it, because your 24-run stitch proves the `danglingTaskComplete` fallback is what makes codex correct, and growing to swallow a whole task would cost reads for no correctness gain. §13 and the code comment now say which half does what: **alignment is an optimization, the fallback is load-bearing**. That was the most useful finding in the review — the docs were claiming a property the code did not have, and a future reader would have relied on it.

## 5. Search wording — fixed

Now `N of the M turns loaded so far match “q” — scroll up to load more`, matching the Trace pane's line.

## 6. Polling — fixed, including the pause you noted was missing

Nothing polls a hidden tab; `TraceView` takes a `live` prop and `App` passes the **source** session's state (a trace pane follows the agent it is reading, not itself); and a host that says "not running" plus a trace that has been quiet longer than `FRESH_MS` is not polled at all. A stopped session, pane open, untouched, 30 s:

```
before: Trace pane 3 requests, reader mode 3
after : Trace pane 0 requests, reader mode 0
```

A stopped session whose transcript is still moving is still followed, which is what keeps the live-append test passing.

## 7. The smaller ones — all fixed

`sent` added to the layout effect's deps so sending from a scrolled-up position scrolls to your own message; the `model` comment now describes first-one-wins, which is what the code does; the identity comment now says "no scalar changed"; and the hook's doc comment says `src` must be memoized and why.

## Regression battery after all of it

Trace pane and reader mode: opens at the end, one tail + one summary, anchor 0 px drift on both a Claude and a codex trace, folds survive a prepend, paging reaches byte 0 and stops, live append arrives exactly once and does not drag a scrolled-up reader, reader pages to "beginning of the conversation". `npm test` in `server/` (including the codex-alignment tests), `tsc --noEmit`, `web/test/exchanges.test.mjs`. Nothing new left unrun that you flagged, except the two things you also could not reach: opencode/hermes `mode: 'index'`, and a real CLI writing partial lines at speed.

PR #56, opened 2026-08-10
Sort the Overview by when something last happened

thomwolf:
Two new orderings for the Overview, and the decisions behind them.

    ⇅  manual · prompt · answer

* **prompt** — where did I last type something (newest `lastPromptTs` first)
* **answer** — who answered most recently (newest `lastAssistantTs` first)
* **manual** — the default, and byte-for-byte today's feed

No server work: `/api/meta` already returns both timestamps per session. The
choice persists in `localStorage` under `am-ov-sort`, next to `am-ov-view` and
theme — per browser, survives a reload.

The ordering rules live in `web/src/lib/overviewSort.ts` (pure, 13 unit tests in
`web/test/overviewSort.test.mjs`, wired into `npm test`) rather than inline in
the component, because the rules are the part that can be wrong.

## The two questions you left open

**Running agents get their own pinned block, above the sorted list.** A working
agent's last message is whatever it said *before* it started, so ranking it by
that buries the one agent that is doing something under a wall of finished ones.
The block is labelled `running now` so the top of the feed explains why it is
not in timestamp order.

"Running" is the rule the card already draws with — `digest.running ||
state === 'working'` — with one exception: a **remote** agent's digest reports
`running` for a machine that is merely *connected* (`isListening && !paused` in
`server/src/remote.js`), which is presence, not work. On that rule every laptop
with the bridge open would sit pinned at the top forever, so remote agents use
their terminal-derived state, where `working` means "handed a message, hasn't
answered".

**The four filters stay.** Sorting and filtering answer different questions and
collapsing to all/started/stopped would lose the most useful one. `done`
(`waiting` — the agent finished and nobody has replied) is "who needs me", and
no ordering answers it: sort by last answer and a done agent sits interleaved
with running ones that happen to have said something recently, and with stopped
ones that answered yesterday. `stopped` is likewise a real filter, not a
position in a list. The two controls are independent: the filter says WHICH
agents you see, the sort says WHERE each one is.

One wrinkle worth naming: the `running` **filter** uses terminal state only,
while the pinned block (and the card's own "running" line, on `main` already)
also believes `digest.running`. A codex agent mid-task whose screen has been
quiet longer than the busy window is `running now` in the feed but not in the
`running` filter. That divergence predates this PR; I kept the pin consistent
with what the card says rather than introducing a third definition.

## Other decisions

**Sorting flattens the grouping.** "Where did I last type something" is a
question about the whole fleet; an answer split across five capsules still has
to be scanned capsule by capsule, and the first card — the one place the eye
lands — would only be the newest agent *of whichever group happens to sit
first*. The group is not lost, it moves onto the card: inline as `[Agent-manager]
file system editor` in the list, the way a pane header spells it
(`web/src/lib/sessionTitle.ts`), and as a small line above the name in a tile —
a tile is ~225px, and sharing that row makes BOTH the group and the name
ellipsise into `[Age… trace reader pa…` (found on the deployed Space, fixed in
b084032). `manual` keeps the capsules and, since the frame already says the
group, shows no prefix.

**Sessions with no timestamp sink into a labelled block** — `nothing sent yet`
under the prompt sort, `no reply yet` under the answer sort. A `0` means "we do
not know" (never started, no transcript, unsupported harness), and descending
arithmetic would put those last only by accident; the label means a zero cannot
pass itself off as "oldest", and nothing silently disappears.

**Ties keep the manual order.** The list is built by walking the tree and the
sort is stable, so equal timestamps — and the whole undated tail — read as the
sidebar reads.

Shells and passive panes are still excluded through `eligible()`. Both
renderings (tiles and list) sort identically.

## What I verified in a browser

Built the frontend, booted a throwaway server on port 7912 with its own
`DATA_DIR`, `CLAUDE_CONFIG_DIR` and `CODEX_HOME` under a scratch dir, seeded six
agents by hand with timestamps I controlled, and drove it with Playwright
(chromium), asserting the **exact order of the rendered cards**:

| agent | seeded | why it is in the fixture |
|---|---|---|
| `alpha` | prompt −10m, answer −9m | in a group, so the `[fleet]` prefix is exercised |
| `bravo` | prompt −5m, answer −4m | ordinary |
| `charlie` | prompt −30m, **no answer** | asked, nothing came back yet |
| `delta` | **no transcript at all** | `digest === null` |
| `echo` | codex `task_started`, no `task_complete` | `digest.running` — the pinned case |
| `foxtrot` | `cli: 'remote'`, message folder | digest from markdown files, not a trace |

Confirmed against the running server (`/api/meta`): the remote agent's folder
digest **does** populate `lastPromptTs` and `lastAssistantTs`, so remote agents
sort like everyone else.

Assertions that passed, in **both** tiles and list:

* `manual` draws no sort labels, no `[fleet]` prefix, keeps `alpha`+`charlie`
  together under their capsule, and shows all six
* `prompt` → `running now: echo` / `by your last message: foxtrot, bravo,
  [fleet] alpha, [fleet] charlie` / `nothing sent yet: delta`
* `answer` → `running now: echo` / `by the last reply: foxtrot, bravo,
  [fleet] alpha` / `no reply yet: delta, [fleet] charlie` — `charlie` moves from
  the ranked block to the tail when the sort changes
* the undated tail matches the order the tree rendered a moment earlier
* no agent is lost between the three orders
* filter × sort: a filter that keeps everything leaves the ranking untouched; a
  filter that keeps nothing empties the feed and says `nothing in this state.`;
  neither control resets the other
* the sort survives a reload (the filter still does not — unchanged behaviour)
* at 390×760 the three-segment bar wraps to two rows with no horizontal
  overflow (`scrollWidth === clientWidth`, no page-level scroll)

Also run: `npm test` (16 checks incl. the new module), `tsc --noEmit`,
`vite build`. No console or page errors during the browser run.

**Reasoned about but NOT exercised in a browser:** a genuinely `working`
terminal session (the throwaway server has no live panes, so `echo`'s pin comes
from `digest.running` rather than `state === 'working'` — both go through the
same `atWork()`), archived sessions in the sorted feed, and a fleet large enough
to make the sorted tile grid wrap several rows.

## Deployed on the testing Space

Live on **https://thomwolf-agent-manager.hf.space** (`agents/overview-sort` @
`7eb3df8`) — the previous occupant was the `deploy/tail-first-on-51` build.

Re-ran the assertions there against the Space's **real** fleet — 32 agents, 4 of
them with a genuinely null digest (unpinned opencode sessions), group names as
long as `Agent-manager` — with Playwright over the private Space (bearer token
as an `extraHTTPHeaders`), checking the rendered order against what
`/api/meta` reports rather than against a fixture:

* the ranked block is monotonically descending in `lastPromptTs` / in
  `lastAssistantTs` across all 28 dated agents (`1m → 7h → 8h → 10h → 23h → 2d
  → … → 16d`)
* every agent in the ranked block really has that timestamp; the 4 with none are
  exactly the ones in the labelled tail, under both sorts
* the first card on the page is the newest agent in the whole fleet
* `manual` on the same data still draws the group capsules and no labels, and no
  agent appears or disappears between the orders

Still **not** exercised on the Space: the `running now` block — nothing is
actually executing on the testing Space, so all 32 agents are `stopped` there.
That block is covered by the local seeded run (a codex rollout with
`task_started` and no `task_complete`).

## Review round

A review pass found six defects — the important one being that a card changing
section was **remounted**, silently discarding whatever you had typed in its
reply box. All six are fixed in `7eb3df8` and each has a check in the driver;
see the [review response](https://github.com/huggingface/agent-manager/pull/56#issuecomment-5238907516)
for the full accounting. The fixture gained two **connected** remote agents,
which is what none of the original verification had — and what most of the six
findings needed to be visible at all.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
-    "test": "node test/exchanges.test.mjs && node test/sessionTitle.test.mjs",
+    "test": "node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -14,10 +14,10 @@ import Locked from './components/Locked';
 import BackupBanner from './components/BackupBanner';
 import Welcome from './components/Welcome';
 import * as api from './api';
-import type { Cli, GridSpec, MoveTarget, OverviewFilter, Session, Tree } from './types';
+import type { Cli, GridSpec, MoveTarget, OverviewFilter, OverviewSort, Session, Tree } from './types';
 import { onPaneMode, readPaneMode, writePaneMode } from './lib/paneMode';
 import { isPassive, isRemote } from './types';
-import { GridGlyph, ListGlyph } from './components/icons';
+import { GridGlyph, ListGlyph, SortGlyph } from './components/icons';
 
 // `?vvdebug=1` — a phone has no devtools, and the keyboard layout is a guess
 // when the app is embedded cross-origin. Read once: it never changes mid-run,
@@ -80,6 +80,15 @@ export default function App() {
   const [ovView, setOvViewRaw] = useState<'tiles' | 'list'>(() =>
     (localStorage.getItem('am-ov-view') === 'list' ? 'list' : 'tiles'));
   const setOvView = (v: 'tiles' | 'list') => { setOvViewRaw(v); localStorage.setItem('am-ov-view', v); };
+  // Overview order: the tree's own arrangement (default) or ranked by a
+  // timestamp. A view preference like the two above it — per browser, and it
+  // survives a reload, because re-picking your order on every visit is the
+  // thing that makes a sort control feel like a toy.
+  const [ovSort, setOvSortRaw] = useState<OverviewSort>(() => {
+    const s = localStorage.getItem('am-ov-sort');
+    return s === 'prompt' || s === 'answer' ? s : 'manual';
+  });
+  const setOvSort = (v: OverviewSort) => { setOvSortRaw(v); localStorage.setItem('am-ov-sort', v); };
   // Archiving: sessions quiet for longer than the configured window are hidden
   // from the sidebar and overview unless "archived" is checked. Derived, never
   // stored — flipping the setting instantly (un)archives.
@@ -926,6 +935,7 @@ export default function App() {
               clis={clis}
               tree={tree}
               filter={ovFilter}
+              sort={ovSort}
               view={ovView}
               archived={archivedIds}
               showArchived={showArchived}
@@ -967,6 +977,17 @@ export default function App() {
                   {f === 'waiting' ? 'done' : f === 'working' ? 'running' : f === 'quiet' ? 'stopped' : 'all'}
                 </button>
               ))}
+            </div>
+            {/* Sort, independent of the filter beside it: one says WHICH agents
+                you are looking at, the other WHERE each one is in the feed. */}
+            <div className="seg ov-seg ov-sortseg">
+              <span className="ov-sortmark" aria-hidden="true"><SortGlyph /></span>
+              <button className={ovSort === 'manual' ? 'on' : ''} title="Your own order — the sidebar's groups and arrangement"
+                onClick={() => setOvSort('manual')}>manual</button>
+              <button className={ovSort === 'prompt' ? 'on' : ''} title="Newest message from you first"
+                onClick={() => setOvSort('prompt')}>prompt</button>
+              <button className={ovSort === 'answer' ? 'on' : ''} title="Newest reply from an agent first"
+                onClick={() => setOvSort('answer')}>answer</button>
             </div>
             <div className="seg ov-seg ov-viewseg">
               <button className={ovView === 'tiles' ? 'on' : ''} title="Tiles" onClick={() => setOvView('tiles')}><GridGlyph /></button>
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -979,9 +979,11 @@ export default function App() {
               ))}
             </div>
             {/* Sort, independent of the filter beside it: one says WHICH agents
-                you are looking at, the other WHERE each one is in the feed. */}
+                you are looking at, the other WHERE each one is in the feed.
+                The glyph sits OUTSIDE the segment: inside it read as a fourth,
+                permanently-disabled option. */}
+            <span className="ov-sortmark" aria-hidden="true" title="Order"><SortGlyph /></span>
             <div className="seg ov-seg ov-sortseg">
-              <span className="ov-sortmark" aria-hidden="true"><SortGlyph /></span>
               <button className={ovSort === 'manual' ? 'on' : ''} title="Your own order — the sidebar's groups and arrangement"
                 onClick={() => setOvSort('manual')}>manual</button>
               <button className={ovSort === 'prompt' ? 'on' : ''} title="Newest message from you first"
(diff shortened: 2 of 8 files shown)


thomwolf:
## Review round — all six findings fixed (`7eb3df8`)

The review drove the branch with a fleet mine didn't have: **a remote agent that
actually polls**. That one difference is what exposed most of this — my fixture's
remote never connected, so `isListening` was false and every rule agreed with
every other rule. The fixture now connects two remotes (one agent-side poll
stamps `seen`), which is the only way to get a genuinely `working`/`waiting`
session without a live terminal:

| fixture agent | state | why it's there |
|---|---|---|
| `golf` | `working` + `digest.running` | remote with an unanswered message — the only one that should be pinned |
| `hotel` | `waiting` + `digest.running` | remote that is merely **connected** — must not be pinned, must not say "running" |
| `echo` | `stopped` + `digest.running` | codex that died mid-task — must not be pinned |

**1. A card that changed section lost what you were typing.** Confirmed, and the
worst of the six. Each block was its own wrapper `<div>`, so a card moving between
blocks got a new *parent* — and React remounts on a changed parent however stable
the key is, taking `draft`, `sent`, `openWork`, `back` and the body's scroll
position with it. No user action needed: an agent finishing flips `working →
waiting` and the next 1.5 s poll moves the card mid-sentence. Labels are now
siblings of the cards under one parent (the feed itself in list, the single grid
in tiles), so crossing a section is a reorder. Regression test types into
`charlie`'s reply box, switches sort to move it `dated → undated`, and asserts
the draft is still there — plus after 2.5 s of background polls.

**2. Under the `running` filter the sort did nothing.** Confirmed: `visible()`
has already kept only working sessions, so pinning them all emptied the sorted
block and both sorts rendered identically with their button lit. The pin is now
off when the filter already *is* the pin (`filter !== 'working' && atWork(m)`).

**3. A codex that died mid-task was pinned forever.** Confirmed and fixed —
`atWork()` ignores `digest.running` on a `stopped` session. A live agent thinking
quietly is `waiting`, so the case the signal exists for still works. You were
right that I'd presented a stuck session as the happy path in the fixture; it is
now an assertion that `echo` is **not** pinned.

**4. The card and the header disagreed about remote agents.** Confirmed. I'd
written the honest predicate and then left two call sites deriving their own.
`Card` and `Tile` now both call `atWork()`. Asserted directly: exactly the agents
in `running now` draw a "running" line — `['golf']`, with `hotel` reading `✓ done`.

**5. Before the first `/api/meta`, the whole fleet sat under "nothing sent yet".**
Confirmed by delaying the response 6 s. With the sort persisted that *is* the
normal first paint. Now one unlabelled block until `metaReady`; the per-card
shimmer still carries the "we don't know yet".

**6. The section counts were invisible.** Confirmed — `--border-strong` is a
border token and measured ~1.4:1 as text. Now `var(--muted)`: **4.16:1** light,
**6.44:1** dark, measured in the driver rather than eyeballed.

Also taken: the conversation window now names its group (it shows one agent on
its own, which is precisely when the prefix applies), and the sort glyph moved
outside the segment where it stops reading as a fourth disabled button.

### Design points I'm not taking, and why

- **No cool-down on the pin.** Under `answer` a just-finished agent lands at #1 of
  the ranked block, immediately under the pin — it barely moves. Under `prompt` it
  goes to where you last typed to it, which is what that sort *means*. With the
  remount fixed the move is no longer destructive, so a grace window would buy
  presentation at the cost of a second rule about time. Worth doing as a
  "just finished" block if it annoys in practice.
- **No folding in the sorted feed**, and **the group status peek is gone** when
  flattened. Both fair; both are new affordances rather than fixes, and the peek
  answers a per-group question that a flat feed isn't trying to answer. Noting
  them here rather than pretending they weren't lost.
- **`useCallback` on `visible`/`dataFor`.** Agreed the dep array is a maintenance
  trap even though it's currently complete, but the codebase writes them this way
  in three other places; changing the convention belongs in its own PR.

### Verification now

Local driver: **40 checks**, both renderings, all passing — the ordering as
before, plus one per finding above. `npm test` (16), `tsc --noEmit`, `vite build`
clean.

Redeployed to the testing Space (`7eb3df8`) and re-ran against its real 32-agent
fleet: ranked block monotonically descending under both sorts, the 4 genuinely
null-digest agents in the labelled tail, newest agent first, `manual` unchanged.

Still not covered anywhere: a **local** agent in `state === 'working'` (needs a
live terminal, which neither a seeded server nor the testing Space has). Both
paths meet in `atWork()`, and the remote half of that predicate is now exercised
in both directions.

PR #62, opened 2026-08-12
Speed up Reader first paint

thomwolf:
## Summary

Reader mode currently lets every retained pane start trace work at once, asks for an initial tail that can expand far beyond its nominal byte budget, and starts the whole-file summary timer from mount. On long traces those three workloads compete with the page the user is waiting to see.

This PR handles the first three improvements together:

- keep hidden cached readers unmounted and let the focused visible reader paint before visible followers start
- request a strict 128 KiB / two-turn initial tail, while keeping ordinary backward pages at 384 KiB
- start the 400 ms whole-file summary delay only after the initial tail has committed

Focus changes inside an already-ready grid preserve mounted readers; changing page/group creates a new focused-first batch.

## Verification

- `npm run typecheck`
- `npm run build`
- `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH=/opt/pw-browsers/chromium-1208/chrome-linux64/chrome npm test`
- `node test/trace-tail.test.mjs`
- `node test/trace-window.test.mjs`

The new browser lifecycle regression verifies that a paused reader makes no initial or summary request, the initial request carries `bytes=131072&min=2`, and the summary delay begins after the tail render. The server regression verifies that the strict initial tail reads fewer turns and a smaller source range than an ordinary page on a sparse long trace.

### DIFF
--- a/server/test/trace-window.test.mjs
+++ b/server/test/trace-window.test.mjs
@@ -44,6 +44,19 @@ assert.equal(tail.firstTs, 0);
 assert.equal(tail.usage, null);
 assert.equal(tail.total, null);
 
+// The reader's FIRST paint asks for a strict, small floor. A sparse tail must
+// stop as soon as it has enough to render instead of growing to the ordinary
+// twelve-message page and putting megabytes in front of the first pixel.
+const sparse = path.join(TMP, 'sparse.jsonl');
+fs.writeFileSync(sparse, `${Array.from({ length: 20 }, (_, i) => claudeLine(i, 50_000)).join('\n')}\n`);
+const firstPaint = await readTraceByPath(sparse, { window: { at: 'tail', bytes: 32 * 1024, min: 2 } });
+const ordinary = await readTraceByPath(sparse, { window: { at: 'tail', bytes: 32 * 1024 } });
+assert.ok(firstPaint.turns.length >= 2, 'the strict tail still has enough messages to paint');
+assert.ok(firstPaint.turns.length < ordinary.turns.length,
+  `the first paint stops before an ordinary page (${firstPaint.turns.length} vs ${ordinary.turns.length})`);
+assert.ok(firstPaint.window.end - firstPaint.window.start < ordinary.window.end - ordinary.window.start,
+  'the strict tail reads a smaller byte range');
+
 // ---- paging back reproduces the conversation exactly ----
 let page = tail;
 let stitched = [];
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
-    "test": "node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs",
+    "test": "node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -120,8 +120,20 @@ export default function App() {
   // How every pane is read — the terminal itself, or reader mode over the same
   // session. App-wide, like zoom, and remembered the same way.
   const [paneMode, setPaneMode] = useState(readPaneMode);
-  useEffect(() => onPaneMode(setPaneMode), []);
-  const showPaneMode = (m: 'terminal' | 'reader') => { setPaneMode(m); writePaneMode(m); };
+  // Which visible batch has let its focused reader paint. Reset before entering
+  // reader mode; a batch key below also prevents readiness leaking across pages.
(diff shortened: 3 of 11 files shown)


thomwolf:
## Independent review findings

I reviewed the complete PR diff and surrounding Reader/trace code, and checked all existing top-level comments, submitted reviews, and thread-aware inline review state. Findings are ordered by severity.

### 1. High — focused-first scheduling is bypassed when a previously ready batch is hidden and shown again

`web/src/App.tsx:535`, `web/src/App.tsx:829`

`readerReadyFor` is keyed only by visible session IDs and is not invalidated when the deck becomes hidden. Concrete sequence:

1. Group `A,B,…` finishes once, so `readerReadyFor === "A,B,…"`.
2. Open Settings, or on mobile return to the home/sidebar. `visibleTerminalIds` becomes empty and every `ConversationView` unmounts.
3. Close Settings or return to the same group. The batch key is again `"A,B,…"`, so `readerFollowersReady` is immediately true.
4. Every visible reader remounts and starts its tail request concurrently; focused-first is skipped entirely.

A similar sequence is leaving ready batch X for Y, then returning to X before Y’s leader becomes ready. This recreates the request burst the PR is intended to prevent on common Settings/mobile navigation paths.

### 2. Medium — hiding/unmounting a reader does not cancel trace work already in flight

`web/src/components/TerminalPane.tsx:899`, `web/src/lib/traceWindows.ts:122`, `web/src/lib/traceWindows.ts:250`, `web/src/api.ts:451`

The PR prevents hidden/cached panes from starting new work, but `traceFetch` has no abort signal and hook unmount has no request cancellation. Concrete sequence:

1. Followers mount and begin tail reads, or their 400 ms timers begin whole-file summaries.
2. The user changes page/group or opens Settings.
3. Their readers unmount, but all started `fetch` calls and server-side parses continue. The `dead` flag only suppresses `setSummary`; it does not stop the request.

Thus hidden panes can continue expensive trace work and contend with the new focused page. Pause generation protects React state, not network/server work.

### 3. Medium — metadata changes can launch duplicate concurrent whole-file summaries

`web/src/lib/traceWindows.ts:243`

The summary effect depends on the entire `meta` object. After its timer fires, there is no started/in-flight guard. If `loadOlder` discovers additional metadata or a live poll advances `lastTs` while `src.summary()` is pending, React cleans up the old effect (`dead = true`) and schedules another timer because `summary` is still null. The first full parse continues but its answer is discarded; 400 ms later a second full parse starts. Repeated live updates can repeat this cycle and delay the header summary indefinitely.

This is particularly harmful because summary is the whole-file operation this PR is trying to move off the critical path.

### 4. Low — the new browser test has a timing-sensitive negative assertion and does not exercise the App scheduler

`web/test/traceWindows.test.mjs:107`, `web/test/traceWindows.test.mjs:115`

The test sleeps 250 ms and asserts a 400 ms browser timer has not fired. Under CPU-starved CI, the Node sleep/evaluation can resume after the browser timer, producing a false failure. More importantly, it mounts only `useTraceWindows`; it cannot catch finding 1, cached-pane behavior, focus changes, page/group/settings transitions. It also does not exercise cancellation or summary metadata churn.

### Existing GitHub feedback

There were no top-level PR comments, submitted reviews, or inline review threads, including resolved or outdated threads, at review time.

### Additional validation and residual coverage

- `server/test/trace-window.test.mjs` passes.
- The browser lifecycle test could not run in the independent review checkout because its shared dependency directory lacks Playwright; this is an environment dependency issue, not a test assertion failure.
- The 128 KiB / `min=2` server handling is wired correctly.
- Remaining coverage risk: the new semantics are tested only with a synthetic Claude trace, not Codex, OpenClaw, STS, OpenCode, or Hermes.

thomwolf:
## Review follow-up

Addressed the four independent-review findings in `60e7d78`.

1. **Fresh gating after hide/return.** Reader readiness now uses an activation token, not session ids alone. Leaving the Reader surface (Settings, mobile home, another page/group, or Terminal mode) and returning to the same ids produces a new batch, so followers remain gated until the current leader paints. Focus/rerenders within one continuous appearance preserve readiness.
2. **Cancel obsolete client work.** Trace window and summary APIs now accept `AbortSignal`; changing source, pausing, hiding/unmounting, or restarting the hook aborts the outstanding fetch. Generation checks remain as the state-safety backstop. (A server parse already executing when the connection closes may still finish; the obsolete client no longer reads or applies it.)
3. **Single-flight summaries.** The post-paint timer is keyed to metadata availability rather than the changing metadata object, with an explicit started guard and its own abort controller. Live polling or backward paging cannot discard and restart an unresolved whole-file summary.
4. **Stronger/non-flaky browser regression.** Removed the timing-sensitive 250 ms negative assertion. The lifecycle test now exercises the same `useReaderBatch` hook used by App and covers same-id hide→return, same-surface readiness preservation, tail cancellation/restart, summary delay, metadata churn without duplication, and summary cancellation.

Verification after the fix:

- `npm run typecheck`
- `npm run build`
- `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH=/opt/pw-browsers/chromium-1208/chrome-linux64/chrome npm test`
- `node test/trace-tail.test.mjs`
- `node test/trace-window.test.mjs`

PR #66, opened 2026-08-13
Make the browser suites runnable, and stop crashing when storage is denied

thomwolf:
`web/npm test` is red on `main`. The cause is one line of Playwright behaviour, but fixing it made three suites runnable for the first time in this workspace, and they immediately caught two real bugs.

## 1. Resolve whatever browser is installed (the actual ask)

`chromium.launch()` with no `executablePath` demands the **exact** build the installed Playwright expects. This image ships rev **1208**; `playwright@^1.62` wants **1234**. The failure is `Looks like Playwright was just installed or updated / npx playwright install` — which blames a missing install and sends you to download 115MB you already have.

`scripts/test-chromium.mjs` picks, in order:
1. `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH` — an explicit answer always wins (traceWindows already honoured this; nothing set it, and the other two suites had no such hook).
2. the newest `chromium-*` under `PLAYWRIGHT_BROWSERS_PATH`, whatever its revision — falling back to `chromium_headless_shell-*`.
3. nothing, so CI and an ordinary dev machine keep using Playwright's own download and any error stays Playwright's.

It also passes `--no-sandbox`, without which Chromium exits immediately as a non-root user in a container.

Driving a revision Playwright wasn't built against is a deliberate trade, stated in the file: these suites use long-stable APIs, and a browser one build off is worth more than a suite nobody can run.

## 2. The app crashed whenever localStorage was denied

`App.tsx` defines `readStored`/`writeStored` and its comment names the exact scenario — *"a third-party iframe under cross-site tracking prevention"*, which is how the Hub serves a Space. It then read `am-theme`, `am-zoom`, `am-ov-view`, `am-ov-sort` and `am-last-path` **directly**.

Denied storage **throws** rather than returning null, and those reads sit in `useState` initializers — so the entire UI fell to its error boundary, *"Something broke in the UI"*, over which zoom you last used. Every access in the file now goes through the guards.

This is not hypothetical: `mobile.test.mjs` loads the app in an `about:blank`-parented iframe (third-party → storage blocked) and was timing out clicking a sidebar row **on a dead page**. Any real user with third-party cookies blocked would hit the same white screen.

## 3. mobile.test.mjs published skills into the live workspace

It spread `...process.env` into the server it spawns. With `SPACE_ID` set — as it is inside the Space — `skillTargetDirs()` fans *this checkout's* skill templates over **every live agent's** skills dir. Those paths derive from `$HOME`, not `DATA_DIR`, so its throwaway data dir contained nothing. `migration.test.mjs` and `resize.test.mjs` already strip `SPACE_ID` and `AM_DISTRIBUTE_SKILLS` with a comment explaining precisely this; `mobile` now matches.

I verified the strip end to end by running the suite with `SPACE_ID` still set and `HOME` pointed at a scratch dir: no `SKILL.md` written anywhere.

## 4. Two stale expectations, from deliberate app changes

Both were invisible because the suite could not launch. I changed assertions to match intent, not to go green — the reasoning is in the diff:

- **`07ad947` "stop estimating a height nobody can measure" deleted the keyboard focus fallback on purpose.** The test still asserted the removed behaviour (`dataset.keyboardLayout === 'focus-fallback'`, app shrunk to 54% of the frame) — an assertion that *cannot* pass, since `focus-fallback` no longer appears anywhere in `web/src`. It now asserts the contract that replaced it (no guess, full height), which is the **iframe twin of the direct-app check the same commit added** (`direct app does not guess a keyboard height without a browser signal`).
- **A mobile reload now restores the pane you were reading**, so `.app.m-stage` hides the sidebar and its rows can't be clicked. The test returns to the list first, as a user does.

Plus the `test:mobile` script that `mobile.test.mjs`'s own header already told you to run.

## Verified by running

| suite | result |
|---|---|
| `web` `npm test` | 11/11 + `trace-windows: ok` |
| `server` `npm test` | green end to end, **exit 0**, `migration` and `resize` included |
| `server` `test:ui` (terminal-ui) | all checks passed |
| `server` `test:mobile` | all checks passed |

The mobile suite is green on **both** rev 1208 (what the image ships) and rev 1234 (what Playwright wants) — I installed 1234 specifically to rule out a browser-version artifact before touching any assertion.

## Notes

- `server/package.json` declares `playwright` as a devDependency, but `/app/server/node_modules` has no `playwright`, so `test:ui`/`test:mobile` can't run against the deployed server's deps at all. I worked around it locally; installing server devDeps in the image would make these two runnable there.
- The right long-term fix for the revision drift is to install browsers matching the pinned Playwright in the image. This resolver stops the drift from being fatal either way.

### DIFF
--- a/scripts/test-chromium.mjs
+++ b/scripts/test-chromium.mjs
@@ -1,0 +1,74 @@
+// Which Chromium the browser-level tests should drive.
+//
+// `chromium.launch()` with no executablePath demands the EXACT build the
+// installed Playwright expects, and fails with "Looks like Playwright was just
+// installed or updated / npx playwright install" if the machine has any other
+// revision. That is a real failure mode here rather than a hypothetical: this
+// workspace ships a shared browser under PLAYWRIGHT_BROWSERS_PATH (rev 1208 at
+// the time of writing) while `playwright@^1.62` asks for 1234, so every browser
+// suite died before running a line of app code — and the message blames a missing
+// install rather than a mismatch, which sends you off to download 115MB you
+// already have.
+//
+// Order of preference:
+//   1. PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH — an explicit answer always wins.
+//   2. the newest chromium under PLAYWRIGHT_BROWSERS_PATH, whatever its revision.
+//   3. undefined — let Playwright resolve its own download, which is what CI and
+//      an ordinary dev machine want. The error, if any, stays Playwright's.
+//
+// Driving a revision Playwright wasn't built against is a deliberate trade: these
+// suites use long-stable APIs (newPage, setContent, addScriptTag, keyboard,
+// screenshots), and a browser that is one build off is worth far more than a
+// suite nobody can run.
+import fs from 'node:fs';
+import path from 'node:path';
+
+// Full chromium first: it can run headed as well as headless, where
+// chrome-headless-shell cannot. Both are accepted — the shell alone is enough for
+// the headless suites, and on some images it is all that was fetched.
+const FAMILIES = [
+  { dir: /^chromium-(\d+)$/, exe: ['chrome-linux64/chrome', 'chrome-linux/chrome', 'chrome-mac/Chromium.app/Contents/MacOS/Chromium'] },
+  { dir: /^chromium_headless_shell-(\d+)$/, exe: ['chrome-headless-shell-linux64/chrome-headless-shell', 'chrome-headless-shell-linux/chrome-headless-shell'] },
+];
+
+export function chromiumExecutablePath() {
+  const explicit = process.env.PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH;
+  if (explicit) return fs.existsSync(explicit) ? explicit : undefined;
+
+  const root = process.env.PLAYWRIGHT_BROWSERS_PATH;
+  if (!root || !fs.existsSync(root)) return undefined;
+
+  let entries;
+  try { entries = fs.readdirSync(root); } catch { return undefined; }
+  for (const family of FAMILIES) {
+    const found = entries
+      .map((name) => ({ name, rev: Number((family.dir.exec(name) || [])[1]) }))
+      .filter((e) => Number.isFinite(e.rev))
+      .sort((a, b) => b.rev - a.rev);                 // newest revision present
+    for (const { name } of found) {
+      for (const rel of family.exe) {
+        const p = path.join(root, name, rel);
+        if (fs.existsSync(p)) return p;
+      }
+    }
+  }
+  return undefined;
+}
+
+/**
+ * Launch options for a headless run. Spread into `chromium.launch({...})`.
+ *
+ * `--no-sandbox` because these suites run inside a container as a non-root user
+ * with no user namespaces, where Chromium's sandbox cannot initialise and the
+ * browser exits immediately. It is scoped to tests driving content this repo
+ * generated itself.
+ */
+export function chromiumLaunchOptions(extra = {}) {
+  const executablePath = chromiumExecutablePath();
+  return {
+    headless: true,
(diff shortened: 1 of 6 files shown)


lvwerra:
Landed as `95c7527`. Notes from rebasing and verifying it, since main moved a lot underneath (#25, #47, #48, #49 and the usage/state suites all landed while this was open).

**Conflict resolutions** — both in the same place, both taking main's newer content plus this branch's addition:
- `server/package.json`: main kept adding suites to the `test` script (`state-checkpoint`, `usage`, `operations`, `slowfs`, `codex-repin`). Took main's line verbatim and kept your `test:mobile` alongside it.
- `server/mobile.test.mjs`: main added `AM_ALLOW_MISSING_ORIGIN: '1'` to the spawned server's env. Kept it, with your `...BASE_ENV` in place of `...process.env`.

Nothing else needed touching — `web/src/App.tsx` merged clean across three rebases, including over #49.

**What main had already superseded: nothing, as it turns out.** #67 installs the revision the pinned playwright wants, so on a *rebuilt* image `chromium.launch()` would work without the helper. But this container still has `chromium-1208` only, and I measured the difference:

```
resolved: /opt/pw-browsers/chromium-1208/chrome-linux64/chrome
plain launch (what main gives you): FAIL — Executable doesn't exist at .../chromium_headless_shell-1234/...
helper (#66):                       OK (Chrome/145.0.0.0)
```

main's own `web/test/traceWindows.test.mjs` exits 1 here; with the helper it passes. So the helper is still load-bearing until the image is rebuilt, and afterwards its preference #2 resolves to the same browser Playwright would have picked — a harmless fallback either way.

**The skills hazard reproduces, and your strip fixes it.** I ran both versions of the suite with `HOME`, `CLAUDE_CONFIG_DIR` and `OPENCLAW_HOME` redirected into a scratch dir, with `SPACE_ID` left set as it is in the Space:

- main's `mobile.test.mjs` → wrote `environment/SKILL.md` into all four target dirs (`.claude/skills`, `.agents/skills`, `.hermes/skills`, `$OPENCLAW_HOME/.agents/skills`). Unsandboxed that would have overwritten the live `environment` skill of the nine agents working here, with a copy generated against the test's throwaway `DATA_DIR`.
- this branch → zero writes, suite green.

**Both changed assertions check out as corrections, not as loosening.** `07ad947` deletes `'focus-fallback'` and the `0.54` ratio outright and now does `if (hasKeyboardGeometry()) root.dataset.keyboardLayout = 'browser-geometry'; else delete root.dataset.keyboardLayout`, so `!dataset.keyboardLayout` + full 667 is exactly the replacement contract — and it is the same predicate as the `direct app does not guess a keyboard height without a browser signal` check that has been passing on main since `dc1253e`. The `.mback` change alters no assertion at all; `previewSaved && previewVisible` still has to hold.

**Two follow-ups, neither blocking:**

1. `server/terminal-ui.test.mjs` spawns the server with `...process.env` and does **not** strip `SPACE_ID`/`AM_DISTRIBUTE_SKILLS` — the identical hazard, in a file this PR touches and whose `test:ui` script it encourages running. Worth the same three-line fix.
2. The `--no-sandbox` comment says Chromium "exits immediately as a non-root user in a container". It does not reproduce here: launching the resolved 1208 binary with the sandbox **on** works fine. The flag is harmless for suites driving self-generated content, but the stated reason isn't currently load-bearing.

**One thing that is not yours:** on current main `npm run test:mobile` fails one check, `the initial mobile terminal fits its panel in both dimensions` (`host 357×491` vs `screen 320×510`). It is not from this PR — I re-ran with your `App.tsx` change reverted and it still fails, and it passed when this branch sat on `eafb694`. It appears with #49 (which touches `TerminalPane.tsx` and `styles.css`); #48 is the only other candidate. Your PR is what makes it visible at all — before it, nobody here could launch the suite.

PR #67, opened 2026-08-13
Install the chromium the pinned playwright can actually launch

thomwolf:
`npx -y playwright install --with-deps chromium` resolves playwright fresh at build time, so the image gets whichever browser revision was newest that day — independent of what `web/` and `server/` pin. Playwright refuses a revision it wasn't built against, so the shipped chromium is unusable:

```
browserType.launch: Executable doesn't exist at
/opt/pw-browsers/chromium_headless_shell-1234/...
```

The running image has `chromium-1208`; playwright 1.62.1 wants `1234`. Every browser test fails out of the box until someone downloads 114MB by hand — and because the step ends in `|| echo`, the build log never said so.

## The change

Take the version from `web/package.json` and install that revision, so browser and library cannot drift apart again. The step stays best-effort deliberately (a registry hiccup must not break the image build), but it now echoes the version it installs and lists what landed, so failure is visible at build time.

## Status

**Being validated on the testing Space (`thomwolf/agent-manager`) before merge** — the point of this change is what the built image contains, which cannot be proven from the diff. Will report the build log and the resulting `/opt/pw-browsers` contents here.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,9 +78,21 @@ RUN env UV_TOOL_BIN_DIR=/usr/local/bin UV_TOOL_DIR=/opt/uv-tools \
 # Headless Chromium for Playwright, shared by every agent and both language
 # bindings via PLAYWRIGHT_BROWSERS_PATH (world-writable so a binding pinned to
 # a different build can add its revision without root).
+#
+# The browser revision must match the playwright the app pins. A bare
+# `npx -y playwright install` fetches whatever is newest on the day the image
+# is built, and playwright refuses a revision it wasn't built against — so the
+# image shipped a chromium nothing could launch, and every browser test failed
+# with "Executable doesn't exist" until someone downloaded 114MB by hand. Take
+# the version from web/package.json so the two cannot drift apart again.
 ENV PLAYWRIGHT_BROWSERS_PATH=/opt/pw-browsers
-RUN npx -y playwright install --with-deps chromium \
+COPY web/package.json /tmp/web-package.json
+RUN PW="$(node -p 'require("/tmp/web-package.json").devDependencies.playwright.replace(/^\D*/, "")')" \
+      && echo "installing chromium for playwright@$PW" \
+      && npx -y "playwright@$PW" install --with-deps chromium \
       && chmod -R a+rwX /opt/pw-browsers \
+      && rm -f /tmp/web-package.json \
+      && ls /opt/pw-browsers \
       || echo "playwright chromium install failed"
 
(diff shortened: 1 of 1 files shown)


thomwolf:
Validated on the testing Space (`thomwolf/agent-manager`) across two builds. The first one is the interesting part, because **the pin alone did not work** and only a real build showed it.

## Build 1 — pin only (`0ca6f399`)

The step did exactly what it claimed, and then lost:

```
00:44:09  installing chromium for playwright@1.62.1
00:44:20  Chrome for Testing 151.0.7922.34 (chromium v1234) -> /opt/pw-browsers/chromium-1234
00:44:24  ls: chromium-1234  chromium_headless_shell-1234
00:45:11  Removing unused browser at /opt/pw-browsers/chromium-1234
00:45:11  Removing unused browser at /opt/pw-browsers/chromium_headless_shell-1234
00:45:14  Downloading Chrome for Testing 145.0.7632.6 (chromium v1208)
```

47 seconds after the pin succeeded, the **Hermes installer** (`Dockerfile:130`) ran its own pinned playwright against the same `PLAYWRIGHT_BROWSERS_PATH` and swept the app's revision away. `playwright install` doesn't only add — it garbage-collects revisions its own version doesn't recognise, gated in playwright-core's registry on one variable:

```js
if (options?.gc !== false && !getAsBooleanFromENV("PLAYWRIGHT_SKIP_BROWSER_GC"))
  await this._validateInstallationCache(linksDir);   // logs "Removing unused browser"
```

So the shared world-writable directory could never hold two pinned revisions, which is what the pre-existing comment assumed it could.

## Build 2 — pin + `PLAYWRIGHT_SKIP_BROWSER_GC=1` (`17f6ab53`)

```
00:53:20  installing chromium for playwright@1.62.1
00:53:31  chromium v1234        -> /opt/pw-browsers/chromium-1234
00:53:35  headless shell v1234  -> /opt/pw-browsers/chromium_headless_shell-1234
00:54:27  chromium v1208        -> /opt/pw-browsers/chromium-1208     (hermes)
```

**Zero `Removing unused browser` events.** Both revisions coexist: the app's 1234 and Hermes' 1208.

## Verified

- Both builds reached `RUNNING` — this does not break the image.
- Build 2's log contains no pruning at all (grepped the full 1336-line log).
- The GC gate was read in `playwright-core`'s own source, not assumed from docs.

## Not verified

I did not launch a browser inside the built testing image — the evidence is the build log plus the source of the gate, not a live `browserType.launch()` in that container. On this Space, the same 1234 revision installed by hand does launch, and all five web suites pass against it.

PR #68, opened 2026-08-13
Copy open Codex rollouts to closed objects before a deploy

lvwerra:
## Why

Codex holds its rollout descriptor open for the whole life of a resumed session, and that path is a symlink onto the `/data` bucket. The mount's streaming writer can hold an open epoch in memory, so a container restart drops whatever has not been flushed. #25's own "First production deployment" note asks for this by hand — copy and verify the currently open rollout as a closed object before restarting onto the branch, or the cutover fires the exact loss mode the branch fixes. This automates that step.

## Measured on prod, 2026-08-13, four open rollouts

| rollout | object | in container | at risk |
|---|---|---|---|
| `2026-08-05T14-33-11` | 9 115 293 | 9 796 319 | 680 KB |
| `2026-08-06T14-40-06` | 926 136 | 926 136 | 0 |
| `2026-08-07T15-30-46` | 7 786 439 | 8 051 317 | 259 KB |
| `2026-08-13T14-09-11` | **absent** | 1 018 056 | 995 KB |

~1.9 MB, and the newest session has no object at all — it has never been committed once. The flush is periodic, so exposure is a moving window rather than the whole session, but a session that never flushes is unbounded.

## What it does

`scripts/migrate-open-rollouts.sh` — no arguments, snapshot mode:

1. Finds rollouts held open **for writing**, from `/proc/*/fdinfo` — `mtime` on the mount tracks the last object commit, not the last append, so a `find -newermt` sweep misses exactly the files at risk. Read-only descriptors are skipped.
2. Snapshots each one, trims a half-written trailing line (append-only JSONL, so a snapshot is a valid prefix), and reports what that cost.
3. Uploads to `state/codex/migration/<run>/` through the bucket API, with a manifest.
4. Verifies by **downloading the object back and comparing sha256**, then exits non-zero if anything failed — so a deploy gated on this stops rather than proceeding.

`--restore` is the boot half: reconcile the newest run, restoring only where the migrated copy is strictly longer than what is on the canonical path, so a clean shutdown is left alone. `--dry-run` works for both.

Copies go to a separate prefix and never to the canonical path: a dying writer flushing its short view over a full copy is the clobber this exists to prevent.

## One finding worth keeping

`get_bucket_file_metadata().size` is wrong for these xet-backed objects — a 9 796 319 B rollout reported 767, a 2 329 B manifest reported 694. `list_bucket_tree` sizes and `download_bucket_files` content are both correct. The first draft of this script verified with the metadata call and reported four false failures on a set of uploads that were in fact byte-perfect. There is a comment in the source saying so; please don't simplify it back.

## Not wired in yet

Deliberately no `entrypoint.sh` change. #25 rewrites that region, and I did not want to hand its author a conflict. The boot hook is one line once that settles:

```sh
scripts/migrate-open-rollouts.sh --restore || echo "rollout restore failed" >&2
```

Whether restore should be fail-closed at boot (per #25's "restore fails closed") rather than warn-and-continue is a call for whoever owns that path.

## Verification

Both modes were run against the live prod Space. Snapshot produced four byte-exact verified objects at `state/codex/migration/20260813T143542Z/`; `--restore --dry-run` correctly declined to touch all four, since the in-container view is longer than the migrated copy. Nothing on a canonical path was written.

### DIFF
--- a/scripts/migrate-open-rollouts.sh
+++ b/scripts/migrate-open-rollouts.sh
@@ -1,0 +1,330 @@
+#!/usr/bin/env bash
+# Get every open Codex rollout onto the bucket as a CLOSED object, before a
+# deploy replaces the container.
+#
+#   scripts/migrate-open-rollouts.sh                # snapshot, upload, verify
+#   scripts/migrate-open-rollouts.sh --dry-run      # report, touch nothing
+#   scripts/migrate-open-rollouts.sh --restore      # boot side: reconcile
+#
+# WHY THIS EXISTS
+#
+# Codex appends to $CODEX_HOME/sessions/YYYY/MM/DD/rollout-<ts>-<uuid>.jsonl and
+# holds that descriptor open for the whole life of a resumed session. That path
+# is a symlink onto the /data bucket (entrypoint.sh), which is FUSE over object
+# storage, and the mount's streaming writer can hold the entire open epoch in
+# memory until close. Reads inside this container see the buffer, so the file
+# looks complete from here. The OBJECT does not have it.
+#
+# Measured on prod, 2026-08-13, three live Codex sessions:
+#
+#     rollout-2026-08-05T14-33-11-…   local 9 671 486 B   object     767 B
+#     rollout-2026-08-07T15-30-46-…   local 8 051 317 B   object     767 B
+#     rollout-2026-08-13T14-09-11-…   local   971 492 B   object  absent
+#
+# 767 bytes is the session header written at open. Everything after it exists
+# only in this container. Kill the container and ~18 MB of conversation across
+# three agents is gone — which is exactly the loss mode PR #25 fixes, and
+# exactly what deploying that fix would trigger one last time on the way in.
+#
+# So: copy the writer's own view out through the bucket API (HTTP, not the
+# mount), verify it landed, and only then let the deploy proceed.
+#
+# WHAT IT DOES NOT DO
+#
+# It does not stop, signal, or touch the Codex processes. It reads. A rollout is
+# append-only JSONL, so a snapshot taken while the writer runs is a valid prefix
+# of the session — we trim a partial trailing line and record how many bytes
+# that cost. Losing the last half-written line is not the failure mode anyone is
+# worried about here.
+#
+# It does not write the canonical path. If it did, the dying writer could flush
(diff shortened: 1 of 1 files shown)


lvwerra:
## Runbook for the prod cutover

Merged as 16717ea. The description covers what the script does; this is the operating order, which matters more than the flags.

### 1. Snapshot — inside the prod container, as the last step before the restart

```sh
scripts/migrate-open-rollouts.sh
```

Two constraints, both easy to get wrong:

- **It must run in the container being restarted.** It finds rollouts by walking `/proc/*/fdinfo` for descriptors held open *for writing*. That is not observable from outside, and it is deliberate: `mtime` on the bucket mount tracks the last object commit rather than the last append, so a `find -newermt` sweep misses exactly the files at risk.
- **It must be the last thing before the deploy.** Exposure grows continuously. Measured on prod 2026-08-13 it was ~1.9 MB across four rollouts; re-measured 2026-08-14 the same four were at 3.2 MB. A snapshot taken an hour early leaves an hour of appends behind.

Gate the deploy on the exit code. Non-zero means at least one object failed its sha256 read-back — stop, do not restart. Note the run prefix it prints (`state/codex/migration/<timestamp>/`).

### 2. Deploy

Push to the Space repo as usual. There is no prod deploy script — `deploy-dev-space.sh` only targets dev instances.

### 3. Restore — after boot, manually

```sh
scripts/migrate-open-rollouts.sh --restore
```

**This does not happen on its own.** The boot hook is deliberately not in `entrypoint.sh` (see the description — #25 rewrites that region). Until someone wires it, the snapshot sits on the bucket doing nothing unless restore is run by hand after the container comes back. The data is not lost in that case, but it is not back in place either.

Restore only writes where the migrated copy is strictly longer than what is on the canonical path, so running it after a clean shutdown is a no-op. Rehearse with `--restore --dry-run` first; it is safe and prints what it would do.

### If you are debugging this

`--dry-run` works in both modes and touches nothing.

Do not verify uploads with `get_bucket_file_metadata().size` — it is wrong for these xet-backed objects (a 9.8 MB rollout reports 767 B). The first draft of this script did exactly that and reported four false failures on byte-perfect uploads. `list_bucket_tree` sizes and `download_bucket_files` content are both correct, and the script uses the latter.

PR #74, opened 2026-08-15
Read the trace again when the operator comes back to the tab

lvwerra:
Switch to another app with the reader open, come back, and it is still showing the turns from when you left. This makes the return itself do a read.

## What is actually wrong

Not a dead socket — the reader has no socket. It follows a trace by polling `loadNewer` on an interval, and two things about that interval break a return from the background:

1. **On a quiet session nothing is scheduled at all.** `web/src/lib/traceWindows.ts` gates the interval on `if (!seekable || paused || hidden || (live === false && !fresh)) return undefined`. `ConversationView` passes `live = session.state === 'working' && !paused`, and `fresh` means the trace moved within `FRESH_MS` (2 min). A reader left open on a session that finished while the operator was away is `live === false` **and** stale, so no interval is created on return, and nothing reads the trace again for as long as they keep looking at it.
2. **When one IS scheduled, its first tick is a whole interval late** — `setInterval` does not fire immediately. That is 3s on a live trace, 10s on a quiet one, staring at content you know is out of date.

There is a third failure that only bites on a phone, which is where this was reported. iOS Safari suspends a backgrounded page, so a window request that was in flight when the operator switched apps **may never settle**. `loading` is the reader's one-request-at-a-time latch and it is released only in that request's `finally` — which never runs. Every later read then returns early against a latch nothing will ever release, so the reader stays stale even once polling resumes.

## The fix

One effect. On `visibilitychange` back to visible — and on `pageshow` with `persisted`, since a back/forward-cache restore fires no visibilitychange and Safari leans on that path — abandon whatever request was outstanding, release the latch, and read immediately:

- Abandoning bumps `gen`, which is what makes the old request's response *and* its `finally` no longer ours. Bumping rather than only aborting matters: the aborted request's `finally` would otherwise clear `loading` out from under the fresh read that replaced it.
- `cursor.current ? loadNewer() : loadTail()` — if the initial tail never landed (frozen with the tab), the return retries the tail instead of paging from a cursor that does not exist.
- A paused (off-screen) pane is skipped: it reloads on unpause already.

Unchanged: a hidden reader still asks for nothing, and the interval's own gating is untouched. This only adds one read at the moment of return. For the non-seekable SQLite harnesses — deliberately never polled, because a window costs a whole-file re-parse — that is one read per return to the foreground, not a poll loop.

## Verified

In a real browser (`web/test/traceWindows.test.mjs`, the existing React/Chromium harness), and **every new check fails without the fix** — confirmed by stashing the source change and re-running:

- turns written while the tab was hidden are on screen **within 2s of the return**, which is under both poll intervals, so it cannot be a tick that did it;
- the returning reader asks `{at: 'after'}` for what it missed rather than re-reading the whole tail;
- a request left frozen across the switch is **abandoned** (`aborted`), and the turns still land — the latch cannot strand the reader;
- a `pageshow` restore refreshes the same way;
- the existing invariant still holds: a hidden reader issues no requests at all (asserted over 1.2s of real time).

The stale case is the one the operator hit, so the fake source returns a trace whose `lastTs` is ten minutes old — `live: false` plus a stale trace is precisely the combination that schedules no interval.

Also: `web` typecheck clean, `web npm test` 11/11 + trace-windows, production build clean, and `server/screenshot-input.test.mjs` — the suite that opens the rendered reader in a browser — passes 20/20 against a build with this change.

### One honest limit on the verification

Chromium cannot be genuinely backgrounded in this container, so the browser's own occlusion is not what drives the test. I tried, and none of these produce a hidden page here: `bringToFront()` with a sibling page (headless and headed), CDP `Browser.setWindowBounds` minimised, CDP `Page.setWebLifecycleState`, a same-window tab via `Target.createTarget`, headed Chromium on Xvfb (Linux Chrome has no occlusion tracking, and there is no window manager to minimise a window), and WebKit, which will not launch without host libs this container lacks. Real bfcache is unavailable too: Chromium under automation refuses to cache the page, `persisted` staying false even over http with the feature forced on.

So the tests override `document.visibilityState`/`hidden` and dispatch the real events through the DOM. The hook reads `document.hidden` and nothing else, so this is the same code path a true app switch takes; what it does not prove is that the browser fires the event, which is the browser's guarantee rather than this app's. Someone with a phone can settle that in ten seconds, and it is worth doing before this is trusted on iOS.

## Note for whoever reviews next

`server/screenshot-input.test.mjs` failed twice for me on its hardcoded port 7896, with two different symptoms (a duplicate `.ovt-tile` strict-mode violation, then a wrong rejection reason). It fails the same way on unmodified `main`, and passes on a unique port with this change applied — so it is cross-talk between agents running suites concurrently, not a regression. The fixed test ports (7893, 7896, 7897) are worth making collision-proof in their own PR.

### DIFF
--- a/web/src/lib/traceWindows.ts
+++ b/web/src/lib/traceWindows.ts
@@ -269,6 +269,10 @@ export function useTraceWindows(src: TraceSource, srcKey: string, opts: {
   // a window can know. If it fails or is slow, the header simply says how much
   // is loaded.
   const summaryReady = !!meta;
+  // Bumped when a return to the foreground abandons a summary that was in
+  // flight. `summaryStarted` is a ref, so clearing it alone would not re-run the
+  // effect below — nothing else in its deps changes when a tab comes back.
+  const [summaryEpoch, setSummaryEpoch] = useState(0);
   useEffect(() => {
     // `meta` is set by the tail response and this effect runs after that render
     // commits. Starting the clock on mount let a slow tail lose a race to every
@@ -287,7 +291,7 @@ export function useTraceWindows(src: TraceSource, srcKey: string, opts: {
         .finally(() => { if (summaryAbort.current === abort) summaryAbort.current = null; });
     }, SUMMARY_DELAY_MS);
     return () => { dead = true; window.clearTimeout(h); };
-  }, [src, srcKey, paused, summaryReady, summary]);
+  }, [src, srcKey, paused, summaryReady, summary, summaryEpoch]);
 
   // The transcript may still be being written — see LIVE_MS above. Except when
   // "a window" costs a whole-file read: the SQLite harnesses have no byte
@@ -339,6 +343,18 @@ export function useTraceWindows(src: TraceSource, srcKey: string, opts: {
         outstanding.abort();
       }
       loading.current = false;
+      // The whole-file summary can be the frozen request instead of a window.
+      // It is single-flight through `summaryStarted`, which nothing else clears,
+      // so the header would keep its "loaded so far" counts — no turn total, no
+      // user turns, no session cost — for the life of this source, while the
+      // turns below it updated normally on every return.
+      const outstandingSummary = summaryAbort.current;
+      if (outstandingSummary) {
+        summaryAbort.current = null;
+        summaryStarted.current = false;
+        outstandingSummary.abort();
+        setSummaryEpoch((epoch) => epoch + 1);
+      }
       void (cursor.current ? loadNewer() : loadTail());
     };
     const onVisible = () => { if (!document.hidden) resync(); };
--- a/web/src/lib/traceWindows.ts
+++ b/web/src/lib/traceWindows.ts
@@ -312,6 +312,49 @@ export function useTraceWindows(src: TraceSource, srcKey: string, opts: {
     return () => window.clearInterval(h);
   }, [loadNewer, lastTs, seekable, paused, hidden, live]);
 
+  // Coming back from another app has to show what happened while away, and none
+  // of the loop above survives the trip. The interval is cleared while hidden,
+  // and the effect that would restore it schedules NOTHING when the host says
+  // the session stopped and the trace has been quiet — so a reader left on a
+  // finished session shows the turns from before the operator switched away,
+  // for as long as they keep looking at it. Even when a poll is scheduled, an
+  // interval's first tick is one whole interval late.
+  //
+  // A frozen request makes it worse. iOS Safari suspends a backgrounded tab, so
+  // a window request that was in flight when the operator left may never settle
+  // — and `loading` is released only in that request's `finally`, so every later
+  // read returns early against a latch nothing will ever release. That is a
+  // reader which stays stale even once polling resumes.
+  //
+  // So a return reads immediately: abandon whatever was outstanding (bumping the
+  // generation, which is what makes its response and its `finally` no longer
+  // ours), release the latch, and ask for the turns written while away.
+  useEffect(() => {
+    const resync = () => {
+      if (document.hidden || cb.current.paused) return;
+      const outstanding = windowAbort.current;
+      if (outstanding) {
+        gen.current += 1;
+        windowAbort.current = null;
+        outstanding.abort();
+      }
+      loading.current = false;
+      void (cursor.current ? loadNewer() : loadTail());
+    };
+    const onVisible = () => { if (!document.hidden) resync(); };
+    // Restoring a page from the back/forward cache fires no visibilitychange:
+    // the tab was never hidden, the whole page was frozen and thawed. Safari
+    // leans on this hard — a swipe back, or an app switch that outlived the
+    // renderer, comes back this way.
+    const onShow = (e: PageTransitionEvent) => { if (e.persisted) resync(); };
+    document.addEventListener('visibilitychange', onVisible);
+    window.addEventListener('pageshow', onShow);
+    return () => {
+      document.removeEventListener('visibilitychange', onVisible);
+      window.removeEventListener('pageshow', onShow);
+    };
+  }, [loadNewer, loadTail]);
+
   const atStart = !!cursor.current?.atStart;
   const blocked = !!cursor.current?.blocked;
   // MEMOIZED, and it has to be: a host that lifts this into its own state — the
--- a/web/test/traceWindows.test.mjs
+++ b/web/test/traceWindows.test.mjs
@@ -26,9 +26,14 @@ await build({
 
       const calls = [];
       const tails = [];
-      const page = (turns, end = 20) => ({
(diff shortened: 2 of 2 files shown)


lvwerra (commented):
Reviewed on both axes: does it do what it claims, and does it fix what the operator reported. It does both. Green here too: `web` typecheck clean, `npm test` 11/11 + trace-windows, on Chromium 1234 from the image.

## Correctness

**The "the transport is dead, not idle" objection does not apply to this surface, and I checked rather than took the PR's word for it.** `new WebSocket` appears exactly once in `web/src` — `TerminalPane.tsx:656`, the terminal. The reader fetches windows over HTTP through `api.window`, and the session state it derives `live` from (`ConversationView.tsx:106`) comes from the tree poll in `App.tsx:414-416`, which already skips while hidden and refreshes on `visibilitychange`. So there is no connection to re-establish here, and a re-read is the whole fix. The connection-level failure that *does* apply on iOS — a fetch frozen in flight and never settled — is the one the PR handles, and it is the part I spent the most time on.

**The tests pin behaviour, not handler registration.** I verified this by breaking the source two ways and re-running:

- `git checkout origin/main -- web/src/lib/traceWindows.ts` with the new tests → fails at `web/test/traceWindows.test.mjs:214`, the "what was written while away is on screen" wait. So the assertion is the turn count on screen within 2s, under both poll intervals, not a listener count.
- Keeping the effect but deleting only the abandon/latch block (`traceWindows.ts:335-341` — the `gen` bump, `windowAbort` clear, `loading.current = false`) → the first case still passes, and it fails at `web/test/traceWindows.test.mjs:237`, the frozen-request case. So the subtlest claim in the PR — that releasing the latch matters independently of doing a read — is the one thing a test actually distinguishes. That is the check I would have asked for if it were missing.

The ordering in `resync` is right for the reason the comment gives: `windowAbort.current = null` *before* `outstanding.abort()`, so the aborted request's `finally` (`traceWindows.ts:235-236`) fails both its guards and cannot null the replacement's controller or clear `loading` under it.

**One gap, minor, same failure mode:** the summary request is not resynced. `traceWindows.ts:280-281` sets `summaryStarted.current = true` before the fetch and the effect returns early forever once it is set (`:277`). If the request frozen by the app switch is the whole-file summary rather than a window, `summary` stays null for the life of that source: turns keep updating on return, but the header never gets `total`, `userTurns`, or the session cost. `resync` abandons `windowAbort` and leaves `summaryAbort` alone. Not a blocker — windows are what the operator noticed — and not a one-liner either, since clearing `summaryStarted` will not re-run an effect whose deps are `[src, srcKey, paused, summaryReady, summary]`. Worth a follow-up rather than a change here.

## Does it fix item 6

Yes, and the follow-through is better than "one read on return":

- The stale case is the reported one and is covered: `live === false` plus a trace older than `FRESH_MS` schedules no interval at all (`:310`), so without this nothing would ever read again — that is why the fake source returning a ten-minute-old `lastTs` is the right fixture.
- For a session that is still running, the catch-up read updates `lastTs`, which makes `fresh` true and gets the 3s interval scheduled again by `:305-313` — so even if `session.state` were momentarily stale on return, the reader keeps updating rather than needing another app switch to unstick it.

One thing to be aware of rather than fix: `resync` does not check `seekable`, while the interval effect does (`:310`). For the SQLite harnesses that means a whole-file re-parse on every return to the foreground, which on a phone is a frequent event. The PR body calls this out as deliberate and I agree with the trade — the server side is `windowIndex` (`server/src/traces.js:1779`), which never sets `gap`, so it splices cleanly and will not reset the reader's scroll position. If it ever shows up as Overview memo eviction, gating resync on `seekable` plus a minimum interval is the knob.

The "one honest limit" section is the right disclosure and I would not block on it: what is unproven is that the browser fires the event, which is the browser's contract, and the `pageshow`/`persisted` handler covers the restore path where it deliberately does not. A ten-second check on a real phone is still worth doing before this is trusted on iOS.

lvwerra:
Thanks — both points answered. The summary gap is **fixed here** (`ca62b3a`) rather than deferred; the `seekable` trade I still hold, with reasoning below.

## The summary is now re-armed on return

You were right that it is not a one-liner, and the reason is worth recording: `summaryStarted` is a ref, and the effect that would start another whole-file read has deps `[src, srcKey, paused, summaryReady, summary]` — none of which change when a tab comes back. Clearing the ref alone leaves the effect un-run and the summary dead for the life of the source.

So a `summaryEpoch` counter joins those deps, and `resync` bumps it — but only when a summary was actually outstanding at the moment of return. A return with no whole-file read in flight costs nothing, and the 400ms post-paint delay is unchanged, so this cannot race a tail the way mount-based timing used to.

```ts
const outstandingSummary = summaryAbort.current;
if (outstandingSummary) {
  summaryAbort.current = null;
  summaryStarted.current = false;
  outstandingSummary.abort();
  setSummaryEpoch((epoch) => epoch + 1);
}
```

The `gen` bump for windows does not touch this: the summary's own guard is the effect-cleanup `dead` flag, not the generation, so abandoning a window cannot discard a healthy summary response.

## Broken two ways, since that is what convinced me last time

The harness's `summary()` now resolves on demand instead of never, so the test can pin recovery rather than just a second request:

- **Drop the summary block from `resync` entirely** → times out at `test:273`, the wait for that read to be abandoned. Nothing abandons it.
- **Keep the abort and the ref clear, drop only the epoch bump** → times out at `test:265`, the wait for *any* live whole-file read. By then every summary has been aborted by an earlier return and no further one was started — the ref-only fix failing exactly as you predicted, and a sharper demonstration of it than I could have written by hand.

With the fix, the retried read is answered and the header fills in (`__traceHead.total === 41`), so what is pinned is the header recovering, not a request count. `web` typecheck clean, `npm test` 11/11 + trace-windows, production build clean.

## On `seekable` — yes, I still hold the trade

Confirming on the record now that it is written down. `resync` deliberately does not gate on `seekable` while the interval effect does:

- The complaint being fixed is "I came back and the reader had not moved". That is exactly as true for a SQLite harness as for a JSONL one, and gating resync on `seekable` would fix it for some readers and leave the others in the reported state.
- The cost is bounded per *event*, not per unit time: one whole-file read each time the tab comes back, versus the poll loop's read every `IDLE_MS` — which is precisely why those sources are excluded from the loop and not from a one-shot.
- As you note, `windowIndex` never sets `gap`, so the result splices instead of resetting, and the reader's scroll position survives.

The knob if it ever bites — a phone alt-tabbing repeatedly on a large SQLite trace showing up as Overview memo eviction — is a minimum interval between resyncs for non-seekable sources only, rather than dropping the resync. I would rather see that measured than pre-empted, since the failure mode it guards against is invisible to the operator while the one it fixes is what they filed.

Still open and still worth ten seconds on a real phone: that iOS actually fires `visibilitychange` on app switch for this page, which is the one thing no test here can prove.

PR #75, opened 2026-08-15
Mobile: the way back rides in the pane header

lvwerra:
Two items from the operator's design list, both mobile chrome:

> on mobile the session window has a go back arrow and title above the session window, costing precious vertical space, i think we can add the go back arrow simply left to the agent logo in the header

> in mobile view the terminal/reader selector can be aligned left

## The bar, and what was actually on it

Staged on a phone, a session had a bar above it holding a back arrow and the session's name — and the pane header immediately below it already showed that name. So the bar's only unique job was the arrow, and it charged a full row for it.

The arrow moves into the pane header, immediately left of the agent's logo, exactly as suggested. **At 390px the terminal grows 544px → 582px**, and the header it moved into is not one pixel taller: the button stretches into the header's existing vertical padding (`align-self: stretch` plus a negative margin) rather than adding any. A header that grew by a row would have handed straight back what dropping the bar won.

## What I did not do, and why — the part worth reviewing

The bar could not simply be deleted. It is the only way back on three other surfaces, and one of them would have been left with no way out at all:

| surface | why the bar stays |
|---|---|
| a **group** | its bar is also the chip pager between the group's agents |
| the **Overview** | it has no pane header at all — nothing to put an arrow in |
| **files** / **trace** panes | flat headers with no identity block; and the files pane's leftmost button is *already* a back button, to the file list. A second arrow there is two backs meaning two different things. |

So the rule is *the bar survives wherever nothing else can hold the arrow*. That is `paneOwnsBack` in `web/src/lib/mobileBack.ts` — extracted and unit-tested rather than left inline, because it decides whether a surface has any way back at all, and the failure mode is a phone with no exit. A group keeps its single arrow in the bar and does **not** grow a second one in the pane header; that case is asserted.

## The mode switch

`terminal | reader` sat adrift in the middle of the bottom bar on a phone. Nothing was centring it — the `.spacer` that pushes the zoom keys right on a wide window was pushing it too. Ordering it ahead of the spacer at the phone breakpoint gives it the left edge and leaves zoom on the right; its left offset goes from 147px to 9px at 390px.

## Verification

**[Before/after screenshots at every viewport &rarr;](https://claude.ai/code/artifact/114d4026-ace0-457a-80e8-77f7af06ff1f)**

Playwright at **320 / 390 / 768 / 1280**, before and after, on a real instance with a loose session, a group of two, a files pane and the Overview.

| | mbar height | terminal | mode switch left | back in header |
|---|---|---|---|---|
| single 320 | 38 → **gone** | 544 → **582** | 77 → **9** | no → **yes** |
| single 390 | 38 → **gone** | 544 → **582** | 147 → **9** | no → **yes** |
| group 390 | 45 → 45 | 537 → 537 | 147 → **9** | no → no |
| 768 / 1280 | — | 615 → 615 | unchanged | no → no |

**Desktop is untouched.** Every measured number is identical, and a chrome-only pixel diff (terminal contents and the animated status dot masked, since both differ run to run) comes to 0–17 pixels — against a **7–39 pixel noise floor established by running the same build twice**. The residual is antialiasing on a rounded corner, not layout.

**#71 holds**: no page scrolls sideways at any width, checked in the same pass. The one new hardcodable number — the header padding the back button cancels — is a `--ph-pad-y` variable rather than a restated `7px`, which is the rule #71 established after a 14px bleed drifted against an 8px gutter.

- 13 browser checks: the arrow appears only where intended, clicking it returns to the list and persists the choice, exactly one way back is on screen per surface, desktop gets none.
- 13 unit checks on `paneOwnsBack` (`node test/mobileBack.test.mjs`, wired into `npm test` in `web/`).
- `web`: 11/11 plus the new file. `server`: all 13 suites green.
- `npm run test:ui` has one flaky timeout under load — it fails on **unmodified main** the same way, and moves between runs (`remote uploads…` on one run, `a shared-terminal watcher…` on the next). Not from this change; called out rather than hidden.

Not verified on a real phone — only at phone viewports in Chromium.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
-    "test": "node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs",
+    "test": "node test/mobileBack.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -19,6 +19,7 @@ import type { Cli, GridSpec, MoveTarget, OverviewChip, OverviewSort, Session, Tr
 import { onPaneMode, readPaneMode, writePaneMode } from './lib/paneMode';
 import { hiddenSessionIds } from './lib/overviewHidden';
 import { useReaderBatch } from './lib/readerBatch';
+import { paneOwnsBack } from './lib/mobileBack';
 import { isPassive, isRemote } from './types';
 import { EyeGlyph, EyeOffGlyph, GridGlyph, ListGlyph, SortGlyph } from './components/icons';
 import { uploadPendingAttachments } from './lib/attachments';
@@ -516,6 +517,13 @@ export default function App() {
     [activeGroup, sessById],
   );
 
+  // A staged agent carries its own way back, in its header left of the logo, and
+  // the bar above it goes away with the row it cost. See mobileBack.ts for which
+  // surfaces still need the bar and why.
+  const ownsBack = paneOwnsBack({
+    isMobile, staged: mobileStage, inGroup: !!activeGroup, cli: activeSingle?.cli,
+  });
+
   // Tile grid for the active group: the chosen layout, or auto (fit the count).
   // On mobile it's always a single pane — the chip strip switches between them.
   const grid: GridSpec = activeGroup && !isMobile ? (activeGroup.layout ?? autoGrid(groupSessions.length)) : { cols: 1, rows: 1 };
@@ -903,6 +911,7 @@ export default function App() {
                 active={shown && deckVisible && s.id === focusedId}
                 dragId={shown && canDrag ? `p:${s.id}` : undefined}
                 isMobile={isMobile}
+                onBack={ownsBack ? () => setMobileStage(false) : undefined}
                 onDragActive={setPaneDrag}
                 onFocus={() => setFocusedId(s.id)}
                 onRename={(name) => renameSession(s.id, name)}
@@ -935,6 +944,7 @@ export default function App() {
                 groupName={groupNameOf[s.id]}
                 focused={visibleSessions.length > 1 && s.id === focusedId}
                 dragId={canDrag ? `p:${s.id}` : undefined}
+                onBack={ownsBack ? () => setMobileStage(false) : undefined}
                 onDragActive={setPaneDrag}
                 onFocus={() => setFocusedId(s.id)}
                 onRename={(name) => renameSession(s.id, name)}
@@ -1063,7 +1073,7 @@ export default function App() {
       />
 
       <div className="main">
-        {isMobile && mobileStage && (
+        {isMobile && mobileStage && !ownsBack && (
           <div className="mbar">
             <button className="icon-btn mback" onClick={() => setMobileStage(false)} title="Back to list">‹</button>
             {activeGroup ? (
--- a/web/src/components/RemotePane.tsx
+++ b/web/src/components/RemotePane.tsx
@@ -5,7 +5,7 @@ import * as api from '../api';
 import Logo from './Logo';
 import { renderMarkdown } from '../lib/markdown';
 import { groupLabel, sessionTitle } from '../lib/sessionTitle';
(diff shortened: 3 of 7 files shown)


lvwerra (commented):
Reviewed as the author of #71, on the two axes asked for. **No change requests — this is good.** Everything below is either a confirmation I ran myself or a note for the merge.

I rebuilt this branch and main against the same instance (a loose agent, a group of two, loose `files`/`shell`/`trace`/`remote` panes, the Overview) and measured at 320 and 390.

## 1. Correctness

**It does what it claims.** My numbers, main → this branch, at both widths:

| | main | #75 |
|---|---|---|
| `.mbar` | 38px | gone |
| `.pane-head` height | 37px | **37px** (unchanged) |
| terminal / reader body | 645px | **683px** (+38, exactly the bar) |
| mode switch left @390 | 155 | 17 |
| mode switch left @320 | 85 | 17 |

The header really does not grow — `align-self: stretch` plus the negative margin lands the button at **31×36**, against the old `.mback`'s **32×30**, so the hit target got *taller*, not smaller. `.zoombar .modebar { order: -1 }` is inside `@media (max-width: 720px)`, so 768 and desktop are untouched; I confirmed the tablet is unchanged.

**#71 is intact.** In reader mode on this branch, at 320 and 390, with a code block, a wide table, a long URL and an unbroken identifier on screen:

```
--cx-gutter: 8px   body padding-right: 8px   band bleed: 0px (flush with the scroller)
.cxv-body scroll room: 0   page scroll room: 0   conversation-level scrollers: none
wide content still scrolling inside itself: pre 796px, table 212px
```

Identical on main and here — the gutter still pays out both the padding and the band's bleed, and nothing in the conversation scrolls sideways.

**The invariant that matters holds on every surface.** The failure mode this rule guards is a phone with no way out, so I drove all seven staged surfaces rather than trusting the truth table. Ways back, this branch: loose agent 1 (header), group 1 (bar, pager intact with its chips), files 1 (bar), trace 1 (bar), remote 1 (header), shell 1 (header), Overview 1 (bar). No surface with zero, none with two visible.

**A long name at 320 is fine**: header stays 37px, back/title/close do not overlap, the title ellipsises. The title does sit ~7px right of centre now, since it is the middle column of a grid whose left column grew — cosmetic, and I would not change it.

**`--ph-pad-y` has no drift risk**: `.pane-head` has exactly one padding declaration and it is the variable, so there is nothing left to restate. (Nice reuse of the #71 lesson, and correctly applied.)

**What the tests actually pin.** 12 cases over the pure function, and — the part I like — `isPassive` is imported rather than restated, so adding a passive CLI moves the rule and the test together. What they do not reach is the wiring: that the bar is hidden *exactly* when the header shows the arrow. Both halves derive from the one `ownsBack` const, so they cannot disagree by accident today, and a DOM-level test would be new infrastructure for this suite — I would not add it. Consider this comment the integration check instead: seven surfaces, both widths, all single-exit.

**One note, not a defect.** Retained (hidden) panes each render their own `.ph-back`, so after visiting a few agents the DOM holds 2–4 "Back to list" buttons — all at 0×0, not focusable, not in the way. Main already does exactly this with `.ph-close`, so it is the existing pattern rather than a regression. The only practical cost: automation must select the *visible* one (`.filter({ visible: true })`) — it cost me a run, so it will cost the next person one too.

## 2. Does it solve the operator's problem?

Yes, and literally as asked — the arrow is left of the agent's logo, and the row it used to cost is now terminal.

> on mobile the session window has a go back arrow and title above the session window, costing precious vertical space

The bar carried an arrow and a name the header immediately below already showed, so the space bought nothing. **+38px of terminal at 390px, +38px at 320px**, and the same +38 in reader mode — where it is worth most, since the reader is the surface they read on a phone.

> in mobile view the terminal/reader selector can be aligned left

Left edge at both widths (155 → 17 at 390, 85 → 17 at 320), zoom keys still right.

The reasoning about *not* deleting the bar is the right call and the table in the description is accurate — I checked the two that would have hurt: the group keeps its pager, and the files pane keeps its single back-to-the-file-list meaning rather than growing a second arrow next to it.

## Merge collisions

I tested these by actually merging, not by reading paths:

- **#75 + #77 (mine): clean.** Only `styles.css` overlaps, and the hunks are ~25 lines apart inside the same mobile block (yours `.zoombar .modebar`, mine `--ov-first-line` in the nested `pointer: coarse` block). Auto-merges.
- **#75 + #76: clean.** #76 touches `.pane-head .ph-image` (in `.ph-right`) while you add `.ph-back` to `.ph-left` — no textual conflict, but you are both reshaping the same header row, so whoever lands second should re-check that row at **320px**, where it is tightest.
- **#77 + #76: clean.**

Worth knowing for ordering: #76 reworks the attachments strip, and my #77 aligns `.image-pick` to the composer's first line — no conflict today, but if #76 changes that strip's markup after landing, the alignment is the thing to re-measure.

PR #77, opened 2026-08-15
Reader: align the newest prompt, trim under the composer, controls on the first line

lvwerra:
Three items from the operator's design list (`improv.md`), all in the rendered conversation's prompt/composer area. Each was reproduced and measured at a real viewport first — desktop 1280×900 and iPhone 13 — not read off the stylesheet.

## 1. The newest prompt sat 16px left of every prompt above it

The optimistic echo — the prompt you just sent, before the transcript catches up — was written out by hand at **both** call sites as a loose `.cx-prompt`:

```jsx
{sent && (<><div className="cx-prompt">{sent.text}</div><div className="cx-running mono">working</div></>)}
```

`.cx-prompt`'s `margin-left: -1.23em` exists to cancel `.cx`'s matching `padding-left`. With no `.cx` around it there was nothing to cancel, so the band hung a full gutter to the left — and on a phone its `❯` was half off the pane edge.

It is a `PendingExchange` now: a `.cx` section, the same shape as the turn it becomes a second later, exported from `Exchange.tsx` and used by the reader **and** the card, so they cannot drift again.

Measured, left edge relative to the reading column:

| | older prompts | newest prompt |
| --- | --- | --- |
| before | `0` | **`-16px`** (parent `.cxv-col`) |
| after | `0` | `0` (parent `.cx`) |

The dead `.cxv-body > .cxv-col > .cx { padding-left: 16px }` rule from #49 goes with it. It matched nothing — every exchange is wrapped in a row div — and the moment something *did* match, it would have pinned that turn to a 16px gutter while every turn above it kept the em and moved under zoom.

## 2. Under the composer was reference material, not an action

The reader ended in a 25px strip of grey: the workspace's **absolute path** and the **date** the conversation started.

- **Removed the path.** Printed under a reply box it is something you read once and then scroll past forever. (An earlier version of this description said "it is in the pane header a few centimetres above" — that holds on a desktop and **not** on a phone, where `styles.css:692` hides `.ph-path` under `@container (max-width: 520px)` and a 390px viewport leaves the pane ~374px. Corrected after review: on a phone the path is now nowhere in the reader, and that is the right amount of nowhere for a path under a reply box.)
- **Kept the date, moved it — twice.** This was the one fact nothing else showed: turn times are clock-only (`fmtClock` → `07:00 PM`), so a conversation read on Friday never said which day it started. It first went onto the turn-count's `title`, which was wrong — touch has no hover, and iOS Safari long-presses into the callout menu, so on the device these items were filed from the fact was shown nowhere at all. It is now **visible text** in `.cxv-bar` (`.cxv-when`, "Aug 14", with the year when it is not this one), beside the model, turns and tokens. Measured before choosing it: the bar `flex-wrap`s at any width, so it costs no height — bar 56px and reader body 548px at 390px, identical with the date and without, and the same at 320px. The full timestamp stays on its own `title` as a bonus for a mouse, never as the only channel.
- **Kept `continue in a new agent ↗`.** It is an action, and it has no other home.
- With no handover to offer, `.cxv-foot` no longer renders at all, instead of leaving an empty bordered strip.

## 3. The arrow and the paperclip drifted down as the draft grew

A composer grows downward as you type, and centred controls follow it. At six lines on a phone the `❯` and the paperclip sat **56px** below the line they belong to, level with the middle of a sentence.

Alignment is `start` now, and each control is offset by half the difference between its own height and the first line's — one expression, three controls. `--ov-first-line` is restated wherever the textarea's size is actually decided (13px in the card, `--cx-base` in the reader, 16px on a touch phone where iOS demands it to prevent zoom-on-focus), because that is the thing it measures.

Controls vs the first line's middle:

| | desktop (2 lines) | phone (6 lines) |
| --- | --- | --- |
| before | arrow/clip/send **+10px** | arrow/clip/send **+56px** |
| after | `0` / `+1px` / `+1px` | `0` / `0` / `0` |

The send key follows the same rule. Leaving it centred while the other two moved would have read as an accident rather than a decision.

## The Overview card

#37 made this renderer shared, so the card was **checked, not assumed** — and it does change, in one intended way:

- its composer's controls land on the first line too: `0.0px` off, empty and one-line drafts, both viewports (before: also ~0, since with one line centring and top-alignment coincide — the visible difference is only in a wrapping draft, screenshot below)
- the reply row is **2.6px shorter on desktop / 0.4px on a phone**, because the negative offsets no longer let the 26px send key set the row's height. If a reviewer wants byte-identical card metrics, a `min-height` on `.ov-live` restores it — I left it out rather than pin a number to an old accident.
- nothing else in the card moves; the compact tile view has no composer and is untouched
- the card's own pending echo goes through the digest's `.ov-prompt` when it has no transcript loaded — that path is unchanged

## Verification

**[Before/after captures, all three items →](https://claude.ai/code/artifact/ca6ab387-8fde-4034-96e0-8f6e5799674b)** — the screenshots below were `/tmp` paths on a dev box whose local disk does not survive a Space restart, so nobody but me could see them. They are now a page anyone can open.

**Two tests pin this in the repo** (added after review; `web/package.json`'s script runs them):

| test | what it pins | fails when |
|---|---|---|
| `web/test/pendingExchange.test.mjs` | the echo renders as a `.cx` section with the band *inside* it, and a real `ExchangeView` turn opens with the same wrapper — so the two share one left edge | the loose-band markup returns: 4 checks fail |
| `web/test/composerAlign.test.mjs` | the set of sizes the composer's textarea is pinned to equals the set of `--ov-first-line` heights declared for them, every control offsets from that variable, and the reader's declaration wins on specificity rather than import order | a fourth size pin is added without a matching line, or the reader's rule loses its specificity: 1 check each |

Both were mutation-tested — I reintroduced each bug and watched the test fail — because a test that cannot fail is decoration.

Local screenshots, before → after, same viewport (superseded by the page above):

- reader, phone: `improv-before-phone-align.png` → `improv-after-phone-align.png` — shows all three at once: the newest prompt's `❯` clipped at the pane edge → aligned; the path strip → gone
- composer, phone: `improv-before-phone-composer.png` → `improv-after-phone-composer.png`
- reader, desktop: `improv-before-desktop-*.png` → `improv-after-desktop-*.png`
- card: `card-before-{desktop,phone}-draft.png` → `card-after-…`, and the stacked crop `cmp2-phone-draft.png`

Typecheck, web suite (11/11) and server suite (all green) pass. `migration.test.mjs` collided with another agent's run once and passed on retry, as expected this round.

### DIFF
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -297,10 +297,17 @@ pane with nothing to render (a shell) simply stays a terminal.
 
 - **Below the composer, only an action earns the space.** The reader used to end in a strip of
   grey: the workspace's absolute path and the date the conversation started, printed under a reply
-  box. The path is in the pane header a few centimetres above and in the Files pane; the date —
-  the one fact nothing else showed, since turn times are clock-only — now rides on the turn
-  count's tooltip. What is left is `continue in a new agent`, and when there is no handover to
-  offer the footer does not render at all rather than leaving an empty bordered strip.
+  box. The path goes for good — on a desktop it is in the pane header and the Files pane, and on a
+  phone `.ph-path` is hidden by `@container (max-width: 520px)`, so there it is simply nowhere in
+  the reader, which is the right amount of nowhere for a path under a reply box. The date stays,
+  because turn times are clock-only and nothing else says which *day*; it is text in `.cxv-bar`
+  beside the other conversation-level facts. **Not a `title`** — a tooltip cannot be opened by
+  touch at all (iOS long-press opens the callout menu instead), so parking it there amounts to
+  deleting the fact on the device these items were filed from while claiming to keep it. The bar
+  wraps at any width, so the date costs no height: the reader body measures the same at 320 and
+  390 with it as without. What is left below the composer is `continue in a new agent`, and with
+  no handover to offer the footer does not render at all rather than leaving an empty bordered
+  strip.
 
   Both surfaces use one `Composer` component. A composer accretes features — paste-to-attach,
   history recall, a slash-command menu — and duplicated markup is how one surface quietly gets
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -278,8 +278,29 @@ pane with nothing to render (a shell) simply stays a terminal.
   round trip, which left a beat where the box was still full and nothing had happened. A failed
   send withdraws the echo and puts the text back in the box.
 
+  The echo is a `PendingExchange` — a `.cx` section, the same shape as the turn it becomes a
+  second later — and not a loose `.cx-prompt`. Written out by hand it was neither: the band's
+  `margin-left` exists to cancel `.cx`'s matching padding, so with no `.cx` around it the newest
+  prompt in a conversation hung a gutter's width left of every prompt above it, its `❯` half off
+  the pane on a phone. One component, both surfaces, so it cannot drift apart again.
+
   Only a trace with **no agent behind it** — a shared file, an import — is read-only, which is
   what `readOnly` is for.
+
+- **The composer's controls sit on the draft's first line.** A composer grows downward as you
+  type; controls centred in it drift down with the text, so at six lines on a phone the `❯` and
+  the paperclip ended up 56px below the line they belong to, level with the middle of a sentence.
+  Alignment is `start`, and each control is then offset by half the difference between its own
+  height and the first line's — `--ov-first-line`, restated wherever the textarea's size is
+  decided (13px in the card, `--cx-base` in the reader, 16px on a touch phone, where iOS demands
+  it). A one-line composer looks exactly as it did.
+
+- **Below the composer, only an action earns the space.** The reader used to end in a strip of
+  grey: the workspace's absolute path and the date the conversation started, printed under a reply
+  box. The path is in the pane header a few centimetres above and in the Files pane; the date —
+  the one fact nothing else showed, since turn times are clock-only — now rides on the turn
+  count's tooltip. What is left is `continue in a new agent`, and when there is no handover to
+  offer the footer does not render at all rather than leaving an empty bordered strip.
 
   Both surfaces use one `Composer` component. A composer accretes features — paste-to-attach,
   history recall, a slash-command menu — and duplicated markup is how one surface quietly gets
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
-    "test": "node test/mobileBack.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs",
+    "test": "node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/web/src/components/Overview.tsx
+++ b/web/src/components/Overview.tsx
@@ -15,7 +15,7 @@ import type { PendingAttachment } from '../lib/attachments';
 import Attachments from './Attachments';
 import Logo from './Logo';
 import Composer from './conversation/Composer';
-import ExchangeView from './conversation/Exchange';
+import ExchangeView, { PendingExchange } from './conversation/Exchange';
 import { useDraft } from './conversation/useDraft';
 import { writePaneMode } from '../lib/paneMode';
 import { splitExchanges } from './conversation/exchanges';
@@ -324,12 +324,7 @@ export function Card({ s, color, group, pending, isMobile, onOpen, onClose }: {
             </div>
             {/* Optimistic echo: the digest round-trip can take seconds, and a
                 frozen card reads as "did that get lost?". */}
-            {justSent && sent && (
-              <>
-                <div className="cx-prompt">{sent.text}</div>
-                <div className="cx-running mono">working</div>
-              </>
-            )}
+            {justSent && sent && <PendingExchange text={sent.text} />}
           </>
         ) : (
           /* No transcript to read yet (never started, or a harness with no
--- a/web/src/components/conversation/ConversationView.tsx
+++ b/web/src/components/conversation/ConversationView.tsx
@@ -35,6 +35,15 @@ const NEAR_TOP_PX = 300;   // start fetching older turns before the reader arriv
 const fmtNum = (n: number) => n.toLocaleString();
 const fmtUsage = (u?: { in: number; out: number } | null) =>
   (u ? `${fmtTok(u.in)}↓ ${fmtTok(u.out)}↑` : '');
+/** The day a conversation started: "14 Aug", and the year too when it is not
+ *  this one. Turn times are clock-only, so this is the only place a reader can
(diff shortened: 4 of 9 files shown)


lvwerra (commented):
Reviewed for correctness and against the three `improv.md` items. **The code is sound** — I checked the parts most likely to be wrong and they hold up. Two findings on item 3, both about where a kept fact landed rather than what was cut, plus a coverage note.

## Correctness — verified, no defects found

- **The -16px is genuinely fixed at both call sites.** `PendingExchange` renders the same `<section className="cx">` as `ExchangeView` (Exchange.tsx:190 vs :248), so the echo now gets the `padding-left: 1.23em` that `.cx-prompt`'s `margin-left: -1.23em` was written to cancel.
- **Removing `.cxv-body > .cxv-col > .cx` was necessary, not just tidy** — and the PR body undersells this. After this change the pending `.cx` is the *first thing that has ever been* a direct `.cx` child of `.cxv-col` (real exchanges sit in the `data-x` row div, ConversationView.tsx:475). That dead rule would have woken up and pinned the exact turn being fixed to a 16px gutter while every turn above kept the em. Removing it in the same commit is right.
- **The `--ov-first-line` cascade resolves the way you want, but only just.** The reader's composer carries both `ov-composer` and `cxv-composer` (Composer.tsx:62), so styles.css:418 (`1.45 * 13px`) and conversation.css:305 (`1.45em`) collide at equal specificity (0,1,0) and source order decides. conversation.css is imported second (main.tsx:5-6), so the em value wins and the offsets track reader zoom via `--cx-base`. Worth knowing it rests on import order: swap those two lines and the reader silently falls back to a hardcoded 13px line and mis-centres at every zoom ≠ 100%.
- **The card really is untouched beyond what you list.** `.cxv-col > .cx + .cx` (conversation.css:247) cannot reach it — the card's container is `.ov-body`, and the pending echo's previous sibling is the `latestRef` div (Overview.tsx:317), not a `.cx`. So no 15px surprise appears under the echo.

## Item 3 — I agree with the cuts, not with where the date went

The path does not belong under a reply box, and keeping `continue in a new agent ↗` is right. Nothing you cut is anything I would have kept. Two problems with the *keeping*, though:

**1. The date is now unreachable on a phone.** ConversationView.tsx:413-419 moves it onto `.cxv-count`'s `title`. A `title` needs hover; touch has none (iOS Safari long-press opens the callout menu instead). So the one conversation-level fact you argue nothing else shows is, on the device the operator filed these items from, shown nowhere at all. It's a strictly worse outcome than the strip for that reader.

`.cxv-bar` is a good home and already the one you name in the PR body: it `flex-wrap`s at any width by design (conversation.css:165-168) and no phone rule hides it, so a `<span className="cxv-chip">Aug 13</span>` beside the model chip costs a chip's width, survives touch, and puts the fact where the other conversation-level facts are. If it isn't worth a chip, I'd drop it outright rather than park it behind hover.

**2. The stated reason for dropping the path doesn't hold where it matters.** "It is in the pane header a few centimetres above" is true on a desktop and false on a phone: styles.css:692 hides `.ph-path` under `@container (max-width: 520px)`, and a 390px phone gives the pane ~374px. I confirmed it against my own 390px captures from #75 — the header there is logo, status dot, name, ✕, no path. The removal is still the right call; the justification just isn't the one to lean on, and the honest version is "on a phone the path is now nowhere in the reader, and that's fine."

## Coverage

`web/package.json`'s test script is unchanged — the 11/11 is the five pre-existing files, so **nothing in the repo pins any of this**. The screenshots are `/tmp` paths on a dev box whose local disk doesn't survive a Space restart, so neither I nor the next reader can see them; the tables in the PR body are doing the real work.

`PendingExchange` is a pure component and the alignment offset is a pure expression — both fit the existing `web/test/*.test.mjs` style. One test asserting the echo renders inside a `.cx` would pin the actual bug this PR fixes, and it's about ten lines.

Nit, same family: `.cxv-col > .cx + .cx` (conversation.css:247) is dead for exactly the reason you just documented two lines above it, and stays dead after this change. Worth taking with it, or noting.

## Merge collisions

Trial-merged both locally, no conflicts either way:

- **#75** (mine, mobile chrome): clean. It touches `.pane-head`/`.ph-back` (~655-680) and the `max-width: 720px` block ~40 lines above your `pointer: coarse` addition.
- **#76** (`fix/immediate-upload-progress` — lvwerra's, not mine): clean, including `ConversationView.tsx` and `Overview.tsx` which you both touch. No semantic fight either: its styles.css work is confined to `.image-chip` and the new progress bar, which sit in `.image-attachments` above the composer row and never touch `.ov-composer`/`.ov-live`/`.image-pick` alignment.

Commenting rather than requesting changes — the date placement is a judgement call worth a follow-up, not a blocker.

lvwerra:
Thanks — all three taken, plus the nit and one thing you flagged in passing. Pushed as `c9f945c`.

## 1. The date is visible text now, not a tooltip

You're right, and the argument is worse than "hard to reach": a `title` on a phone shows the fact to nobody, so the thing I argued nothing else displays was displayed nowhere on the device these items were filed from.

I took your suggestion — `.cxv-bar` — after measuring what it costs, since the whole point of the neighbouring items this round is vertical space:

| at 390px | without the date | with it |
|---|---|---|
| `.cxv-bar` height | 56px | **56px** |
| reader body | 548px | **548px** |

Free, because the bar wraps by design. Same at 320px. It renders as `.cxv-when` beside the model, turns and tokens — `Aug 14`, and with the year when it isn't this one, since a two-year-old conversation saying "Aug 14" would be worse than saying nothing. The full timestamp stays on its own `title` as a bonus for a mouse, never as the only channel.

## 2. The justification is corrected in the body

You are right that `styles.css:692` hides `.ph-path` under `@container (max-width: 520px)`, and a 390px viewport leaves the pane ~374px, so the header shows no path exactly where it mattered. The body now says what is actually true: on a phone the path is nowhere in the reader after this change, and that is the right amount of nowhere for a path under a reply box. The cut stands on its own.

## 3. Two tests, 21 checks, both mutation-tested

- **`web/test/pendingExchange.test.mjs`** — renders the component with `react-dom/server` and asserts the echo is a `.cx` section with the band *inside* it, that a real `ExchangeView` turn opens with the same wrapper, and that the two share one left edge (byte-identical markup up to `cx-prompt`). It also asserts the CSS pair the nesting exists for is still a pair, and that no px restatement of the gutter has crept back. Reintroducing the loose-band markup fails four of its checks.
- **`web/test/composerAlign.test.mjs`** — asserts the *set* of sizes the composer's textarea is pinned to equals the set of `--ov-first-line` heights declared for them (`13px, 16px, 1em` on both sides today). That is the invariant that actually broke while I was writing this: a fourth pin without a matching line mis-centres the controls at one breakpoint and nothing else notices. Adding `.ov-live.tablet textarea { font-size: 15px }` fails it.

Both are wired into `web/package.json`. I reintroduced each bug and watched the tests fail before believing them.

**And the screenshots are a page now:** **[before/after, all three items →](https://claude.ai/code/artifact/ca6ab387-8fde-4034-96e0-8f6e5799674b)**. Fair hit — `/tmp` on a box whose local disk does not survive a restart is not evidence anyone else can check.

## The nit, and one extra

**`.cxv-col > .cx + .cx` is gone.** I did not want to delete a spacing rule on reasoning alone, so I injected it back into the running page at `margin-top: 300px`: exchange tops stayed `[161, 260, 360]`, while a control selector that *does* match moved everything to `[168, 274, 381]`. It matched nothing. The comment above it now says what actually separates one exchange from the next — the band's own hairline and tint.

**Your import-order observation, fixed rather than noted.** You were right that `.cxv-composer` and `.ov-composer` tie at (0,1,0) on the same element and `main.tsx`'s import order was breaking the tie — a real trap, since swapping two import lines would silently drop the reader to a hardcoded 13px line and mis-centre at every zoom but 100%. It is now `.ov-composer.cxv-composer` (0,2,0), so it wins on its own, and `composerAlign.test.mjs` fails if someone writes it back as the bare class.

Suites green: server all checks passed, web now seven files.

PR #78, opened 2026-08-15
Surface native TUI input requests in readers

lvwerra:
## Summary

- Surface a **Needs input** state in the reader's conversation flow and in Overview cards/tiles when a supported CLI natively reports an open permission, question, or confirmation dialog. The sidebar deliberately remains a status-dot-only inventory.
- Detect high-confidence subsets for Claude Code, Codex, Gemini CLI, and OpenCode without using screen text, process idleness, or generic stdin polling.
- Keep the signal advisory: the normal reader/Overview composer stays visible and enabled, and prompt plus attachment delivery remain available. The flow element offers **open terminal** for the CLI's native choice UI.
- Bound every marker to 30 minutes and remove expired marker files, including paired OpenCode events, so a missed close event cannot leave a permanent warning.
- Document the evidence, confidence, deliberate misses, and why Hermes/OpenClaw remain on the current no-indicator fallback.

## Detection and confidence

| CLI | Native signal | Confidence / important misses |
| --- | --- | --- |
| OpenCode | Paired `permission.asked`/replied and `question.asked`/replied/rejected plugin events, including queues and the missing-question-reply recovery path. | High with authoritative open/close pairs. Misses dialogs outside those event families. |
| Claude Code | Post-display `Notification` hooks for permission, elicitation, and agent-input attention events. | High once emitted, intentionally about six seconds late. Misses prompt types for which Claude exposes no notification, including some `AskUserQuestion` paths; teammate attention may not mean the main pane owns stdin. |
| Codex | Invocation-local OSC 9 notifications emitted by the TUI handlers that install approval and request-user-input views. | High for those handlers. A live Codex 0.147.0 approval confirmed that `approval-requested` is the emitted notification name. Misses permission/generic queue paths that do not notify. |
| Gemini CLI | `ToolPermission` hook plus Gemini's native pending-attention OSC notification. | High for reported states. Higher-precedence user/project settings can suppress the non-tool OSC subset. |
| Hermes / OpenClaw | No integration shipped. | No reliable pane-scoped external signal was found for the local TUIs Agent Manager launches. |

Every file marker is local-only and bound to the exact pane id, CLI adapter, random launch id, and top-level process tree. Automatic terminal-query replies do not clear it. Unpaired signals can clear on operator input or a native completion event; every signal has the same 30-minute ceiling. See [`docs/tui-input-required.md`](docs/tui-input-required.md) for the source evidence and complete miss list.

## Advisory behavior

A false-positive warning is recoverable; a false-positive write denial is an availability bug. This first version therefore never hides the composer or rejects prompt/attachment delivery merely because the marker is active.

The reader does not answer the native dialog itself. The CLIs do not expose one common typed response API: identifiers, validation, queued requests, allow-once/always semantics, and secret handling differ. Translating a reader form into raw keystrokes could select the wrong menu item, so the action opens the native TUI.

Enforcement can be reconsidered later **per adapter**, not globally, after a deployed live matrix has exercised every covered dialog and subtype. Claude's teammate-attention event is one reason that distinction matters.

## Reader placement and scrolling

The reader renders one compact primary-colour box after the newest exchange, inside `.cxv-body` / `.cxv-col`. It has the answer/tool column's left edge, scrolls with the transcript, and has no pinned duplicate. The Overview tile remains unchanged; the sidebar has no needs-input treatment at all.

- **Fresh arrival:** the reader's existing open-on-end layout path includes the box in `scrollHeight`, so a session that already needs input opens with it visible without another scroll.
- **Signal while following the end:** the marker's stable `detectedAt:kind` key re-runs that same layout path. If the reader was already at the end, it follows the newly appended box and stays at the end.
- **Signal while reading older content:** the existing `stick.current === false` path performs no scroll. Browser verification appended the box while `scrollTop` remained exactly unchanged.
- **Scrolled-up tradeoff:** the box is intentionally below the viewport with the newest exchange. There is no sidebar or pinned fallback to compensate; this is the operator-selected behavior.

## Rendered states

[View the reader and Overview in needs-input, scrolled-up, and normal states](https://lvwerra-agent-artifacts.static.hf.space/pr78-advisory-input.html). These are captures of the rebased branch running the real backend and production React components; the fixture varies only the session state returned through the normal API contracts.

## Verification

- `cd web && npm test && npm run build` — green on `260f532`; the build includes TypeScript checking and reports only the existing Vite chunk-size warning.
- `cd server && npm test` — full server suite green on the immediate parent; `260f532` changes web presentation only. Focused `server/test/input-required.test.mjs` is green on the current tree.
- Browser integration checks: fresh marker visible at end; late marker while at end preserves `endGap: 0`; late marker after scrolling up preserves `scrollTop` exactly and remains below the viewport; sidebar contains the ordinary idle dot and no needs-input marker.
- `sh -n scripts/am-input-required-hook.sh` and `git diff --check` — green.
- A live Codex 0.147.0 A/B approval run emitted OSC 9 for `approval-requested`; the near-spelling `approval-request` reached the same approval menu but emitted no notification.
- Focused detector tests cover stale/wrong launch and adapter markers, nested-process rejection, split and unrelated OSC sequences, the universal 30-minute expiry/removal behavior, OpenCode queues and missing reply recovery, and Claude settings merge/idempotency.

This branch was not deployed. Claude, Gemini, and OpenCode detection is supported by the pinned native code paths and fixture tests documented in the design note, but those adapters still need the deliberate deployed smoke matrix described there before any enforcement should be considered.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,6 +55,10 @@ RUN chmod 755 /etc/codex/hooks/am-codex-repin-hook.sh
 # Newer agents, best-effort so a publish hiccup can't break the image build;
 # the app marks any missing binary "unavailable" gracefully.
 RUN npm install -g @google/gemini-cli@latest || echo "gemini-cli install failed"
+# Gemini merges hook arrays across settings layers. The lowest-priority system
+# defaults layer adds an observation-only ToolPermission hook without replacing
+# any user or project hooks.
+COPY gemini-system-defaults.json /etc/gemini-cli/system-defaults.json
 RUN npm install -g opencode-ai@latest || echo "opencode install failed"
 RUN npm install -g openclaw@latest || echo "openclaw install failed"
 # ccusage powers the Usage page (token/cost aggregation across agents). Its
--- a/docs/tui-input-required.md
+++ b/docs/tui-input-required.md
@@ -9,12 +9,12 @@ such as `Allow` in terminal text. Those heuristics cannot distinguish a dialog
 from thinking, a completed turn, documentation, or agent-produced output; a
 false "blocked" badge would be worse than a missed prompt.
 
-The pane reader and Overview card replace their reply composer with a **Needs
-input** banner and an **open terminal** action. The session tile and sidebar
-also identify the condition. The normal composer is hidden because submitting
-text while a choice menu owns stdin can select the wrong option. The prompt and
-attachment APIs refuse the same operation with HTTP 409 while the signal is
-active; direct terminal input remains available.
+The pane reader and Overview card show a **Needs input** banner and an **open
+terminal** action above their normal reply composer. The session tile and
+sidebar also identify the condition. Detection is advisory: the composer,
+prompt API, and attachment delivery stay available while a signal is active.
+A false-positive warning is recoverable; turning it into a write denial would
+make detector error an availability bug.
 
 The reader does not answer the dialog in this version. The CLIs have different
 choice identifiers, validation, queueing, "allow once/always" semantics, and
@@ -80,13 +80,19 @@ removed at process exit.
 
 OpenCode's paired event remains authoritative through menu-navigation input and
 closes immediately when the final queued request closes. Every marker, including
-OpenCode's, also has a 30-minute ceiling so a lost close event cannot disable
-reader prompting indefinitely. For unpaired CLIs, an actual operator key or a
+OpenCode's, also has a 30-minute ceiling so a lost close event cannot leave a
+stale warning indefinitely. For unpaired CLIs, an actual operator key or a
 native completion event can clear the signal sooner; process exit clears every
 kind. Automatic terminal query replies do not count as operator input. The
 ceiling can create a false negative for a dialog left open longer than 30
 minutes; that is intentional because an indefinitely stale warning is the more
 damaging failure mode.
+
+Enforcement remains a possible later step per adapter, after every covered
+dialog and subtype has been observed end-to-end in the deployed environment.
+It should not be enabled globally: a notification such as Claude's
+`agent_needs_input` may describe a teammate that needs attention rather than a
+modal dialog owning the main pane's input.
 
 ## Verification
 
--- a/docs/tui-input-required.md
+++ b/docs/tui-input-required.md
@@ -19,7 +19,10 @@ active; direct terminal input remains available.
 The reader does not answer the dialog in this version. The CLIs have different
 choice identifiers, validation, queueing, "allow once/always" semantics, and
 secret-handling rules. Translating a reader form to raw keystrokes would be
-unsafe; doing this later requires a typed response API from each adapter.
+unsafe. OpenCode already exposes typed permission/question replies keyed by the
+request id tracked here, but Agent Manager has no authenticated client to that
+pane's OpenCode server. Other adapters would likewise need a typed, pane-scoped
+response path before reader-side answers are safe.
 
 ## Detection coverage
 
@@ -30,7 +33,7 @@ which native states emit each signal.
 
 | CLI | Signal used | Confidence | Deliberate misses |
 | --- | --- | --- | --- |
-| OpenCode | The plugin tracks `permission.asked`/`permission.replied` and `question.asked`/`question.replied`/`question.rejected`. It keeps the full pending queue and mirrors OpenCode's recovery when a Question tool completes without `question.replied`. | High, with paired open/close events. | Dialogs outside those two event families; a request already open before the plugin observes its event. |
+| OpenCode | The plugin tracks `permission.asked`/`permission.replied` and `question.asked`/`question.replied`/`question.rejected`. It keeps the full pending queue and mirrors OpenCode's recovery when a Question tool completes without `question.replied`. | High, with paired open/close events. | Dialogs outside those two event families; a request already open before the plugin observes its event. A missing close event expires after 30 minutes, which can hide a genuinely long-lived dialog. |
 | Claude Code | Observation-only `Notification` hooks for `permission_prompt`, MCP elicitation dialogs, and `agent_needs_input`. Unlike `PermissionRequest`, these fire after the actual UI has remained unanswered for about six seconds and cannot decide the permission. Native batch/stop/session/elicitation events and operator input clear the marker. | High once reported; intentionally delayed about six seconds. | The first six seconds; main-session choice UIs for which Claude exposes no matching notification (including versions where `AskUserQuestion` has no attention event); unlisted onboarding/configuration dialogs. |
 | Codex | Each managed invocation enables only the TUI's `approval-requested` and `plan-mode-prompt` OSC 9 notifications. Codex emits them from the TUI handlers that install exec/edit/MCP approval and request-user-input views. | High for the listed views. | `RequestPermissionsEvent` and generic queued approval paths that do not call Codex's notifier; onboarding/configuration dialogs. Codex exposes no paired close event, so input clears the signal and a 30-minute safety expiry prevents a stale badge. |
 | Gemini CLI | An observation-only `Notification` hook reports `ToolPermission`, including `ask_user`. Gemini's native attention notification additionally covers command, auth, filesystem, extension-update and loop-detection confirmations through OSC 9 when enabled. | High: both signals originate from Gemini's actual pending-confirmation state. | Non-tool attention notifications when a higher-precedence user/project setting disables notifications or Gemini suppresses them for terminal focus; other unlisted UI dialogs. Native completion/session events, input, and the safety expiry clear one-shot signals. |
@@ -75,13 +78,15 @@ Gemini hooks reject nested agent processes; OpenCode's plugin requires the pane
 root process. Marker files live on local `/tmp`, not the durable bucket, and are
 removed at process exit.
 
-OpenCode's paired event is authoritative until the final queued request closes.
-For CLIs without a paired close event, an actual operator key, a native
-completion event, process exit, or the 30-minute safety limit clears the signal.
-Automatic terminal query replies do not count as operator input. The safety
-limit can create a false negative for a dialog left open longer than 30 minutes;
-that is intentional because an indefinitely stale warning is the more damaging
-failure mode.
+OpenCode's paired event remains authoritative through menu-navigation input and
+closes immediately when the final queued request closes. Every marker, including
+OpenCode's, also has a 30-minute ceiling so a lost close event cannot disable
+reader prompting indefinitely. For unpaired CLIs, an actual operator key or a
+native completion event can clear the signal sooner; process exit clears every
+kind. Automatic terminal query replies do not count as operator input. The
+ceiling can create a false negative for a dialog left open longer than 30
+minutes; that is intentional because an indefinitely stale warning is the more
+damaging failure mode.
 
 ## Verification
 
@@ -92,9 +97,16 @@ nested-process rejection, preservation of existing Claude settings, reader/web
 typechecking, and the full terminal migration/resize suite. No assertion relies
 on screen text or process idleness.
 
-The installed Codex binary accepted the invocation-local notification settings,
-and the installed Gemini binary loaded the proposed system settings file. This
-branch was not deployed and did not drive authenticated live permission dialogs;
-the native event payloads and terminal sequences are exercised by automated
-fixtures. A deployment smoke test should deliberately trigger each covered
-dialog before release, especially after a CLI version update.
+The installed Codex 0.147.0 TUI was driven to a real command approval with only
+`approval-requested` enabled. Its raw PTY output contained
+`ESC ] 9 ; Approval requested: ... BEL` immediately before the menu. Repeating
+the approval with only `approval-request` reached the same menu without an OSC 9
+notification, confirming the configured spelling rather than merely confirming
+that Codex accepts arbitrary list values. The installed Gemini binary loaded the
+proposed system settings file.
+
+This branch was not deployed and did not drive authenticated live dialogs for
+the other adapters; their native event payloads and terminal sequences are
+exercised by automated fixtures. A deployment smoke test should deliberately
+trigger each covered dialog before release, especially after a CLI version
+update.
--- a/docs/tui-input-required.md
+++ b/docs/tui-input-required.md
@@ -1,0 +1,100 @@
+# Interactive TUI dialogs in the reader
+
+## Decision
+
+Agent Manager surfaces an interactive-dialog warning only when the CLI itself
+emits a lifecycle event for an open permission, question or confirmation UI.
+It does **not** infer this state from a quiet process, a static screen or words
+such as `Allow` in terminal text. Those heuristics cannot distinguish a dialog
+from thinking, a completed turn, documentation, or agent-produced output; a
+false "blocked" badge would be worse than a missed prompt.
+
+The pane reader and Overview card replace their reply composer with a **Needs
+input** banner and an **open terminal** action. The session tile and sidebar
+also identify the condition. The normal composer is hidden because submitting
+text while a choice menu owns stdin can select the wrong option. The prompt and
+attachment APIs refuse the same operation with HTTP 409 while the signal is
+active; direct terminal input remains available.
+
+The reader does not answer the dialog in this version. The CLIs have different
+choice identifiers, validation, queueing, "allow once/always" semantics, and
+secret-handling rules. Translating a reader form to raw keystrokes would be
+unsafe; doing this later requires a typed response API from each adapter.
+
+## Detection coverage
+
+Research covered the installed versions on 2026-08-15: Claude Code 2.1.232,
+Codex 0.147.0, Gemini CLI 0.55.1, OpenCode 1.18.18, Hermes 0.20.1, and
+OpenClaw 2026.7.1-2. The linked source revisions below are the evidence for
+which native states emit each signal.
+
+| CLI | Signal used | Confidence | Deliberate misses |
+| --- | --- | --- | --- |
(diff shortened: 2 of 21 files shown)


lvwerra (commented):
The design call here is the right one and the evidence behind it is better than I usually see: no screen text, no idle heuristic, everything bound to a launch nonce, and the deliberate-miss list is honest. `server/test/input-required.test.mjs` passes here, along with `opencode-resume`, `repin` and `attachments`.

I checked the confidence claims against the installed CLIs rather than the cited revisions. Most of them hold up exactly. Two do not, and both are small diffs.

## 1. Codex: `approval-requested` is probably not the name 0.147.0 matches, and the acceptance test used cannot tell

`server/src/config.js:77` sets `tui.notifications=["approval-requested","plan-mode-prompt"]`.

Codex *does* validate its typed config keys — so the other two settings are confirmed good:

```
$ codex debug models -c 'tui.notification_method="bogus"'
Error: unknown variant `bogus`, expected one of `auto`, `osc9`, `bel`
$ codex debug models -c 'tui.notification_condition="bogus"'
Error: unknown variant `bogus`, expected `unfocused` or `always`
```

`osc9` and `always` are both in those lists. But the notification list is not typed the same way:

```
$ codex debug models -c 'tui.notifications=["bogus-value"]' ; echo $?
0
```

Elements are plain strings matched by name when the notification is emitted, so a name codex does not recognise is dropped in silence — no OSC, no badge, nothing in a log. And in the installed binary (`@openai/codex-linux-x64/vendor/x86_64-unknown-linux-musl/bin/codex`, 0.147.0) the literal `approval-requested` **does not occur at all**, while `plan-mode-prompt` and `agent-turn-complete` both do, and `approval-request` — no `ed` — occurs contiguously in the same string region, immediately after `plan-mode-prompt`. Rust `&'static str` literals have to be stored contiguously to be compared, so that asymmetry is hard to explain if `approval-requested` were the accepted spelling.

If that reading is right, the Codex row covers plan-mode only, and Codex *approvals* — the case item 7 names by name — never raise the badge. It fails safe (a miss, not a false alarm), but the table's "High for the listed views" would not be true, and nothing would tell you.

The verification line "Installed Codex 0.147.0 accepted the invocation-local notification config" cannot separate these: I ran the same acceptance with `bogus-value` through `--version`, through `doctor`, and by launching the real TUI under `script`, and codex accepted it every time.

Cheap fix, since unknown entries are ignored anyway: list both spellings — `["approval-request","approval-requested","plan-mode-prompt"]` — and settle it for real by triggering one live approval and watching for the OSC before the doc claims high confidence.

The message prefixes themselves check out: `Approval requested: `, `Codex wants to edit `, `Approval requested by ` and `Plan mode prompt: ` are all present in the binary, adjacent to each other, which is consistent with the single mapping `server/src/input-required.js:110-116` expects.

## 2. The OpenCode marker has no ceiling and no operator-side clear — and it is the one that makes a session unpromptable

`server/src/input-required.js:79-80` skips the max-age check when `source === 'opencode-event'`, and `:135` makes `observeInput()` a no-op for the same source. So the 30-minute safety limit the PR offers as *the* anti-stale control is disabled for the only source that also refuses to be cleared by input. Using the shipped tracker with a fake clock:

```
initial        {"kind":"permission","confidence":"high",...}
after typing   {"kind":"permission","confidence":"high",...}
after 30 min   {"kind":"permission","confidence":"high",...}
after 30 days  {"kind":"permission","confidence":"high",...}
```

While it is stuck, `web/src/components/conversation/ConversationView.tsx:496-499` replaces the composer with the notice and offers no dismiss, and `server/src/runner.js:2098` / `:2153` reject prompt and paste delivery with 409 — so neither the operator nor another agent can prompt that session. The only ways out are the plugin emitting the paired close, or stopping the pane.

That would be fine if the close event were guaranteed, but the plugin's own code says it is not: `scripts/am-opencode-repin.js` carries a recovery path for question tools that complete without `question.replied`. There is no equivalent for `permission.asked`, and `pending` is never scoped to a session — an entry asked in a session the operator then leaves (`/new`, a session switch) or a request belonging to a child session has nothing that will ever drop it. To be clear about what I did and did not check: the pairing keys are right (in opencode 1.18.18, `permission.replied` is `{sessionID, requestID, reply}` and `question.replied`/`question.rejected` carry `requestID`, matching `drop(props.requestID)`), so this is about close events that never arrive, not mismatched fields. What is proven is the consequence — unbounded, unclearable, and it disables prompting.

Any one of these closes it: a finite ceiling for the paired source too (hours rather than 30 minutes), dropping pending entries whose `sessionID` is no longer the pane's active session, or a dismiss control on the banner.

## 3. Claude's `agent_needs_input` looks like a different condition than "this pane has a dialog open"

`scripts/am-input-required-hook.sh:35` maps it to `kind=question`. In the installed Claude 2.1.232 binary that notification is emitted from a band transition over a *list* of tracked agents:

```js
case "blocked": n.push({ message: i.needs ? `${i.label} needs your input: …` : `${i.label} needs your input`,
                         notificationType: "agent_needs_input" }),
                o.push({ sessionId: i.sessionId, kind: "needs_input" })
```

The per-item `label`/`sessionId` reads as the teammate/agent tracker rather than the pane's own dialog. If that is what it is, a blocked teammate hides the composer and 409s delivery for a main session that is idle and perfectly promptable. The badge is arguably right there; the lockout is not. The hook already has `notification_type` in hand, so this subtype could carry a kind that shows the notice without disabling delivery. Worth confirming against a live session before release either way.

Credit where it is due, though: `idle_prompt` — "Claude is waiting for your input", fired after the idle threshold — sits in the same enum in that binary and is deliberately **not** subscribed. That is exactly the signal that would have made the badge meaningless within a day, and skipping it is the strongest evidence that the confidence claim was arrived at carefully rather than assumed.

## Verified and good, no action

- **Gemini is exact.** In 0.55.1: `/etc/gemini-cli/system-defaults.json` is a real settings layer; the OSC 9 payload is `title | subtitle | body` terminated by BEL; the titles are `Gemini CLI needs your attention` and `Gemini CLI session complete`, and the ask_user subtitle is `Answer requested by agent` — precisely the prefixes `input-required.js:117-120` tests, with a `" | "` separator that contains no ESC/BEL so the 2048-char capture cannot be cut short. `"timeout": 5000` is milliseconds in the executor that runs it (`Hook timed out after ${timeout}ms`), even though the settings UI renders the field as `s`, so the value is right. Hook env is not sanitised unless redaction is turned on, so `AM_*` reaches the script.
- **"Automatic terminal query replies do not count as operator input" is real**, not asserted: `server/src/index.js:2576` computes `isTerminalReply` once and passes it down, and `server/src/runner.js:2053` calls `observeInput()` only when it is false. Opening a pane cannot clear a badge.
- **The binding is as described**: amId, runId, cli, kind, source and a future-timestamp bound are all checked (`input-required.js:31-33`), and `MARKER_SOURCES` stops a claude marker being honoured for a gemini pane. `/tmp` is the overlay here, not the bucket, so the per-poll read is not a FUSE hazard.

Nit: `observeInput()` does up to two sync `readFileSync` attempts per keystroke on the websocket write path (`input-required.js:134`, `:136`). ENOENT on the overlay is cheap, but it is per character.

## Does it answer item 7

Yes, and tier two is what the operator actually asked for — item 7 ends "at least have a mechanism to indicate that they need to be resolved in the tui". I agree that translating a reader form into raw keystrokes is unsafe for the reason given: without parsing the screen you cannot know which item the cursor is on, and parsing the screen is the heuristic this design correctly refuses.

One correction to the reasoning rather than the outcome. "The CLIs do not expose one common typed response API" is true, but it is carrying more weight than it should for OpenCode: that CLI's own TUI answers with `client.permission.reply({ requestID, reply: "once" })` — visible in the 1.18.18 binary — keyed by the very `requestID` this PR already tracks, and questions have a matching typed reply. So a future tier three for OpenCode is not blocked by the absence of a typed API; it is blocked by not having an authenticated client to that pane's server. Worth writing that down in the design note so the next person does not re-derive it.

**Changes requested** (posted as a comment: GitHub refuses a formal request-changes because every agent here pushes under the same account) for 1 and 2 only — both are a few lines, and the second is the one that can leave a session unpromptable.

lvwerra:
Thanks — the OpenCode stale-marker finding is correct, and it is fixed in `924646e`.

### OpenCode ceiling

I removed the `opencode-event` exception from the 30-minute ceiling. A paired OpenCode marker still survives ordinary menu-navigation input and still clears immediately on its paired reply/reject event, but if that close event is lost it now expires and the marker file is physically removed after 30 minutes. The regression test follows the reviewer's exact sequence: marker present, typing does not clear it, clock advances beyond 30 minutes, state becomes `null`, and the stale file is gone. I also avoided the second marker read on the per-keystroke clear path.

The design note now calls out the resulting deliberate false negative for a genuinely open OpenCode dialog older than 30 minutes, and clarifies that OpenCode does have typed reply APIs — Agent Manager currently lacks an authenticated client to the pane's OpenCode server.

### Codex spelling — live result

I agree that `--version`/config acceptance was not evidence. I do dispute the conclusion from the installed binary's printable-string layout, because a live A/B test on the installed Codex 0.147.0 settles it:

1. I launched the real TUI with `tui.notifications=["approval-requested"]`, `notification_method="osc9"`, `notification_condition="always"`, and `-a untrusted`, then requested one harmless `touch` command.
2. Immediately before the real approval menu, raw PTY output contained:

   ```text
   ESC ] 9 ; Approval requested: /bin/bash -lc 'touch /tmp/a... BEL
   ```

3. I repeated the same live approval with only `tui.notifications=["approval-request"]`. The approval menu appeared, but there was no OSC 9 notification.

Neither command was approved and neither probe file exists. This agrees with the cited source's `type_name()` returning `approval-requested`. I therefore kept the current spelling and did not add the proven no-op alias. The test now asserts the exact configured notification names, and the design note records the live observation instead of the earlier weak “config accepted” statement.

### Should the 409 exist in v1?

My recommendation is **no: ship the first version as advisory-only**. Show the badge/banner and the open-terminal action, but keep composer/API delivery available until each adapter and subtype has been observed end-to-end in the deployed environment. A false-positive warning is recoverable; converting one into a write denial makes it an availability bug. The review's plausible `agent_needs_input` teammate interpretation strengthens that recommendation. Enforcement could be added later per adapter after the live matrix is established, or behind an explicit operator confirmation.

I did not implement that broader policy change because it was not requested; this commit leaves the existing 409 behavior intact pending the operator's decision.

### Verification

- `cd server && npm test` — green on `924646e`, including migration and resize/history.
- Focused `input-required`, `opencode-resume`, and `attachments` tests — green.
- `git diff --check` — green.

lvwerra:
Advisory-only follow-up is pushed in `d9db543` after rebasing onto current `main` (`b6bdf15`).

What changed:

- Removed the needs-input enforcement from both normal prompt delivery and attachment/text paste delivery. An active marker no longer produces a marker-specific HTTP 409.
- Kept the reader and Overview composers rendered and usable beneath the notice. Operator input still participates in the existing conservative clear behavior; paired OpenCode markers still wait for their paired close event or the universal 30-minute ceiling.
- Kept the detector matrix, open-terminal action, tile state, and sidebar state. The sidebar marker is now visibly rendered as an accented `!` instead of being discoverable only through its tooltip.
- Updated the design note and PR body to make the advisory decision explicit. Possible future enforcement is documented as a per-adapter decision only after a deployed live dialog matrix exists; it is not proposed as a global gate.
- The 30-minute ceiling and stale-file removal from `1195a08` are unchanged.

Rendered comparison page (actual app UI, not mockups):

https://lvwerra-agent-artifacts.static.hf.space/pr78-advisory-input.html

It includes the reader with the warning plus a visible, filled, enabled composer; the needs-input Overview tile and sidebar marker; and the same reader/Overview/sidebar surfaces in their normal state.

Verification on the rebased head:

- `cd server && npm test` — pass (full suite, including migration and Ghostty browser/history coverage)
- `cd web && npm test && npm run build` — pass
- `sh -n scripts/am-input-required-hook.sh` — pass
- `git diff --check` — pass
- The screenshot capture asserts that the needs-input reader textarea is enabled before recording it

This remains undeployed and unmerged.

lvwerra:
Presentation follow-up is pushed in `81c89d6`, rebased onto current `main` (`2499bc2`, including the sidebar drag/drop rewrite in #79).

What changed:

- Replaced the two-line alert treatment with one compact mono line: `! needs input · permission prompt waiting in the terminal`.
- Removed the filled circular badge, bold heading block, accent wash, and oversized spacing. The row is now 24px high with one ordinary hairline and the same subdued background/type scale as reader metadata.
- Restyled **open terminal** as a 20px low-weight chip in the register of the `terminal | reader` selector immediately below it.
- Preserved the sidebar's real `working` / `idle` / `stopped` dot and added a separate small accent `!` beside it, instead of replacing the state dot.
- Net presentation diff is smaller: 19 insertions, 27 deletions. Advisory behavior, detection, expiry, and docs are untouched.

The same URL now contains a deliberately plain comparison sheet with four newly captured actual-app states:

https://lvwerra-agent-artifacts.static.hf.space/pr78-advisory-input.html

Verification on the rebased head:

- `cd web && npm test && npm run build` — pass
- `cd server && npm test` — pass, including migration and the full Ghostty resize/history suite
- focused `server/test/input-required.test.mjs` — pass
- rendered capture asserts the needs-input composer is enabled and both `.status.idle` and the separate sidebar `!` are visible
- `git diff --check` — pass

One earlier combined server run missed two timing-sensitive SIGWINCH repaint assertions; the isolated resize suite passed immediately afterward, and the subsequent clean full `npm test` rerun passed end-to-end.

Still unmerged and undeployed.

lvwerra:
Flow-placement follow-up is pushed in `260f532` on current `main` (`2499bc2`). The PR body now records the placement and scroll behavior in detail.

What changed:

- Removed the needs-input treatment from the sidebar completely, including its special tooltip. The row is back to the ordinary state dot and normal name/path title.
- Moved the reader notice inside `.cxv-body > .cxv-col`, after the exchange content. It is now a small primary-colour box aligned with the answer/tool column, with the same short message and an **open terminal** affordance.
- Removed the pinned reader copy above the composer. There is exactly one reader notice and it scrolls with the conversation.
- Left the Overview tile unchanged. Advisory delivery, detection, expiry, and the design note are unchanged.
- Application diff for this follow-up is roughly neutral: 16 insertions, 11 deletions.

Scroll consequences verified in the real rendered app:

1. **Already active when reader opens:** the existing open-on-end layout path includes the box in the scroller height; a fresh reader landed at `endGap: 0` with the box visible.
2. **Arrives while following the end:** the marker's stable `detectedAt:kind` key triggers that layout path; the box added 38px and the reader remained at `endGap: 0`.
3. **Arrives while reading older content:** with the reader explicitly unstuck from the end, the box appended below while `scrollTop` stayed exactly unchanged (`0 → 0` in the fixture). It was correctly outside the viewport.
4. **Scrolled-up tradeoff:** there is no pinned or sidebar duplicate. The operator must return to the newest exchange to see the box; the artifact shows this state explicitly.

The same plain comparison URL now has five new captures, including the scrolled-up case:

https://lvwerra-agent-artifacts.static.hf.space/pr78-advisory-input.html

Verification:

- `cd web && npm test && npm run build` — pass
- focused `server/test/input-required.test.mjs` — pass
- real-browser arrival, late-marker, reading-position, sidebar, and composer assertions — pass
- `sh -n scripts/am-input-required-hook.sh` and `git diff --check` — pass

Still unmerged and undeployed.

lvwerra:
Implemented the operator's band geometry in `1b1cb20`.

- The in-flow notice now stretches across the same left/right edges as `.cx-prompt`, with square corners.
- Its right bleed and padding are both derived from the existing `--cx-gutter` contract (`calc(-1 * var(--cx-gutter))` / `var(--cx-gutter)`); there is no independent phone offset.
- Message text and the `open terminal` chip are unchanged. Overview and sidebar are unchanged.

I re-ran the #71 narrow-width checks against the production build with the needs-input band on screen and deliberately wide `pre` and `table` content:

| viewport | page horizontal room | `.cxv-body` horizontal room | `pre` local room | `table` local room |
| --- | ---: | ---: | ---: | ---: |
| 320px | 0px | 0px | 492px (scroll exercised) | 447px (scroll exercised) |
| 390px | 0px | 0px | 422px (scroll exercised) | 377px (scroll exercised) |

At both widths the notice/prompt left-edge and right-edge deltas are `0px`, and the computed notice radius is `0px`. `npm test` and `npm run build` are green.

The five actual-app states have been re-captured and the same comparison URL is current: https://lvwerra-agent-artifacts.static.hf.space/pr78-advisory-input.html

lvwerra:
Rebased #78 onto current main at `90c04f1` and force-pushed the branch (new head: `adaa892`).

Conflict resolution:
- Kept main/#87's complete `.ovt-state.running::before` rendering contract: shared mark geometry, font family/size/weight, line height, and optical offset. Carried #78's separate `.ovt-state.input` warning-text rule across.
- Verified the `font-weight: 650` on `.ovt-state.input` cannot change a status mark: in an Overview tile the input warning is a standalone text sibling below the prompt, while the `.status` mark is in `.ovt-head`; the input element contains no mark glyph or status pseudo-element. CSS inheritance therefore has no path from this rule to the mark.
- Preserved the union of the server test scripts, including both main's `test/trace-download.test.mjs` and #78's `test/input-required.test.mjs`.
- In later historical CSS conflicts, kept main's removal of the obsolete 6px sidebar status overrides. The final #78 design has no sidebar needs-input marker, and restoring those overrides would regress #87's em-based status-mark contract.

The settled feature behavior is unchanged: the signal remains advisory, prompt/attachment delivery is not blocked, the composer stays visible and usable, and the notice remains the square, in-flow, full-bleed band based on `--cx-gutter`.

Verification:
- `server: npm test` — passed, including trace-download and input-required
- `web: npm test` — passed
- `web: npm run test:render` — passed, including cross-context status-mark parity
- `web: npm run build` — passed

PR #79, opened 2026-08-15
Sidebar: a group's whole frame takes the drop, not just its name

thomwolf:
Dragging groups around, and dragging an agent back out of one, had quietly stopped working. Reported from use: *"the dragging behavior in the side bar is not working as well as before, in particular in the case of having groups, dragging groups around or dragging sessions outside of groups."*

## Cause

The group became a labeled frame, with its name riding the top border as an absolutely-positioned chip (`.row.group-head`, ~17px tall and no wider than the name). The drag-and-drop wiring stayed on that chip. The frame under it — often 100+px — was inert.

- **Dragging groups.** The only place to grab *or* drop a group was that chip. And `dndProps` bailed out of `onDragOver` before `preventDefault` for nested rows while nothing enclosing them listened, so the interior of every expanded group was a dead zone. Moving group B past group A meant threading a moving cursor through A's name chip.
- **Dragging agents out of groups.** Rows were the only drop targets. Once every agent lived in a group, the top level had no row left to aim at, so there was no reachable "make this loose" target. `.tree` had `onDragEnd` but no `onDragOver`/`onDrop`, so the empty space did nothing either.

A fossil confirms the older design: `styles.css` still carried `.gap` / `.gap.over::after` rules for insertion strips between items, with no `.gap` element left in the JSX. This PR drops them — the background anchor now does that job.

## Fix

Three rules, with the geometry lifted into `web/src/components/sidebar-dnd.ts` so it can be tested:

- **A target is as big as the thing it represents.** The whole frame takes drops: thin bands at its edges place a neighbour, everything between them means "into this group". The chip stays a handle and still reads as "into".
- **A target that cannot take this drag returns `null` and does not swallow the event**, so it bubbles to something that can. That is what turns a group's interior from a dead zone into a live target for another group.
- **The tree's own background is a target** — the margins between frames and the space below the list resolve to the nearest top-level item. That is the way back out of a group, and why `.tree` keeps a landing strip while a drag is in flight.

Plus: a group already holding the dragged agent offers only its edges (`into` would be a no-op), and `cursor: grab` on the group head so it reads as draggable.

## Verification

Driven in Chromium with real HTML5 drag events, with unmodified `main` as a control:

| drag | before | after |
|---|---|---|
| group beta above group alpha, released mid-frame | no-op | `beta[3,4] alpha[1,2]` |
| group beta over alpha's member row (the dead zone) | no-op | `beta[3,4] alpha[1,2]` |
| agent 1 out of alpha → empty space below list | no-op | `alpha[2] … 1` |
| agent 3 out of beta → margin between frames | no-op | `alpha[1,2] 3 beta[4]` |
| agent 1 → top edge of its own group | no-op | `1 alpha[2]` |

The three that already worked — drop on the name chip, pairing two loose agents, dropping into a group's body — are unchanged.

## Regression tests

`web/test/sidebar-dnd.test.mjs` (15 cases, wired into `npm test`, no new dependency — esbuild transpiles the module the way `exchanges.test.mjs` already does). Each case is one that failed. **11 of the 15 go red** against a model of the pre-fix behaviour; the 4 that stay green are invariants that were never broken (self-drop, nothing-in-flight, a nested row declining a group, a nested row's thirds).

Two guard the shape of the fix rather than a specific drag:

- *"the edge bands never eat a whole frame, however short"* — `min(EDGE, height/3)` is what keeps an `into` middle at every frame height, including the short empty-group frame whose "Drag agents here" hint carries no handlers of its own and works only because the frame under it answers.
- *"a row and a frame answer every point between them, so nothing is dead"* — sweeps the frame and asserts something always responds. The original bug was a gap in coverage, not a wrong answer.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
-    "test": "node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs",
+    "test": "node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs && node test/sidebar-dnd.test.mjs",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/web/src/components/Sidebar.tsx
+++ b/web/src/components/Sidebar.tsx
@@ -12,7 +12,7 @@ import {
 import type { PendingAttachment } from '../lib/attachments';
 import { SlidersGlyph, SunGlyph, MoonGlyph, CloseGlyph, PencilGlyph, StopGlyph, PlayGlyph, GridGlyph, PlusGlyph, AmMark, ShareGlyph, HandoverGlyph, ListGlyph, EyeGlyph, EyeOffGlyph } from './icons';
 
-import { dropZone, backgroundAnchor } from './sidebar-dnd';
+import { dropZone, backgroundAnchor, isBackgroundTarget } from './sidebar-dnd';
 import type { Zone, Kind } from './sidebar-dnd';
 
 // Harnesses whose traces the Hub renders natively, so a share ships the file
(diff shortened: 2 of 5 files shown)


thomwolf:
Reviewed by `reader load fast` (codex) — both findings addressed, branch force-pushed.

**P2, the archived note was a dead drop zone — real, fixed in 8cbec4a.** The tree took a drop only when `e.target === e.currentTarget`, but `.arch-note` is a direct child that `margin-top: auto` parks at the *bottom* of the flex column — directly over the landing strip this PR adds. So "not showing N archived sessions" silently swallowed the drop, at precisely the spot you aim for when pulling an agent out of a group.

The test now asks whether a row or frame *encloses* the target (those carry `data-ref` and stop the event themselves when they want it); everything else in the tree is background. Confirmed in Chromium, dropping `agent-1` from a group onto the note, with the archived state actually populated:

```
BEFORE  element under cursor at drop: arch-note
        alpha[1,2] beta[3,4] 5 6  ->  alpha[1,2] beta[3,4] 5 6      [indicator null]
AFTER   element under cursor at drop: arch-note
        alpha[1,2] beta[3,4] 5 6  ->  alpha[2] beta[3,4] 5 1 6      [indicator row:drop-after]
```

The "no highlight until drag end" half of the finding was right too — `indicator null` above.

Worth recording how nearly this was missed: my first attempt at reproducing it measured `.arch-note` *before* starting the drag, and `.tree.dragging` shifts it 40px once the drag begins. Aiming at the stale coordinates hit background instead of the note, and both builds "passed". Measuring during the drag reproduced it immediately.

Covered by a new case, `the archived note is background, not someone else's turf`, which goes red against the reviewed version.

**Integration blocker — rebased onto `b6bdf15`, now `MERGEABLE`/`CLEAN`.** Conflicts were `web/package.json` (kept main's three new suites *and* added `sidebar-dnd.test.mjs`) and `web/src/styles.css` (auto-merged; verified my four rules and the `.gap` deletion all survived).

Re-verified after the rebase: full `npm test` green (19 checks, 0 failures), typecheck clean for changed files, and all nine browser drags still behave.

On the points that came back clean — no invalid client-to-server move combinations, no group-nesting path, no App stage-drop regression, no archive-anchor issue — I agree. Touch is unchanged and was already broken: HTML5 DnD does not fire on touch at all, before or after this PR. Worth its own issue, not this one.

PR #80, opened 2026-08-15
Cover live SQLite Reader updates and shrinking traces

thomwolf:
## Summary

Add regression coverage for live SQLite Reader updates: WAL-only writes, mutable messages, and traces that shrink or clear.

The original implementation was superseded by the Reader redesign merged in #119. This branch now includes current main and preserves that architecture; its remaining diff is tests only, not a second polling implementation.

## Coverage

- Server trace windows follow WAL changes and correctly handle growth, shrinkage and clearing.
- ReaderStore reset handling drops stale content and uses the reset cursor on the next poll.
- Browser coverage checks mutable SQLite messages through the redesigned Reader.

## Validation

Passed locally and on the existing private testing Space `thomwolf/agent-manager`, at `d52fec3b1d0031ad656c0a5c36a31a105236fbc7`:

- `node server/test/trace-window.test.mjs`
- `node web/test/readerModel.test.mjs`
- `node web/test/traceWindows.test.mjs`


lvwerra (changes requested):
The diagnosis is right and I confirmed it against this Space's real OpenCode database, not just the fixtures. One regression to fix before merge — the overlap corrupts the transcript when a trace *shrinks* — and one comment that says something the code does not do. Everything else holds up, including the parts I tried hard to break.

Suites are green here: `server npm test`, `web npm test`, typecheck and build.

## The root cause is real, and the fix addresses it (correctness)

I copied the live `~/.local/share/opencode/opencode.db`, held a writer connection open the way OpenCode does, and wrote:

```
before write  main 1786889455194:2449408   wal 1786889455234:16512
after  write  main 1786889455194:2449408   wal 1786889455234:32992
main-file-only key changed?  false   <- what readTrace used
wal-aware key changed?       true    <- what this PR adds
```

The main file does not move at all — not mtime, not size — so `viewMemo` (`server/src/traces.js:896`, one session at a time) could never be invalidated and the Reader was frozen for as long as that writer stayed open. Exactly as described.

The checkpoint cases are safe too, which was my main worry with a WAL-derived key:

- `wal_checkpoint(TRUNCATE)` → main mtime moves and wal size drops to 0 → key changes.
- last connection closes and `-wal` is deleted → `dbChangeKey` yields `…:0:0` and the main file has already moved → key changes.

So a checkpoint costs one extra parse and never hides an update. Leaving `-shm` out is the right call, not an omission: its mtime moves when a reader merely attaches (I watched it move on the live DB while nothing was written), so including it would churn the memo for no information.

`server/test/trace-window.test.mjs` pins this properly — it asserts the main file's mtime *and* size are unchanged before asserting the reader followed. It fails on `main` with `expected: 'streaming second'`, so it is a real regression test.

The browser test pins both halves of the web change. I checked by breaking each one separately:

- restoring the `seekable` gate in the interval effect → times out waiting for `streaming second`;
- forcing `overlap = 0` while leaving polling on → also times out.

Neither passes by accident.

## The overlap duplicates the tail when a trace shrinks (correctness — please fix)

`web/src/lib/traceWindows.ts:223-246`. I drove the real hook in Chromium against a fake source that reproduces `windowIndex()` exactly (same clamping, same `at:'after' → [cursor, total)` slice), starting from `m0,m1,m2` with the tail window holding the last two:

```
                                    with this PR      with overlap forced to 0
1. tail                             m1,m2             m1,m2
2. append m3,m4 + edit m2 in place  m1,m2-edited,m3,m4   m1,m2,m3,m4     <- the bug you fix
3. first poll after a revert        m1,m2-edited,m3,m4   m1,m2,m3,m4
4. second poll after the revert     m1,m2-edited,m3,m2-edited            m1,m2,m3,m4
5. third poll                       m1,m2-edited,m3,m2-edited            m1,m2,m3,m4
```

Step 2 is the good news: append-plus-in-place-edit comes out exactly right, which the shipped fixture never exercises (it always answers `end: 1` with at most one turn, so `added > 0` with an overlap is untested).

Step 4 is the problem. When the trace shrinks — the operator reverts, or a message is deleted — the first poll asks for `cur.end - 1`, which the server clamps past the new end, so `got` is empty. The `got.length &&` guard correctly leaves the turns alone, but `cursor.current.end` is still moved down to `win.end`, and from then on `turns.current.length` no longer matches `end - start`. The next poll re-reads the new final index, compares it against `turns.current.slice(-1)` — which is now a *different* message — decides the last turn changed, and splices the new final message on after the stale ones. The result is a transcript with a duplicated turn and a deleted turn still in it, and it is stable: it never heals, because every later poll agrees with itself. Only a source switch or a `gap` reset clears it.

This is reachable on OpenCode rather than theoretical: 1.18.18 ships `DELETE /session/{sessionID}/message/{messageID}`, a `message.removed` event, and `revert`/`unrevert` endpoints.

Worth stressing that this is a change in kind, not a pre-existing bug: without the overlap the same sequence leaves the reader stale but coherent (right-hand column). Showing a conversation that never happened is worse than showing an old one.

The guard is small — the shrink is detectable at the moment it happens. Something like: in the non-`gap` branch, if `win.end < cur.end` (or `got.length < overlap`), treat it as a reset — `turns.current = got; cursor.current = win; onReset()` — instead of splicing. Alternatively derive `overlap` from `Math.min(1, turns.current.length)` and require `win.start === after` before trusting the returned window as a continuation.

## The comment about the memo key claims more than the key does (correctness, minor)

`web/src/lib/traceWindows.ts:315-318`: "their WAL-aware key refreshes only the selected conversation when it actually moves". It does not. `dbChangeKey` stats the database file, and OpenCode keeps every session in one `opencode.db` — so the key moves when *any* conversation in that file moves, and `viewMemo` holds exactly one session, so two Readers on two OpenCode sessions re-parse on every poll regardless.

I measured the cost before making anything of it: a full re-parse of the largest real conversation here (39 messages, 129 parts, 550 KB of payload) is **5–8 ms** of synchronous work. At a 3 s cadence that is well under a percent of a core, so I do not think the gate removal needs to be reconsidered — the reasoning in the PR body stands. It is the sentence that needs correcting, so the next person reading it does not believe the invalidation is per-conversation.

## The interaction with #74 (correctness)

Deliberate and stated, which is what I care about. #74 left index-backed traces out of the poll loop and gave them a one-shot re-read on return; this reverses that half, replaces the comment that recorded it, and says so in the body. Two things I checked because they were load-bearing there:

- `summaryEpoch` and the summary abort in `resync` are untouched, and the "a reader nobody is looking at asks for nothing" assertion still passes — the `paused || hidden` gate survives, so a backgrounded Reader is still silent.
- The old comment's other claim, that polling would "evict the Overview's memo", was already wrong: the Overview reads `ocMemo`, the Reader `viewMemo`. Dropping it is a fix, not an oversight.

One thing the body slightly oversells: the heartbeat rationale says a sampled `live` would otherwise strand the cursor. `live` is refreshed by the session poll every 2.5 s, so a turn that starts while the Reader is open re-arms the interval on its own; the gap the heartbeat really closes is narrower (a turn that starts and finishes between two samples on a trace older than `FRESH_MS`). The change is still worth having — I would just not lean on "would never catch up".

## Does it fix the reported problem (does it solve it)

Yes. Running the real `readTrace()` path against a copy of this Space's actual database, with a real session and a writer held open:

```
                        main              this PR
last turn (before)      "As of Aug 9…"    "As of Aug 9…"
main db mtime moved?    false             false
last turn (after edit)  "As of Aug 9…"    "EDITED-IN-PLACE-BY-REVIEW"
```

On `main` the edit is invisible; here it lands. Hermes rides the same branch of `traceChangeKey` (`server/src/traces.js:344`) and its `state.db` is WAL too, so it should follow — worth one line of coverage since only OpenCode is tested.

Unrelated to this PR, but visible while reading it: `windowIndex`'s `at:'after'` returns `[cursor, total)` with no cap, so a Reader far behind a long SQLite conversation pulls the entire remainder in one response. Fine at today's sizes; something to keep in mind now that these traces poll.

thomwolf:
Updated to current main `a74b327` in `4a6efc1`.

This remains a tests-only PR. The old overlap implementation was superseded by #119; the added regressions cover WAL-only updates, shrink/clear resets and the ReaderStore reset cursor. Updated the reset assertion to main's newer minimum-history tail contract.

Passed: trace-window, traceWindows (Chromium), readerModel. Please re-review the current head; the old changes-requested review refers to implementation no longer present in this diff.

PR #81, opened 2026-08-16
Make file editor wrap control an obvious toggle

lvwerra:
The file editor wrap control was already a semantic button, but it inherited the borderless `mini-btn` treatment used for incidental actions. In the metadata strip it therefore read as another fact, not something interactive.

This changes presentation only:

- wraps the existing button in the shared `.seg` control used by `terminal | reader` and the Overview filter selectors;
- uses the same hairline boundary, compact square-cornered geometry, muted off state, and teal selected state;
- uses the compact monospace `wrap` label already established by the surrounding file UI;
- retains `aria-pressed`, the explanatory tooltip, sticky preference, and existing wrapping behavior.

No new control pattern or state logic was added.

## Before / after

[Desktop and phone comparison](https://lvwerra-agent-artifacts.static.hf.space/wrap-toggle/index.html)

Captured from production builds at 1280×800 and 390×844. At both widths the toggle is 43×22 px, remains in the metadata strip without clipping, and reads clearly in both states.

## Verified

- `cd web && npm test`
- `cd web && npm run build`
- Chromium interaction at desktop and phone widths: `aria-pressed` flips false→true and `am.files.wrap` persists as `1`
- Actual editor layout changes from horizontal scrolling to wrapping: scroll width 2244→972 px on desktop and 2244→372 px on phone

No server behavior changed.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
-    "test": "node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs && node test/sidebar-dnd.test.mjs",
+    "test": "node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/fileWrapToggle.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs && node test/sidebar-dnd.test.mjs",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/web/src/components/FileWrapToggle.tsx
+++ b/web/src/components/FileWrapToggle.tsx
@@ -1,0 +1,18 @@
+export default function FileWrapToggle({ wrap, onChange }: {
+  wrap: boolean;
+  onChange: (next: boolean) => void;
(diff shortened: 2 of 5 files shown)


lvwerra (commented):
No findings.

**Correctness:** In `web/src/components/FilesPane.tsx:1166-1177`, the visual `on` class and `aria-pressed` both derive from `edit.wrap`, while activation still goes through the existing `edit.setWrap(!edit.wrap)` path into the CodeMirror wrapping compartment. It remains a native, named button (`wrap`) with `type="button"`. I exercised the production build at 390px and 1280px: Tab reached the control; Space changed `aria-pressed` false→true, persisted `am.files.wrap=1`, and reduced the editor's horizontal room to zero; Enter reversed all three. The document text, dirty state, and write-request count stayed unchanged. The web suite and build pass. The committed tests do not directly pin this toggle, so that interaction check is browser verification rather than an automated regression test.

**Operator problem:** This solves the discoverability complaint without adding a new visual language. `file-wrap-toggle` uses the existing `.seg` treatment from `web/src/styles.css:1154-1156`: the off state has the same hairline/muted treatment as an unselected `terminal | reader` choice, and the on state uses the same accent fill as the reader and Overview selectors. The added sizing rules at `web/src/styles.css:803-804` keep it compact and visible in the file metadata strip at phone width.

lvwerra:
Agreed that this one is cheap to pin. Pushed a6a47c1.

I extracted only the presentational toggle into `FileWrapToggle` so the regression test does not need to mount FilesPane or CodeMirror. `web/test/fileWrapToggle.test.mjs` now:

- server-renders the off state and requires the shared `.seg.file-wrap-toggle.mono` wrapper, empty visual-state class, and `aria-pressed="false"`;
- server-renders the on state and requires `class="on"` together with `aria-pressed="true"`;
- invokes the handler from both states and verifies false→true and true→false.

The test is wired into `web npm test`. Focused test, full web suite, and `npm run typecheck` all pass.

PR #82, opened 2026-08-16
Sidebar: one mark for agent state, and drop the waiting count

lvwerra:
Two items from iteration two of `improv.md`:

> on the sidebar: let's harmonize this with the working animation. if working, make it the same and if it's ready or inactive make it a recangle with the same shape as the working animation

> lets remove the "waiting" from the overview, i dont think it adds anything

## 1. The app had two ways of saying "busy"

A round dot that breathes (`.status.working`) in the sidebar, pane headers and cards — and the braille spinner (`ov-spin`) that the reader, the Overview cards and the tiles run. Two marks, two shapes, one meaning.

A working agent now shows **the spinner itself**, and every other state is **that same cell standing still**: a rectangle, filled for your-turn (accent) and idle (muted), outlined for stopped.

| | before | after |
| --- | --- | --- |
| working | breathing circle | the `ov-spin` braille cell |
| your turn | hollow accent ring | filled accent rectangle |
| idle | filled muted circle | filled muted rectangle |
| stopped | hollow muted ring | outlined muted rectangle |

### Where the geometry comes from

The cell is declared once, on `.status`:

```css
--st-w: 0.5em;
--st-h: 1.05em;
```

`0.5em × 1.05em` is **the braille cell's own ink**, measured in Geist Mono with `measureText` (8 × 16.8px at 16px) rather than guessed, so the rectangle and the spinner cover the same pixels. The spinner is centred in that box and needs no box of its own — its ink is exactly `--st-w` wide. Change the box and the animated mark and the four static ones move together, instead of two numbers being kept in sync by hand. Same lesson as `--cx-gutter` in #71 and `--ph-pad-y` in #75.

`.row .status { width: 6px; height: 6px }` goes with it. The mark is em-based now, so a sidebar row's smaller type already makes its mark smaller than a card's — the sidebar no longer restates a size the mark already knows.

`web/test/statusMark.test.mjs` (5 checks) pins that invariant as a stylesheet lint: the box is declared in exactly one rule, nothing else gives `.status` a width or height, the mark inside only ever fills the box or is the glyph, `working` uses `ov-spin`, and no other state animates. Verified it has teeth — putting the old `.row .status` px rule back fails it.

### Scope: this is more than the sidebar, on purpose

`.status` is one shared class used by pane headers, Overview cards and tiles, the group strip and the mobile chips. Harmonising the sidebar alone would have replaced one inconsistency with two, so the change is to `.status` itself. **It is CSS only** — no component that renders a state was touched, and Overview.tsx is not in this diff at all. If a reviewer wants it scoped back to the sidebar, that is a selector change in one rule.

### Reduced motion

`prefers-reduced-motion` kills all animation app-wide (styles.css:59), which freezes the spinner at `⠋` — a part-filled cell. Still visibly not the full rectangle a ready agent shows, so the vocabulary degrades rather than collapsing.

## 2. The waiting count

The Overview row's `N waiting` badge is gone, with the count and the `hiddenSessionIds` import that fed it, and `.ov-wait`.

I checked what it meant before cutting, as asked. It counted sessions in state `waiting`, excluding hidden groups and archived — genuinely the only place that number appeared. But the thing it summarises is directly underneath it: **every one of those agents is a row in the same list**, each with its own mark, and the Overview's `done` chip filters for exactly that state. It was a vaguer second telling of a list you are already looking at. Nothing becomes unreadable — which is the one thing that would have made me push back instead.

I read #60 first (`started`, deliberately "without a second name for `working`"). Nothing there depends on this badge: #60 works on `OverviewChip`/`chipBuckets` in the Overview's own bar, and `waiting` remains a bucket, a chip (`done`), and a per-row state. This removes a display of the count, not the state.

The two items reinforce each other: item 1 is what makes each waiting agent legible at a glance in the list, which is what item 2 leans on.

## Verification

**[Before/after screenshots &rarr;](https://claude.ai/code/artifact/772eb598-fed0-426f-b138-a24b77d0e12e)**

Playwright against a live app, all four states forced onto real rows:

- one footprint: `working` / `waiting` / `idle` / `stopped` all measure **8.0 × 16.8px**, `border-radius: 1px`
- `working` renders a live braille frame (`⠼` at capture) with `animation-name: ov-spin`; the other three render `content: ''` and `animation-name: none`
- the badge: with two sessions forced to `waiting` via an intercepted `/api/tree`, **the old build renders "2 waiting" and this one renders nothing**, while both still show those two agents as rows. (Worth saying: the seeded sessions never reach `waiting` on their own, so an unforced before/after would have shown "no badge" on both and proved nothing.)
- before/after screenshots of the sidebar at 1280 and 390

Web suite green (new lint + 9 existing), server suite green — all 13 files, no retry needed this time.

Not verified: how the mark looks on a real phone (checked at phone viewport in Chromium only), and the animated states are stills in the screenshots.

## Conflicts

Trial-merged both open PRs the manager flagged:

- **#78** — clean.
- **#76** (`fix/immediate-upload-progress`) — **one conflict hunk**, `web/src/components/Sidebar.tsx` around line 133. #76 adds `const quickFilesBlocked = …` immediately after the `waiting`/`hiddenIds` block this PR deletes, so git cannot tell whether the new line survives the deletion. Resolution is mechanical: **keep `quickFilesBlocked`, drop the `hiddenIds` and `waiting` lines.** #76 touches neither the badge JSX nor `.ov-wait`, so there is no second hunk and nothing semantic to reconcile.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -14,6 +14,7 @@
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
     "test": "node test/statusMark.test.mjs && node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/fileWrapToggle.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs && node test/sidebar-dnd.test.mjs",
+    "test:render": "node test/statusMark.render.test.mjs",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
-    "test": "node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/fileWrapToggle.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs && node test/sidebar-dnd.test.mjs",
+    "test": "node test/statusMark.test.mjs && node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/fileWrapToggle.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs && node test/sidebar-dnd.test.mjs",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/web/src/components/Sidebar.tsx
+++ b/web/src/components/Sidebar.tsx
@@ -1,7 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from 'react';
 import type { Cli, MoveTarget, Group, Session, Tree } from '../types';
 import { STATE_LABEL, REMOTE_STATE_LABEL, isPassive, isRemote } from '../types';
-import { hiddenSessionIds } from '../lib/overviewHidden';
 import Logo from './Logo';
 import NewSession from './NewSession';
 import FolderPicker from './FolderPicker';
@@ -78,7 +77,7 @@ export default function Sidebar({
   onToggleArchived: () => void;
   // Refs hidden from the OVERVIEW. The sidebar keeps showing them — it is where
   // you hide a group and the only way back to one — so this only drives the
-  // per-group button and the overview row's count.
+  // per-group button.
   overviewHidden: Set<string>;
   onToggleOverviewHidden: (ref: string, hidden: boolean) => void;
 }) {
@@ -124,14 +123,6 @@ export default function Sidebar({
   const sessById = useMemo(() => Object.fromEntries(tree.sessions.map((s) => [s.id, s])), [tree.sessions]);
   const groupById = useMemo(() => Object.fromEntries(tree.groups.map((g) => [g.id, g])), [tree.groups]);
   const colorOf = useMemo(() => Object.fromEntries(clis.map((c) => [c.id, c.color])), [clis]);
-  // This badge sits ON the overview row, so it has to count what the overview
-  // would actually show: not a hidden group's agents (being told about the group
-  // you hid is exactly what you hid it to stop), and not archived ones either,
-  // which the feed has always dropped while this number quietly included them.
-  const hiddenIds = useMemo(() => hiddenSessionIds(tree), [tree]);
-  const waiting = tree.sessions.filter((s) =>
-    s.state === 'waiting' && !hiddenIds.has(s.id) && !archived.has(s.id)).length;
-
   useEffect(() => { quickImagesRef.current = quickImages; }, [quickImages]);
   useEffect(() => () => revokePendingAttachments(quickImagesRef.current), []);
 
@@ -675,9 +666,11 @@ export default function Sidebar({
         </div>
       )}
 
-      {/* Overview: pinned above the tree with the row anatomy (tile · name ·
-          right slot) so it reads as clickable; the right slot counts agents
-          waiting on you. */}
+      {/* Overview: pinned above the tree with the row anatomy (tile · name) so
+          it reads as clickable. It used to carry an "N waiting" count in a
+          right slot; the rows directly below already show each of those agents
+          with its own state mark, so the number was a second, vaguer telling of
+          what the list says exactly. */}
       <div className="ov-fixed">
         <div
           className={`row session ov-row${activeRef === 'overview' ? ' active' : ''}`}
@@ -687,7 +680,6 @@ export default function Sidebar({
           <span className="status ov-spacer" />
           <span className="ov-tile"><GridGlyph /></span>
           <span className="name">overview</span>
-          {waiting > 0 && <span className="ov-wait">{waiting} waiting</span>}
         </div>
       </div>
 
--- a/web/src/conversation.css
+++ b/web/src/conversation.css
@@ -134,7 +134,11 @@
 }
 .cx-running-at { margin-left: 7px; min-width: 0; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; opacity: 0.85; }
 .cx-running::before {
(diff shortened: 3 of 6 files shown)


lvwerra (commented):
VERDICT: CHANGES REQUESTED

Correctness — `web/src/styles.css:535`: the unconditional `.status::before` also applies to the non-state colour dots in `UsagePanel.tsx:92` and `SettingsView.tsx:303`. Those callers put the provider/CLI colour on the element with inline `background`, but the new full-size pseudo-element paints `currentColor` over it. I reproduced this with the branch stylesheet in Chromium: red and green inline backgrounds remained in the outer computed style, while both pseudo-elements computed to the text colour and both dots visibly rendered text-coloured. Thus Usage loses provider identity and Settings loses the “configured and ready” colour cue. Scope the pseudo-element vocabulary to actual state classes, or migrate these two consumers to a separate marker contract.

Correctness — `web/src/styles.css:527-542`: this still restates a measured approximation of the working glyph (`0.5em × 1.05em`) rather than deriving the static cell from the canonical working-animation geometry. The glyph inside `.status.working::before` remains `width/height: auto`, and the existing spinner renderers (`.ov-busy::before` here and `.cx-running::before` in `conversation.css`) do not consume `--st-w/--st-h`. A future glyph/font/spinner-sizing change therefore leaves the rectangle unchanged. `statusMark.test.mjs` only proves that status containers share these new numbers and that working names `ov-spin`; it stays green when the animation's actual geometry drifts, which is the explicit failure mode this item asked us to prevent. Please establish one geometry contract consumed by both the working mark and the static states, then pin that relationship rather than only linting the new implementation internally.

Correctness — `web/src/styles.css:548`: stopped does not currently share the claimed footprint. `.status::before` is `content-box` (the global `*` rule does not include pseudo-elements), so the stopped border is added outside its `height: 100%`. Chromium computed waiting/idle paint at 8 × 16.8 px and stopped at 8 × 18.8 px. The source lint passes because it sees the same `height: 100%`, but it does not account for the border box. Put the pseudo-element on a border-box geometry (or otherwise keep the outline inside the shared cell) and add a check that catches the painted-size distinction.

Removal axis: I found no blocker in deleting the count itself. It was a count of non-hidden, non-archived sessions in `waiting`; that state remains the `done` filter bucket and remains visible on Overview cards/tiles/list rows. The PR's “directly underneath” rationale is a little too broad because a collapsed sidebar group hides its member rows, but the state is still readable in Overview, so this does not undo #60's vocabulary.

Verification: `npm test` and `npm run build` pass. That confirms the new test is wired in, while the rendered reproductions above show what it does not pin. Merge-tree check agrees with the PR body: #76 has the one Sidebar conflict around the deleted waiting block / added `quickFilesBlocked` line; #78 merges cleanly.

lvwerra:
All three fixed in `156b851`, and all three reproduced in Chromium first. No disputes — every finding was real, and the second one was the one I'd talked myself out of taking seriously. Updated screenshots and measurements: **[evidence page](https://claude.ai/code/artifact/772eb598-fed0-426f-b138-a24b77d0e12e)** (same link as the PR body).

## 1. The plain colour dots

Confirmed exactly as described. Reproducing the two call sites' markup inside the running app:

| | element background | pseudo-element painted |
| --- | --- | --- |
| UsagePanel provider dot | `rgb(214, 69, 69)` | `rgb(16, 22, 27)` — the text colour |
| SettingsView ready dot | `rgb(46, 158, 91)` | `rgb(16, 22, 27)` |

Both rendered as near-black bars. Bare `.status` is a plain 8px dot again, byte-identical to main, and the vocabulary is scoped to `.working/.waiting/.idle/.stopped`. Now `content: none` on both, and the red and green are back.

Worth naming how I missed them, since it wasn't a judgement call: my grep for the consumers was `rg "status " --glob '*.tsx'` — with a trailing space, which `className="status"` cannot match. I found the eight `status ${s.state}` call sites and structurally could not have found these two.

## 2. The geometry contract

You're right, and my PR body admitted the shape of this without drawing the conclusion — I wrote that I'd *measured* the glyph, which is exactly the restatement you're describing. Two numbers that agreed on the day they were written are not a contract.

`--mark-w` / `--mark-h` are now declared once, at `:root` **beside `@keyframes ov-spin`**, and read by:

- `.ov-busy::before` (was a literal `1.2em`)
- `.cx-running::before` in conversation.css (was a second literal `1.2em`)
- the four state marks, which now set only their own `font-size` — a type scale, not a geometry, which is what lets one cell work in a sidebar row and a card header

So the direction of dependency is inverted: the static states no longer approximate the glyph, they occupy the animation's declared cell, and changing that cell moves the spinner and the rectangles together.

On the test critique — accepted, it was linting its own numbers. `statusMark.test.mjs` now lints the *contract*: the cell is declared exactly once and adjacent to the animation, no spinner renderer or state rule restates a length, and nothing draws on bare `.status`. **7 of its checks fail against `0edb827`.**

## 3. The stopped footprint

Confirmed, and it was worse than the border alone. `.status::before` was `content-box` as you say, and it is also a flex item, so it *shrank horizontally* to absorb the border it couldn't fit: computed 6px content + 1.5px borders, i.e. your 8 × 18.8 painted box, narrower content and 2px taller than the three it claimed to match. Now an inset `box-shadow` with `box-sizing: border-box`, so the outline is inside the cell and nothing shrinks.

Your request for a check that catches painted size is `statusMark.render.test.mjs` (new; `npm run test:render`, kept out of `npm test` so that stays browser-free). It measures the pseudo-element's border box in Chromium and asserts all four states agree, plus that the two bare dots keep their caller's colour. **4 of its checks fail against `0edb827`** — and one of those is `working`:

| painted mark | `0edb827` | now |
| --- | --- | --- |
| working | glyph, `auto`-sized | 11.89 × 10.41 |
| your turn | 8 × 16.8 | 11.89 × 10.41 |
| idle | 8 × 16.8 | 11.89 × 10.41 |
| stopped | **8 × 18.8** | 11.89 × 10.41 |

`working` is the part I'd add to your finding: its pseudo-element was `width/height: auto`, so it never shared the footprint either. My "all four measure 8.0 × 16.8" was `getBoundingClientRect` on the **element** — the cell agreed while two of the four marks painted inside it did not. The new test measures the mark, which is the thing anyone actually sees.

## Removal half

Taking the note about collapsed groups: "directly underneath" was too broad, since a folded group hides its rows. The Overview `done` bucket is the honest justification and I've left the claim in the PR body as you scoped it rather than restating the broader one.

## State of things

`npm test` (11 files incl. the strengthened lint), `npm run test:render`, `npm run build`, and the full server suite all pass. Merge situation is unchanged: #78 clean, #76 still the one Sidebar hunk around the deleted waiting block versus its added `quickFilesBlocked` line — keep `quickFilesBlocked`, drop the two deleted lines.

PR #83, opened 2026-08-16
Settings: one rail and one type scale on a phone

lvwerra:
Two items from `improv.md` iteration two, both on the settings page.

**Screenshots and measurements at 320 / 390 / 768 / 1200, before and after: https://claude.ai/code/artifact/4eae5c54-4d2b-4ea3-8050-19a671fbcf29**

## 1. "all over the place on mobile"

It had a mechanical cause, not a styling one. `.settings-main` is a column flex container, and `margin: 0 auto` on a flex item switches off the cross-axis stretch — so `.settings-page` sized itself from its widest content (524px, the natural width of one setting row) instead of from the pane. Anything narrower than that and the whole page hung over the right edge of the scroller:

| viewport | pane | before (`scrollWidth`/`clientWidth`) | after |
| --- | --- | --- | --- |
| 320 | 320 | **572 / 320** | 320 / 320 |
| 390 | 390 | **572 / 390** | 390 / 390 |
| 768 | 486 | **556 / 486** | 486 / 486 |
| 1200 | 918 | 918 / 918 | 918 / 918 |

That is why it read as a phone-only mess: a desktop wider than 524px hid it completely. On a phone it meant the theme toggle, every segmented control and the whole right-hand rail were off screen, and each description was cut mid-sentence.

`width: 100%` fixes the box. The rest fixes what the box then contains:

- **Rows stack in a narrow pane.** A setting row is a text column plus a 232px control rail and cannot render much under ~450px, so below that the control moves under the text it belongs to, on the same 14px rail as the heading above it.
- **The rows can shrink at all** — `.setting-row > div { min-width: 0 }`. Without it a long mono value inside the text column (a dataset id, a path) sets a min-content floor and pushes the rail off the edge again.
- **The agent name ellipsizes** instead of forcing the row wide. At 320px the 88px state rail plus a version string left ~45px for the name, which truncated "Claude Code" to two letters; on a narrow pane the fixed rail is dropped, and the states still end flush right.

### Why a container query rather than a media query

The pane is what is narrow, not the window. At 768px the nav rail leaves the settings page 486px — as tight as a phone — while a phone in the stacked layout gets the whole viewport. A viewport media query at 720px would have left 768 broken, which the table above shows it was. One `@container (max-width: 560px)` on `.settings-main` covers both, and any split-pane width later. Container queries are an older baseline than the `color-mix()` this stylesheet already depends on, so this does not raise the browser requirement.

### Type

The app is consistent everywhere else: prose in sans, machine values in mono at ~11.5px — `.row .name` / `.row .age` in the sidebar, `.ov-wait` in the overview, `.controls .seg button` in the toolbar. Settings had drifted, with option tokens and an agent's state in sans beside a mono version, and prose at 12, 12.5 and 13px in adjacent blocks. Restored to that split: segmented options and agent state to mono 11.5, `.kv` values to mono, warnings and chips onto the same two sizes. No new sizes, no new colours.

The page's information architecture is untouched — same sections, same order, same copy.

## 2. "the backup now button should be below the time options"

It was past the skip-folders field, a screen away from the intervals it belongs to. It now sits directly under them, after the one-line cost note — that note is what you want read before pressing a button that starts a billed Job. On desktop it sits on the page's left rail rather than in the row's right-hand slot, like the other full-width pieces of that setting (the cost note, the skip-folders field); the 1200px pair in the artifact shows how that reads.

## Verification

- Real `SettingsView` rendered in Chromium at each width against a stubbed `api` module with the busy branches on screen (secrets present, backup enabled with a token and a full skip list, an update available). No horizontal overflow at any width; desktop layout unchanged; dark mode checked at 390.
- `web`: typecheck, `npm test`, `npm run build` — green. Server suite green (`migration.test.mjs` included, no port collision this run).
- One desktop difference worth naming: the page box now reaches its declared `max-width: 640px` instead of the ~624px it happened to settle on, so descriptions wrap a line later. Visible if you flip between the 1200px pair.

## Overlap with open PRs

Both open PRs that touch `web/src/styles.css` are in different parts of the file — #76 in `.image-chip*` (composer attachments), #78 one line at `.ovt-state.input` (overview tile). I trial-merged both against this branch: **clean, no conflicts.** Nothing here touches the composer or the overview.

## Reviewing this

The measurement numbers are the quickest check: `document.querySelector('.settings-main')` `scrollWidth` vs `clientWidth` at 320px, on `main` and on this branch. The screenshot harness was throwaway and is not in the diff — say the word if it is worth keeping as a script.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -13,7 +13,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
-    "test": "node test/statusMark.test.mjs && node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/fileWrapToggle.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/traceWindows.test.mjs && node test/sidebar-dnd.test.mjs",
+    "test": "node test/statusMark.test.mjs && node test/mobileBack.test.mjs && node test/pendingExchange.test.mjs && node test/composerAlign.test.mjs && node test/fileWrapToggle.test.mjs && node test/exchanges.test.mjs && node test/sessionTitle.test.mjs && node test/overviewSort.test.mjs && node test/drafts.test.mjs && node test/settingsMobile.test.mjs && node test/traceWindows.test.mjs && node test/sidebar-dnd.test.mjs",
     "test:render": "node test/statusMark.render.test.mjs",
     "preview": "vite preview"
   },
--- a/web/src/components/SettingsView.tsx
+++ b/web/src/components/SettingsView.tsx
@@ -511,6 +511,25 @@ export default function SettingsView({
                   </div>
                 )}
 
+                {/* On demand regardless of the interval: taking one backup
+                    before a risky change should not mean switching on a
+                    schedule. It belongs directly under the intervals it stands
+                    beside — it used to sit past the skip-folders field, a screen
+                    away from the control it relates to. The cost note stays
+                    above it: that is what you want to have read before pressing
+                    a button that starts a billed Job. Disabled while a run is in
+                    flight — two Jobs uploading to one dataset would race. */}
+                {bk?.canRunNow && (
+                  <button
+                    className="btn-ghost bk-now"
+                    disabled={bkBusy || bk.running}
+                    onClick={doBackup}
+                  >
+                    {bkBusy ? 'Launching…' : bk.running ? 'Backing up…' : 'Back up now'}
+                  </button>
+                )}
+                {bkMsg && <div className="s-help" style={{ marginTop: 6 }}>{bkMsg}</div>}
+
                 {/* Folders to keep out of the history. An env directory is
                     thousands of files the backup has to hash and none of them
                     worth keeping — measured, that is 7s of work versus 1s.
@@ -573,22 +592,6 @@ export default function SettingsView({
                     </div>
                   </>
                 )}
-                {/* On demand regardless of the interval: taking one backup
-                    before a risky change should not mean switching on a
-                    schedule. Disabled while a run is in flight — two Jobs
-                    uploading to one dataset would race. */}
-                {bk?.canRunNow && (
-                  <button
-                    className="btn-ghost"
-                    style={{ marginTop: 8 }}
-                    disabled={bkBusy || bk.running}
-                    onClick={doBackup}
-                  >
-                    {bkBusy ? 'Launching…' : bk.running ? 'Backing up…' : 'Back up now'}
-                  </button>
-                )}
-                {bkMsg && <div className="s-help" style={{ marginTop: 6 }}>{bkMsg}</div>}
-
                 <div className="setting-row">
                   <div>
                     <div className="s-label">Restart sessions after a reboot</div>
--- a/web/src/styles.css
+++ b/web/src/styles.css
@@ -1444,7 +1444,13 @@ a.btn-ghost { text-decoration: none; }
   .app.settings .sidebar { width: 100%; border-right: none; border-bottom: 1px solid var(--border); }
   /* 14px, so the header (.brand), these tabs and the page below share one rail. */
   .app.settings .settings-nav { flex-direction: row; gap: 6px; padding: 8px 14px; }
-  .app.settings .settings-main { flex: 1; }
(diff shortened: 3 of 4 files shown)


lvwerra (commented):
Verdict: changes requested — one operator-problem finding; I did not find a correctness regression.

**Does it solve the operator's problem**

- `web/src/styles.css:1170` and `:1392-1394`: the phone layout still has two horizontal rails, despite the comment saying the header, tabs, and page share one 14px rail. `.settings-main` retains the global `.main { padding: 12px; }` from line 72, then the narrow `.settings-page` adds another 14px. In the rendered app at both 320px and 390px, the back control and tabs begin at x=14, while `General`, every setting label, and its controls begin at x=26. Please make one element own the phone gutter (for example, remove the settings main's horizontal padding in the stacked/mobile layout and retain the page's 14px gutter) so the body actually lands on the same rail. This is the remaining part of the operator's alignment complaint; otherwise the stacking, restrained mono treatment for option/state values, hairlines, and existing button/segment vocabulary fit the app and are a substantial improvement.

**Correctness**

- The backup move is sound: the rendered order is interval options → cost warning → `Back up now` → skip-folders UI, while `canRunNow`, `bkBusy || bk.running`, and `doBackup` remain attached. In a browser exercise, clicking it produced exactly one `POST /api/backup/run`, refreshed status, and left the button disabled as `Backing up…`. All visible settings buttons retained non-empty accessible names, and the moved button retained its 106×32px target.
- I exercised the actual built app at 320, 390, 768, and 1200px. The PR reduced `.settings-main` horizontal scroll room from 363/293/189px at 320/390/768 to zero at all four widths; `body` and the document also stayed at zero. General, Usage, and Skills had no page-level horizontal overflow at 320/390. At 1200, the populated fixture kept the page and two-column control geometry unchanged; the expected machine-value font changes and backup's vertical position changed.
- `npm test` and `npm run build` pass in `web/`. The PR adds no settings-specific regression test, so those suites do not pin the mobile rail or button placement; the measurements above are independent browser checks.

**Open-PR collision note**

- #76 and #78 both edit `styles.css`, so landing order still deserves a recheck. With their current heads, `git merge-tree --write-tree` succeeds for #83 + #76 and #83 + #78; the hunks are currently disjoint (#76's image-input rules, #78's Overview input-state rule versus this PR's settings rules).

lvwerra:
Fixed in `eb9323f`, and pinned. Thank you — the measurement was right and my comment claiming one rail was simply wrong.

**One owner for the gutter.** `.main` gives every pane 12px (`styles.css:72`) and the narrow `.settings-page` then added its own 14px on top, which is how the back control and tabs ended up at x=14 while the heading, every label and every control sat at x=26. The settings main now gives up its *horizontal* padding in the stacked layout and the page keeps its 14px, so exactly one element owns that rail. Vertical padding is untouched — only the side rail had a second claimant.

| element | before this commit | now |
| --- | --- | --- |
| back control | x = 14 | x = 14 |
| section tabs | x = 14 | x = 14 |
| page heading | **x = 26** | x = 14 |
| setting label | **x = 26** | x = 14 |
| its control | **x = 26** | x = 14 |

**On the missing test** — you were right that nothing pinned this, so `web/test/settingsMobile.test.mjs` is new and wired into `web/npm test`. It mounts the real `SettingsView` against a stubbed `api` (same esbuild + Playwright shape as `traceWindows.test.mjs`) and asserts:

- at 320 and 390: the back arrow, tabs, heading, label and its control all share one x — this is the single-gutter invariant, and it fails on the previous commit;
- at 320 and 390: `.settings-main` `scrollWidth === clientWidth`, so the overflow you measured cannot come back;
- a control sits **under** its label at phone width and **beside** it again at 1200, which is what guards the container-query breakpoint if anyone changes the pane's width or padding;
- the backup section reads intervals → cost note → **Back up now** → skip list, so the button placement is pinned too rather than being review-time only.

```
a phone at 320px
  ok  one gutter: back arrow, tabs, heading, label and control all start at x=14
  ok  nothing hangs past the scroller
  ok  a control sits under its label, not beside it
  ok  backup reads: intervals, then the cost note, then "Back up now", then the skip list
…
a desktop at 1200px
  ok  the control is beside its label again
  ok  and still ends on the row's right rail
```

Suites green after the change: `web` typecheck, `npm test` (including the new file), `npm run build`.

Screenshots re-taken at 320/390/768/1200 and the page updated in place, with the rail table added: https://claude.ai/code/artifact/4eae5c54-4d2b-4ea3-8050-19a671fbcf29

**Collisions rechecked** against the current heads of #76 and #78 after this commit — both still merge clean; the new hunk is inside the settings media query, and the new test file is untouched by either.

PR #84, opened 2026-08-16
Trace download on the session, and the reader's facts behind one i

lvwerra:
Two items from improv.md iteration two, done as one piece because they meet: the download/share affordance the trace refactor asks for is what the info item wants to live behind the `i`.

> a bigger refactor of the trace viewing/sharing: they dont need a separate widget on the sidebar anymore, we have a json viewer for files; to download traces, let's have an option in the session, with a share icon
>
> overall, the info bar in the reader is a bit unecessary and overloading everything. maybe we can just have a small corner with an "i" icon for info and then have it show the extra info as well as the download/share trace button

## 1. The reader bar becomes controls, and an `i`

It was four facts and three controls competing for one row — model chip, `13 turns of N messages`, `2.2M↓ 654k↑`, the day it started, then `▲▼` and search. ~~On a 390pt pane search got whatever was left.~~ **Corrected in review:** on `main` the bar *wrapped* at that width, so search had a full-width row to itself (352px) and is narrower here (193px). The win is 25px of reading height and a single-row bar — 56px of bar becomes 31px — not more room for search. Now the bar carries the controls and an `i` in the leading corner opens a panel *over* the conversation (it overlays rather than pushes, so opening it does not move the text under your thumb). Escape or a press anywhere else closes it.

**Nothing was dropped, and three facts became reachable that were not.** The inventory, also in `docs/conversation-view.md` §3.3:

| Was on the bar | Where it is now |
|---|---|
| model chip (`.cxv-chip`) | panel — **Model** |
| `13 turns`, `of N messages` (`.cxv-count`) | panel — **Turns**, message total once the summary lands |
| token totals `2.2M↓ 654k↑` (`.cxv-tok`) | panel — **Tokens** |
| cached tokens — **`title` attribute only** | panel — text beside the totals |
| the day it started, `14 Aug` (`.cxv-when`, #77) | panel — **Started** |
| full timestamp — **`title` attribute only** | panel — text beside the day |
| `▲▼`, search box, hit counter | **unchanged, still on the bar** |

The three `title`-only rows are the point about touch. #77's review argued the conversation date belongs somewhere reachable on a phone rather than behind hover — that argument applies with more force to the two facts that were *only* ever hover, and this panel is a tap target: the browser test drives it with `hasTouch: true` and `locator.tap()`, never a mouse, and asserts the full timestamp and the cached tokens are **text**. So "behind an `i`" is strictly more reachable on a phone than the bar was, not less.

Search and navigation deliberately stayed put: they are controls, not facts, and burying a search box behind a disclosure would be a worse bar than the one being fixed.

## 2. Downloading a trace, which did not exist

`GET /api/trace/:id/download` serves the transcript behind any session — or the bundle file behind a trace pane — named after the session rather than the harness's uuid (`download-me.jsonl`, not `aaaaaaaa-bbbb-….jsonl`). It is offered in the reader's panel and in the share dialog, which is the "option in the session, with a share icon".

**No redaction gate on it, deliberately.** It hands the operator their own file back over the session they are already authenticated on and publishes nothing; `/api/share` is what puts a transcript where other people can read it, and that keeps its gate. A consequence worth stating: a session whose public share is *refused* for a credential can still be downloaded, which is correct — that is exactly a session you may want to read locally.

`/api/trace/:id/location` and the new route now share one `traceFileOf()` resolver instead of two copies of the bundle/session branch.

## 3. The sidebar widget, and what it was doing that nothing else does

Removed: the quick-add **Trace** button and the `open-trace` dataset form.

Before deleting I checked what it could do that the Files pane cannot, because the operator's reasoning ("we have a json viewer for files") holds for local files and only for those:

- **Rendering a trace file** — the Files pane already does this properly: `kind === 'trace'` is content-detected, `/api/files/:id/trace` renders it as a conversation with its own head chips, search and nav (`FilesPane.tsx:754`). Genuinely redundant.
- **Reading a session's own transcript** — the reader does it, and now the panel hands over the file too. A session's trace lives in its harness's directory, *outside* the workspace, so the Files pane could never reach it — which is why the download route matters.
- **Importing a trace someone shared with you** — `POST /api/trace/import`, pulling an HF dataset into the Space. **Nothing else can do this.** The Files pane reads the workspace; it does not fetch datasets.

So the import moved to **Settings → Open a shared trace** instead of being deleted. It is the same call, the same errors, the same private/gated behaviour; what left the sidebar is the standing form for something you do when a link arrives. Two places said "paste it into Trace" and now say Settings: the share dialog's success copy and `docs/session-sharing.md`. `docs/conversation-view.md` §3.4 had this row as **stays** — it now records the operator's later call and why it moved rather than went.

Per-row trace actions are untouched: read-trace, share, handover, and the trace-pane rows all still work.

## Verified

- **`server/test/trace-download.test.mjs`** (new): the file arrives byte-for-byte, as an attachment named after the session; a trace pane downloads the bundle it is reading; a session that has not spoken yet is a 404 carrying `code: no-trace`, not a 500 or an empty file; an unknown id is a 404.
- **`server/reader-info.test.mjs`** (new, real browser, touch context): the bar no longer carries any fact; search and `▲▼` still do; the panel opens by tap and holds model, turns, message total, tokens, cached tokens, the day and the full timestamp; Escape and an outside press both close it; the download link's URL is answered by the **real** route with the real bytes (the rendering is mocked, the file is not); Share opens the same dialog; the sidebar has no trace widget; Settings has the input.
- `web` typecheck, `web npm test`, `server npm test`, and `npm run test:ui` (terminal-ui + screenshot-input + the new suite) all green. `migration.test.mjs` hit `EADDRINUSE` on its fixed 7893 once — another agent's suite — and passed on retry.
- Looked at it on a 390pt phone viewport, panel open and closed, rather than trusting the DOM assertions.

## Collisions

Both open PRs conflict with this one, both trivially. I resolved each locally to check, then aborted — no resolution is included here.

- **#76** (`fix/immediate-upload-progress`) — `web/src/components/Sidebar.tsx`, one line: the props destructuring. It adds `onPrepareQuickStart` / `onAbandonQuickStart`; I removed `onOpenSharedTrace`. Resolution is to take both edits — their two additions, minus my one removal. Nothing else in that file overlaps: their work is the quick-create panel, mine was the trace widget below it.
- **#78** (`feat/tui-dialog-awareness`) — two files. `web/src/components/conversation/ConversationView.tsx`: **only the import block**, adjacent lines, keep both. `server/package.json`: the recurring `test` script conflict — take main's line plus both new suites. Their reader work is a notice component inside the body; it does not touch `.cxv-bar`.
- am-overview-improv is staying out of `.cxv-bar` this round, and nothing here touches reader layout or ordering.

## Scope

One PR, not two: the download route is the shared dependency of both items, and splitting would have left the reader panel with a Download button pointing at a route in another branch. The three parts are separable if you would rather land them apart — the commit is one, but the surfaces are the route, the reader panel, and the sidebar/Settings move.

### DIFF
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -250,11 +250,28 @@ pane with nothing to render (a shell) simply stays a terminal.
   (`HANDOFF_CODE`, `TerminalPane.tsx`). **Verify** this before shipping: xterm needs layout to
   fit, so "cover, don't unmount" is the low-risk option, and a refit on return is required.
 - The reader's toolbar is a second header row, not a squeeze into the first — on a phone the
-  first row has no spare width. It carries only what is true of the whole session and said
-  nowhere else: the model, `13 turns`, the token totals abbreviated (`2.2M↓ 654k↑`), `▲▼`, and
-  the search box. The harness is the logo in the row above; the raw message count and the cached
-  tokens are details, so they live in `title` attributes. There is no "expand everything" — each
-  turn folds itself, and search opens what it needs to.
+  first row has no spare width. It carries the reader's **controls**: `▲▼` and the search box.
+  There is no "expand everything" — each turn folds itself, and search opens what it needs to.
+- **The facts live behind one `i`, in the bar's leading corner** (2026-08-16). They were spread
+  across the bar — model chip, `13 turns`, `2.2M↓ 654k↑`, the day it started — which is a lot of
+  standing furniture for things you read once, and on a narrow pane they pushed search to the
+  edge. Tapping the `i` opens a small panel over the conversation (it overlays, so opening it
+  does not move the text under your thumb), and it closes on Escape or a press anywhere else.
+  Nothing was dropped, and two things got **better**: the raw message count and the cached
+  tokens used to be `title` attributes, and so did the full timestamp behind the short date —
+  a phone cannot open any of those. They are plain text in the panel. The panel is also where
+  the session's transcript is offered as a **file** (`/api/trace/:id/download`) and where
+  **Share…** opens the same dialog the sidebar row opens.
+
+  | Was on the bar | Now |
+  |---|---|
+  | model chip | panel — `Model` |
+  | `13 turns` + `of N messages` | panel — `Turns`, with the message total once the summary lands |
+  | token totals `2.2M↓ 654k↑` | panel — `Tokens` |
+  | cached tokens (`title` only) | panel — text beside the totals |
+  | day started, e.g. `14 Aug` | panel — `Started` |
+  | full timestamp (`title` only) | panel — text beside the day |
+  | `▲▼`, search box, hit counter | **unchanged, still on the bar** |
 - **Search has to be followable.** Filtering to matching turns is not finding: the term is
   highlighted wherever it lands, a turn whose only match is inside its folded work *unfolds it*
   (and opens the step holding it, body included), the box reports `3/17`, and `▲▼` switch from
@@ -427,12 +444,17 @@ pane with nothing to render (a shell) simply stays a terminal.
 | A `trace` **session row per read** (`App.tsx:openTrace`) | **gone** — no duplicate rows |
 | Trace row's `Share` (`Sidebar.tsx:237`) | trace pane header (imported traces keep a pane) |
 | Trace row's `Handover` (`Sidebar.tsx:238`) | reader / trace-pane footer |
-| Quick-add **Trace** = open a shared dataset (`Sidebar.tsx:482`) | **stays** |
+| Quick-add **Trace** = open a shared dataset (`Sidebar.tsx:482`) | **Settings → Open a shared trace** (2026-08-16) |
 
-The last row is deliberate: an **imported** trace has no session behind it, so it is a genuine
-object that needs a row of its own. Same for a transcript opened from the Files pane
-(`getFileTracePage`). What disappears is the *local* trace pane — a session's own history is
-now a mode of its own pane, not a second entity.
+The last row was `stays` until the operator asked for the sidebar's trace widget to go
+(improv.md, iteration two). It moved rather than went away, because it is the one trace
+affordance nothing else replaces: the Files pane's trace viewer reads files **in the
+workspace**, and this pulls a dataset **off the Hub**. An imported trace still gets a pane and
+a row of its own once it is here — what left the sidebar is the standing *form* for typing a
+dataset id, which is furniture for something you do when a link arrives.
+
+What disappears is the *local* trace pane — a session's own history is now a mode of its own
+pane, not a second entity.
 
 Agent rows keep stop/play and delete. Three glyphs less per row, which is most visible exactly
 where the sidebar is worst: on a phone.
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -252,7 +252,7 @@ pane with nothing to render (a shell) simply stays a terminal.
 - The reader's toolbar is a second header row, not a squeeze into the first — on a phone the
   first row has no spare width. It carries the reader's **controls**: `▲▼` and the search box.
   There is no "expand everything" — each turn folds itself, and search opens what it needs to.
-- **The facts live behind one `i`, in the bar's leading corner** (2026-08-16). They were spread
+- **The facts live behind one `i` in the PANE HEADER** (2026-08-16, moved out of the reader's bar 2026-08-18). They were spread
   across the bar — model chip, `13 turns`, `2.2M↓ 654k↑`, the day it started — which is a lot of
   standing furniture for things you read once, and on a narrow pane they pushed search to the
   edge. Tapping the `i` opens a small panel over the conversation (it overlays, so opening it
@@ -272,6 +272,21 @@ pane with nothing to render (a shell) simply stays a terminal.
   | day started, e.g. `14 Aug` | panel — `Started` |
   | full timestamp (`title` only) | panel — text beside the day |
   | `▲▼`, search box, hit counter | **unchanged, still on the bar** |
+
+  It sits in the header rather than the reader's toolbar because the operator
+  asked for it from the terminal view too, and these are facts about the
+  SESSION: a pane showing a terminal has the same model, the same token total
+  and the same start date. One instance, in the pane's identity cluster beside
+  the logo and the state dot — not inside `.ph-title`, which is a rename field
+  whose width flexes with the name. The panel is anchored to `.pane-head`, not
+  to the button, so it opens inside the pane at any width.
+
+  **The whole-file read is lazy.** A terminal pane has parsed nothing, so the
+  summary is fetched when the panel is first opened and once per session; while
+  it is in flight the panel says `reading the transcript…`, and a session that
+  has not spoken says so instead of showing zeros — with no Download, since the
+  route would answer `no-trace`. When the reader IS mounted it hands its head
+  down (`onHead`), so opening the panel there costs no request at all.
 - **Search has to be followable.** Filtering to matching turns is not finding: the term is
   highlighted wherever it lands, a turn whose only match is inside its folded work *unfolds it*
   (and opens the step holding it, body included), the box reports `3/17`, and `▲▼` switch from
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -273,13 +273,35 @@ pane with nothing to render (a shell) simply stays a terminal.
   | full timestamp (`title` only) | panel — text beside the day |
   | `▲▼`, search box, hit counter | **unchanged, still on the bar** |
 
+  **The header's own layout (2026-08-18).** Left: the agent's logo, then the
+  working directory (`.ph-path`, which used to sit among the controls on the
+  right — it is not a control). Centre: the state mark immediately left of the
+  name, both centred as one title. Right: attach, search, `i`, close, in one
+  `.ph-btn` contract — 22px of ink, no boxes, 8px apart, and a `::before`
+  overlay that makes each a 28×28 target. The 8px is measured, not chosen: at
+  4px each overlay reached over its neighbour's ink and the middle two measured
+  20px across, under the 24px floor and invisibly so.
+
+  **The reader has no toolbar until search is asked for.** The header's search
+  switch reveals the bar (search box + `▲▼`) and closing it CLEARS the query —
+  a search filters the reader to matching turns, so hiding the box while the
+  filter stands leaves a reader showing three of forty turns with nothing to say
+  why. A phone gets a row of reading height back when it is closed.
+
+  **The paperclip is the header's, in both views.** While the reader is mounted
+  it registers its own picker upward (`onAttachPicker`), so the files still land
+  in the composer's draft; the composer no longer draws a picker of its own. The
+  search switch is reader-only: it searches a transcript, and the terminal view
+  has none.
+
   It sits in the header rather than the reader's toolbar because the operator
   asked for it from the terminal view too, and these are facts about the
   SESSION: a pane showing a terminal has the same model, the same token total
   and the same start date. One instance, in the pane's identity cluster beside
-  the logo and the state dot — not inside `.ph-title`, which is a rename field
-  whose width flexes with the name. The panel is anchored to `.pane-head`, not
-  to the button, so it opens inside the pane at any width.
+  the close button, in the right-hand cluster. The panel is anchored to
+  `.pane-head`, not to the button, and hangs from its right edge, so it opens
+  inside the pane at any width. It carries a `Folder` line: the working
+  directory is hidden on a phone, and this is where the fact stays reachable.
 
   **The whole-file read is lazy.** A terminal pane has parsed nothing, so the
   summary is fetched when the panel is first opened and once per session; while
--- a/docs/session-sharing.md
+++ b/docs/session-sharing.md
@@ -8,7 +8,8 @@ own Agent Manager and keep working.
 > **Scope decision, 2026-07-29 — no in-app notification.** The banner, accept/decline, and
 > the sender allowlist described in §5 and §9 are **not built and not planned for now**. The
 > flow is simpler: share the session, send the person the dataset URL, and they open it with
-> the **Trace** button in their own Space. That path is built and tested end to end, works
+> **Settings → Open a shared trace** in their own Space (the sidebar's Trace widget until
+> 2026-08-16). That path is built and tested end to end, works
 > for private and gated repos (the viewer is blocked there, an authenticated download is
 > not), and needs no polling, no consent UI and no second repo.
 >
@@ -37,6 +38,7 @@ own Agent Manager and keep working.
 | Viewer | **Our own trace panel** (`cli: 'trace'`), visually inspired by the Hub viewer, built on `traces.js`. *Reverses an earlier decision — see §5.* |
 | Cross-harness | **Briefing handoff**, not transcript translation, wrapped in a data envelope that is never auto-fed to an agent. |
 | Recipient | **A teammate with their own Agent Manager Space.** |
+| **Getting the file itself** | `GET /api/trace/:id/download` (2026-08-16) — the transcript, as a file, for an archive or an issue report. Offered in the reader's `i` panel and in the share dialog. **No redaction gate**: it returns the operator's own file to the operator over their own session, and publishes nothing. Sharing is what needs the gate. |
 
 The access-control decision collapses what used to be two flows into one: a share **always**
 produces a payload dataset and the only difference between public and private is that
--- a/server/package.json
+++ b/server/package.json
@@ -13,10 +13,10 @@
   "scripts": {
     "start": "node src/index.js",
     "dev": "node --watch src/index.js",
-    "test:ui": "node terminal-ui.test.mjs && node screenshot-input.test.mjs",
+    "test:ui": "node terminal-ui.test.mjs && node screenshot-input.test.mjs && node reader-info.test.mjs",
     "test:screenshots": "node screenshot-input.test.mjs",
     "test:mobile": "node mobile.test.mjs",
-    "test": "node test/attachments.test.mjs && node state-checkpoint.test.mjs && node test/usage.test.mjs && node test/operations.test.mjs && node test/hidden.test.mjs && node test/slowfs.test.mjs && node test/spawn-group.test.mjs && node test/revive.test.mjs && node test/repin.test.mjs && node test/codex-repin.test.mjs && node test/opencode-resume.test.mjs && node test/terminal-modes.test.mjs && node test/trace-tail.test.mjs && node test/trace-window.test.mjs && node migration.test.mjs && node resize.test.mjs"
+    "test": "node test/trace-download.test.mjs && node test/attachments.test.mjs && node state-checkpoint.test.mjs && node test/usage.test.mjs && node test/operations.test.mjs && node test/hidden.test.mjs && node test/slowfs.test.mjs && node test/spawn-group.test.mjs && node test/revive.test.mjs && node test/repin.test.mjs && node test/codex-repin.test.mjs && node test/opencode-resume.test.mjs && node test/terminal-modes.test.mjs && node test/trace-tail.test.mjs && node test/trace-window.test.mjs && node migration.test.mjs && node resize.test.mjs"
   },
   "engines": {
     "node": ">=20.19"
--- a/server/reader-info.test.mjs
+++ b/server/reader-info.test.mjs
@@ -319,6 +319,35 @@ try {
     await page.locator('.share-card a[download]').getAttribute('href') === `/api/trace/${id}/download`);
   await page.keyboard.press('Escape');
 
+  // The reply row fills the composer. It stopped when the paperclip left column
+  // 1 of that grid: auto-placed, the row took the `auto` track and sized to its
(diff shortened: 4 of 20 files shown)


lvwerra (commented):
**Verdict: approve with two small things to fix — neither blocking.** The `i` is a genuine tap target with no hover dependency, nothing from the bar was dropped, and the download route does what it says. The two findings are a 17px hit target and a traversal hole the download route inherits (and escalates) from `PUT …/source`.

Reviewed as the person who put most of that bar there (#56, #60, #61, #77). I ran your two new suites and then went at it independently — touch context, `tap()` only, never a mouse.

## Correctness — verified

**The `i` does not recreate the #77 problem.** It is a real `<button>` with `aria-expanded` / `aria-haspopup="dialog"` / `aria-label`, the panel is `role="dialog"`, and I enumerated `[title]` inside the panel: **none** — every fact is text. Tapping it opens, tapping it again closes, Escape closes, and a tap clear of the panel closes. The two facts that were previously `title`-only (cached tokens, the full timestamp) are readable on a phone for the first time, which is strictly better than what #77 landed.

*Two false alarms I chased, so nobody repeats them:* my first outside-tap probe reported "still open" — I was tapping inside the panel's own box. And `▼` looked like it stopped moving the reader until I noticed a search filters the list to the matching turn (10 exchanges → 1), so there was nowhere to scroll; with no search, `▼` moves it. Both fine.

**The download route.** Your test pins the right things — byte-for-byte content, the attachment name, the bundle case, `code: no-trace` for a session that has not spoken, 404 for an unknown id — and all five pass here. Through the UI, the panel's link returns the real 14 KB transcript as `attachment; filename="order.jsonl"`. Not gating it is the right call and the sharp edge (a share-refused session is still downloadable) is stated rather than hidden.

### Finding 1 — the `i` is a 17×17px hit target

Measured on an iPhone 13 viewport: `.cxv-info-btn` is **17×17** CSS px (`▲`/`▼` beside it are 18×18). WCAG 2.2 SC 2.5.8 puts the floor at 24×24 — the spacing exception may well cover it, since its nearest neighbour is across the bar — and iOS guidance is 44. It is consistent with the bar's existing scale, so this is not a regression in style; the reason to raise it is that **this one button is now the only route to five facts that previously needed no tap at all**, on the device the item was filed from.

Cheapest fix keeps the visual size: pad it out and pull the padding back with a negative margin, exactly what #75 did for `.ph-back` (`.pane-head .ph-back { align-self: stretch; margin: calc(-1 * var(--ph-pad-y)) …; padding: 0 8px }`).

### Finding 2 — `ref: ".."` makes the download route read `DATA_DIR/*.jsonl`

`/^[\w.-]+$/` matches `..`, at `PUT /api/trace/:id/source` (index.js:2357) and again in `traceFileOf()`. `path.join(DATA_DIR, 'traces', '..')` is `DATA_DIR`, and the first `*.jsonl` there is served. Reproduced against this branch:

```
POST /api/sessions {"cli":"trace","name":"pane"}            → pane-96e137
PUT  /api/trace/pane-96e137/source {"kind":"bundle","ref":".."}  → 200 {"ok":true}
GET  /api/trace/pane-96e137/download                        → 200
     content-disposition: attachment; filename="pane.jsonl"
     {"version":1,…,"method":"POST","path":"/api/sessions",…   ← operations.jsonl (0600)
```

**Impact today is small and I do not think it blocks:** one operator, and `/api/operations` already serves that same log to the same caller, so nothing is exposed that was not. The reason to fix it here is that the hole changes character in this PR — `/location` disclosed a *path*, this route returns *bytes*, so it is now a file-read primitive whose only guard admits `..`, and whatever `.jsonl` lands in `DATA_DIR` next inherits the exposure. One line in `traceFileOf()`: reject a ref containing `..`, or `path.resolve` and assert the result stays under `DATA_DIR/traces`. Worth a case in `trace-download.test.mjs`, which has no traversal test — and while you are there, the suite also has nothing asserting that a *non-trace* session can only ever resolve to its own transcript (true today, since `ref = s.id` is hard-coded, but that is the property the route rests on).

## Does it solve the operator's problem?

**Inventory: nothing dropped, two facts gained.** I compared the live DOM on both builds with the same fixture rather than reading your table:

| | main | this branch |
|---|---|---|
| bar text | `claude-opus-4 · 10 turns · 53.0k↓ 5.5k↑ · Aug 15` + `▲▼` + search | `i` + `▲▼` + search |
| panel | — | Model · Turns `10 turns · 32 messages` · Tokens `53.0k↓ 5.5k↑ · 180k cached` · Started `Aug 15 · 8/15/2026, 7:00:00 AM` · Download · Share |
| cached tokens, full timestamp | `title` only | **text** |
| hit counter, `▲▼`, search | on the bar | on the bar, still working (`1/2`, 2 highlights, filter note) |

**And it buys real space**, which is what "overloading everything" was about — iPhone 13, same session:

| | main | this branch |
|---|---|---|
| `.cxv-bar` height | 56px (wraps to two rows) | **31px** |
| reader body | 470px | **495px** |
| search field | 352px | 193px |

One correction to the framing, not the change: the body says "on a 390pt pane search got whatever was left". On main it is the opposite — the bar wrapped, so search had a full-width row to itself at 352px, and it is *narrower* now at 193px. The win here is 25px of reading height and a single-row bar, and it is worth stating that way, because a reviewer who measures will find search lost width, not gained it.

**The sidebar widget.** I checked the redundancy claim before agreeing with it: the Files pane really does render traces (`FilesPane.tsx:613,754,801`, content-detected `meta.kind === 'trace'`), per-row read-trace/share/handover survive (`Sidebar.tsx:374`), `SHAREABLE_CLIS` moved into `types.ts` and is now shared rather than a second copy, and the one capability nothing else has — pulling a shared dataset via `POST /api/trace/import` — moved to Settings rather than being deleted. That is the right call and the docs were updated to record it.

## Collisions

Trial-merged all three:

- **#85 (mine, reader ordering): clean.** We both touch `ConversationView.tsx` but not the same regions — yours is the bar and the props signature, mine is the `running=` prop and `Exchange.tsx`. No semantic overlap either: nothing in mine reads the bar.
- **#78: conflicts, exactly as your body says** — `ConversationView.tsx` is only the adjacent import lines (`DownloadGlyph, ShareGlyph` vs `InputRequiredNotice` + `writePaneMode`), keep both; `server/package.json` is the recurring test-script line, take both suites.
- **#76: conflicts** in `Sidebar.tsx`, the props destructuring, as described.

Your collision section matched what I found line for line, which made this part quick.

lvwerra:
Both fixed in `df5a723`, and the second one turned out to be wider than the route this PR adds. Thank you for the measurements — the hit-target one in particular is the kind of thing that reads as fine until someone puts a thumb on it.

## 1. The tap target — 16×16 → 32×28, circle unchanged

I measured it the way you did and then by hit-testing, which is what a finger actually resolves against:

```
before:  circle 17×17   target 16×16
after:   circle 17×17   target 32×28
```

Your suggested idiom does not quite port: `.ph-back` draws no box, so padding plus a negative margin is free there, but this button **is** the circle — padding would inflate the ink. So the target is an overlay (`.cxv-info-btn::before { inset: -8px -9px }`), which grows what has to be hit and leaves the circle at 17px.

Not 44: the bar is 31px tall, so a 44px target would reach out over the conversation and swallow taps meant for the first turn. 28 fills the bar's height, 32 across is comfortably past the 24 floor, and the nearest neighbour (`.cxv-nav`) is 97px away, so the spacing exception was never doing the work anyway.

It is pinned now, in `reader-info.test.mjs`, by sweeping `elementFromPoint` outward from the centre until it stops resolving to the button — so it asserts the *target*, not the CSS technique, and stays honest if someone reimplements it with padding later.

## 2. `ref: ".."` — and it was on the reader's route too

Your repro is right, and the reason to look further was your own framing: the hole "changes character" here. So I checked whether the download route was the only one holding it, and it was not.

`GET /api/trace/:id` — the window route the **reader** calls — had the identical guard and the identical `path.join(DATA_DIR, 'traces', ref)`. Reproduced against this branch with a canary that sorts before `operations.jsonl`:

```
PUT  /api/trace/:id/source {"kind":"bundle","ref":".."}  → 200
GET  /api/trace/:id/download   → 200  attachment; filename="pane.jsonl"   *** canary ***
GET  /api/trace/:id?tail=1     → 200                                      *** canary ***
```

So it was not only bytes-vs-path: the same ref renders an arbitrary `DATA_DIR/*.jsonl` into the reader as a conversation, on `main`, today. A check inside `traceFileOf()` would have left that standing.

The fix is one resolver instead of three copies of a regex:

```js
const TRACES_DIR = path.join(DATA_DIR, 'traces');
function bundleDir(ref) {
  const name = String(ref ?? '');
  if (!/^[\w.-]+$/.test(name)) return null;
  const dir = path.resolve(TRACES_DIR, name);
  const rel = path.relative(TRACES_DIR, dir);
  if (!rel || rel.startsWith('..') || path.isAbsolute(rel) || rel.includes(path.sep)) return null;
  return dir;
}
```

used by the window route, the download route, and `PUT …/source`, which now refuses to persist one at all.

### The tests you asked for, plus the property

Four new checks in `trace-download.test.mjs`, each verified to fail before the fix:

| check | pins |
|---|---|
| a traversing ref is refused before it is stored | `..`, `.`, `../..`, `a/../..` → all 400 |
| a stored traversing ref cannot be downloaded | 400, and the canary is **not** in the body |
| nor rendered as a conversation by the reader | the window route, same canary |
| a non-trace session downloads its own transcript, whatever source is planted on it | the property the whole route rests on |

The second and third are the interesting ones: a rejection at the setter proves nothing about a ref that is **already stored**, which is the state a Space upgraded from a build that accepted `..` boots with. So the test stops the server, writes `sessions.json` directly with `traceSource: { kind: 'bundle', ref: '..' }` on both a trace pane and a normal session, and starts it again. Reverting just the resolver check makes it fail with `{"status":200,"leaked":true}`.

The last one plants a bundle source on a **claude** session and asserts the download still returns that session's own transcript byte-for-byte — pinning the `ref = s.id` hard-coding you flagged as load-bearing but unasserted.

## And your correction to my framing, which was right

> On main it is the opposite — the bar wrapped, so search had a full-width row to itself at 352px, and it is *narrower* now at 193px.

Correct, and my sentence was wrong. The win is 25px of reading height and a single-row bar, not more room for search. I have left the PR body's claim struck rather than quietly rewritten, since your measurement is the one a later reviewer should find.

Green after the changes: `web` typecheck and tests, `server npm test` (156 checks, including the four new ones), and `reader-info.test.mjs` including the new tap-target assertion. Nothing merged, nothing deployed.

lvwerra:
Both done in `2872167`, pushed onto the rebase (`7b9fec1` → `2872167`, base `575f1e0`). The panel itself is untouched.

**[Before/after captures and the measurements →](https://claude.ai/code/artifact/4dd3c0d6-43c3-4e90-a1c4-671f8e766778)**

## 1. It looks like the rest of the app now

`border-radius: 50%` and a filled circle was the only round control in an app built from square hairlines — the `terminal | reader` selector, the search field beside it *in the same bar*, the Overview chips, #81's wrap toggle. It is now the same square mini-control as the `▲▼` nav keys, on the `--r-sm` token the app reserves for mini buttons and chips, keeping its hairline border and panel fill so it still reads as an affordance rather than bare text.

## 2. Both pairs are one size

Both readings were true, so both are fixed:

| | was | now |
|---|---|---|
| the `i`, drawn | 17 × 17 | **18.4 × 18.4** |
| nav keys `▲▼` | 18.3 × 18 | **18.4 × 18.4** |
| corner radius | 50% | **4px** |
| Download | 124.3 × 23.2 | **138 × 26.2** |
| Share | 75.8 × 23.2 | **138 × 26.2** |

The bar pair is sized by one rule now — `.cxv-bar .cxv-mini, .cxv-info-btn` — rather than each falling out of its own padding, so they cannot drift apart again. The panel pair is `flex: 1 1 0`, which is what makes an `<a download>` and a `<button>` the same width.

**One trade, named rather than buried:** *Download trace* became *Download*. At equal widths the longer label wrapped onto two lines at 390pt, and half a pair with a wrapped label is worse than a shorter word. The full phrase moved to the `aria-label`, so it costs a screen reader nothing and adds no hover dependency — **there is still not one `title` inside this panel**, which is the property that makes it readable on a phone at all. The panel is a little wider (19rem) so both labels sit on one line.

## The tap target did not move

34 × 30, hit-tested. It is pinned by the `elementFromPoint` sweep in `reader-info.test.mjs` rather than by reading CSS, which is exactly why squaring the button did not disturb it — the test measures what a finger resolves against, not the technique. On the specific thing you flagged: `border-radius: inherit` is gone from the overlay, since there is no circle left to inherit from and a rectangle is what the sweep measures. All 17 checks in that suite pass, `web` typecheck and tests are green.

Re-verified at 390pt in a touch context, which is where both captures come from. Not merged, not deployed — yours to redeploy when you want it.

lvwerra:
Pushed as `d423b7a`, on top of your `a842eba` (upright `i` kept — fetched first, nothing overwritten).

**[Captures: the header in both views, phone and desktop →](https://claude.ai/code/artifact/f235c0b0-205d-4218-8034-c3c9e06564ca)**

## One `i`, in the pane header — the argument

I went with your lean, and here is the reasoning rather than just the result.

**Why one and not two:** these are facts about the **session**, not about the reader. A pane showing a terminal has the same model, the same token total and the same start date as the same pane showing the transcript. So "reachable from both views" is satisfied by one button that does not move between them — and the reader's toolbar loses its last non-control, leaving `▲▼` and search, which is the direction the original complaint pointed. Two would mean two tap targets to keep to the 24px floor, two panels to maintain, two sets of test selectors, and a second whole-file read in the view that has already done it.

**Why not `.ph-title`, which is what the operator's words point at:** that column is a rename field. Its width flexes with the name (`.ph-name` never shrinks; `max-width` bounds it), so a control inside it drifts sideways as the name grows, and it would sit inside the double-click-to-rename target. It went into `.ph-left` instead — beside the logo and the state dot, the cluster that already means *what is this pane* — which is a fixed anchor at every width and in both views.

**One thing the first capture caught:** the panel was anchored to the button, and the button sits a back arrow, a logo and a state dot into the row, so at 390pt a 19rem panel started ~100px in and ran off the right edge. It is anchored to `.pane-head` now, with `max-width: calc(100% - 16px)`, and measured inside the pane at both widths (phone 321 ≤ 382, desktop 607 ≤ 1168).

## The two things you flagged

**The facts are not loaded in terminal view, so the read is lazy and says so.** The summary is fetched when the panel is *first opened*, once per session — no terminal pane pays for a whole-file parse nobody asked to see. States, all in the captures:

| state | panel says | actions |
|---|---|---|
| reader mounted | the facts, **no request** — the reader hands its head down via `onHead` | Download · Share |
| terminal, first open | `reading the transcript…`, then the facts | Download · Share |
| terminal, re-opened | the facts, **no second read** | Download · Share |
| agent has not spoken | `No transcript yet — this agent has not written one. It appears here once it answers.` | Share only |
| read failed | `Could not read the transcript. Close this and try again.` | Share only |

**Download for a session with no trace:** the link is only rendered once the panel knows there is a file. Your point about `code: no-trace` is exactly why — a button that 404s is worse than a sentence saying why it is not there. Share stays, because the share dialog explains its own refusal (`This session hasn't written a transcript yet`).

## The tap target survived, in both views

`inset: -8px -9px` is untouched. Hit-tested at **34 × 32 in the terminal view and 34 × 32 in the reader**, against a 19px drawn box. The `elementFromPoint` sweep is now run twice — once per view — so the floor is pinned where the button actually lives rather than where it used to.

`reader-info.test.mjs` is 24 checks now and covers both views: zero summary reads before the panel opens, the in-flight wording, no second read on re-open, the no-transcript sentence with no download link, and the tap target in each view. Web typecheck and tests green; all three browser suites green (52 checks).

`docs/conversation-view.md` §3.3 records the move, the placement reasoning and the lazy read. Not merged, not deployed — yours to redeploy.

lvwerra:
Done in `8ee3bc0`, rebased on `575f1e0`. This supersedes the placement I chose last round.

**[Captures: both views, both widths, search open →](https://claude.ai/code/artifact/0d72bdd0-9405-4e02-bb17-93a21ca3dd7b)**

## Three zones

| | holds | |
|---|---|---|
| **left** | back · logo · `workspace/ni/` | the path is not a control, so it left the control cluster |
| **centre** | `■ nightly-index` | the state mark leads the name, both centred as one title; it never shrinks, so the name still gives way first |
| **right** | attach · search · `i` · `×` | one `.ph-btn` rule: 22px of ink, no boxes, 8px apart |

The `i` gave up its box to join that rule and `×` came off `mini-btn`, so all four are one contract and cannot drift apart — which is what you asked for after the `i`/`▲▼` mismatch.

## The five things you flagged

**1. The tap target survived losing the box, and the sweep caught a real regression.** The overlay is still there, but with the boxes gone and a 4px gap, **each button's overlay reached across its neighbour's ink** — the middle two measured **20 × 28**, under the 24px floor, and invisibly so. At 8px each is **28 × 28**. That is why the gap is 8 and not 4, and the sweep now measures **all four buttons in both views** rather than the `i` alone.

**2. The phone.** Moving the path left created no room, so it is still hidden under `520px` — and the info panel gained a **Folder** line to carry the fact. That line needs no read, so it is there even for an agent that has not spoken. The header does not wrap and the title still centres at 390pt; the captures show it.

**3. Closing search clears the query.** Not just the box — a live filter behind a hidden search box is a reader showing three of forty turns with nothing to say why, which is exactly the trap you named (and which a reviewer already fell into, reading it as dead `▼` keys). Pinned: filter to zero turns, close, and the turns come back; re-open and the box is empty.

**4. The paperclip changed hands, not jobs — it works, so I implemented it as asked.** In the reader it opens the composer's *own* picker: the reader registers an opener upward through `onAttachPicker` while it is mounted, so a header click still puts files in the draft being typed, with the reader's own disabled state and reason. The composer no longer draws a second picker. #76's suite is updated to the new contract (`one paperclip, in the header, wired to the reader composer`) rather than left asserting the old one.

**5. One style, one rule.** `.pane-head .ph-btn` sizes and spaces all four; the sweep asserts a single ink size across the cluster, so a future fifth button inherits it or fails the check.

**One judgement call worth flagging:** the search switch appears in the reader only. It searches a transcript, and the terminal view has none, so the cluster is four buttons while reading and three while watching. The alternative — a search icon in the terminal view that silently switches you to the reader — seemed worse than an honest absence, but say the word and I will make it do that instead.

`reader-info.test.mjs` is 30 checks now, including the three zones asserted by geometry (path in `.ph-left`, state mark left of the name, the four in order). All three browser suites green (59 checks), server suite green, web typecheck and tests green. Not merged, not deployed.

lvwerra:
Fixed in `6b8c956`. Your diagnosis was right, and I measured it before changing anything.

**[Before/after and the measurements →](https://claude.ai/code/artifact/0d72bdd0-9405-4e02-bb17-93a21ca3dd7b)** (new section at the end)

## Confirmed, with numbers

`.ov-composer` is `auto minmax(0, 1fr)` and column 1 **is** the paperclip. With the reader's picker moved to the header, `Attachments` renders nothing until a file is attached, so the composer had a single child — auto-placed into the `auto` track, sizing to its own content:

```
reader, empty   reply row 176 of 872   columns: 175.8px  664.2px
reader, typed   reply row 210 of 872   send key 650px from the right edge
reader, chips   reply row 848 of 872   columns: 0px 840px   ← already correct
```

That last line is why it only showed with an empty composer: `has-images + .ov-live { grid-column: 1 / -1 }` was already spanning the row whenever a chip was present.

## I did not collapse it to one column, and here is why

The second column is not the empty one. **The Overview card still draws its own paperclip in column 1** — it has no pane header to move it to — and its composer measures correct today: paperclip 28px, reply row 568px, both on one row, picker left of input. Collapsing the grid would drop that paperclip onto a row of its own to fix a bug in a different surface.

So the reply row names its column instead:

```css
.ov-composer > .ov-live { grid-column: 2 / -1; }
```

The card keeps its two columns unchanged; in the reader the empty `auto` track measures **0px** and the row fills the composer. The grid *does* end up one column wide in the reader — it gets there by measuring rather than by a second template, which is one line instead of a modifier class plumbed through `Composer` from two call sites.

The `has-images` / `has-note` spans stay rather than being cleaned up: they put the chip row and the reply row on separate full-width rows, which still matters in the card where column 1 is 28px.

## Every state, both builds

| state | reply row was | now | send to edge |
|---|---|---|---|
| reader, empty | **176** of 872 | **840** of 872 | — |
| reader, typed | **210** of 872 | **840** of 872 | 650 → **12** |
| reader, four lines | **210** of 872 | **840** of 872 | 650 → **12** |
| reader, chips attached | 848 | 848 | 12 → 12 |
| Overview card | 568, paperclip beside it | 568, unchanged | 0 → 0 |

**Item 10 is not regressed:** with four lines of draft the send key still centres on the first line (`--ov-first-line`), asserted in the same measurement rather than eyeballed.

## Pinned twice, both failing on the broken build

- `composerAlign.test.mjs` — asserts `.ov-composer > .ov-live` exists and reaches `-1`; deleting the line fails it. (Its regex is anchored: `.ovw-win .ov-composer { flex: none }` contains the same selector as a substring and matched first on my first attempt.)
- `reader-info.test.mjs` — measures the *rendered* row against the composer's right edge, empty and typed. On the pre-fix build: `604px` and `570px` short. That is the one that would catch a future cause the CSS pin cannot see.

Web typecheck, web tests, and all three browser suites green (61 checks). Not merged, not deployed.

PR #85, opened 2026-08-16
Reader: the answer where it was said, one working line, a still turn column

lvwerra:
Three items from **iteration two** of `improv.md`, all in the reader's layout. Each was reproduced on a running instance and measured on both builds before anything was edited.

**[Before/after captures →](https://claude.ai/code/artifact/b0ff0c11-3b20-4433-857c-7fe14a2a216d)**

## 1. The last assistant message was printed after work it predates

> the last assisstant message is sometimes rendered at the very end even when the tools are expanded which looks odd as it's not in order with the tools

The substantive one, and it is an ordering bug in `splitExchanges`, not spacing. **Source order first:** `markFinalTurns` (server/src/traces.js:1496) marks the last assistant *text* turn before the next user turn as `final` — that text can be followed by more tool calls. When an exchange doesn't END on the agent's words, `splitExchanges` searched backwards for that `final`, **spliced it out of the middle of the steps** and rendered it below all of them. ~~`unmarkTrailingFinal` means this only bites *completed* exchanges, which is why it shows up when scrolling back rather than while watching.~~ **Corrected after review:** `unmarkTrailingFinal` only runs when the window does *not* reach the end (`traces.js:1720`, `if (!atEnd && openEdge)`), so on a live tail the trailing `final` *is* marked and a turn that answers and then starts another tool gets `answerAt` while you are watching it. The behaviour is the same either way; the blast radius is larger than that sentence claimed.

The fixture makes it self-evident — the reply promises the two things printed above it:

```
✓ Bash  psql -c "explain analyze select …"
✓ Edit  …/migrations/014_usage_index.sql
✓ Bash  node scripts/nightly.mjs --dry-run
Found it — no index behind the group-by. I will add one and re-run.
```

`Exchange.answerAt` now records the index the answer was lifted from, and `ExchangeView` renders the work as two runs with the answer between them: **Bash → answer → Edit → Bash**. Step grouping runs over each side separately on purpose — two `Read`s either side of the reply are two things the agent did, and collapsing them into `Read ×2` would erase the sequence this exists to keep.

The card shares the renderer, so it gets the same treatment when one of its exchanges has an `answerAt`.

## 2. Two working lines could be on screen at once

> the working widget in the reader is sometimes below and above.

Send a prompt while the agent is still working and you get both: the live exchange's `working` at its bottom, and the optimistic echo's `working` below that. Measured 2 → 1 at both 1280 and 390.

The Overview card has guarded exactly this since it grew an echo (`running={running && !justSent}`); the reader had not. One line changes.

**Which position, and why** — either is defensible, so: **the last thing in the reader, once.** The line carries what the agent is doing *now* (`working · Bash …`), so it continues the step rail rather than preceding it. Putting it above the steps would report the present before the past — the same disorder as item 1. It also sits where new content lands, which is where the reader is already looking while it follows the tail. Consequence worth knowing: with the work collapsed there is nothing between the meta row and the line, so it sits right under the prompt; expanded, it is below the steps. That is the same rule ("last"), not a second position.

## 3. The turn column stepped sideways at ten

> the time also doesnt seem to be super aligned in the reader

Ambiguous, so I measured all the candidates before picking one:

| candidate | measured |
|---|---|
| clock right edges, 10 turns | `1253` on every row, both builds — already aligned |
| baselines within the meta row (summary / turn / clock) | 0.0px apart |
| clock vs the prompt band's right edge | 14px (desktop) / 8px (phone) — the band's deliberate full bleed from #49 |
| **`turn N/M` left edge** | **1126 for single digits, 1119 at `turn 10/10` — the cluster shifts 7px** |

So the clock is stable and its *neighbour* is not: the whole right-hand cluster steps left the moment a session passes nine turns. That is the one thing visibly off, and it is what I fixed — the number is padded to the width of the total with figure spaces (U+2007), which is exactly a digit wide in this mono, tabular row. Distinct left edges: 2 → 1.

I left the band-edge difference alone: that is #49's full bleed, the band is a tint rather than text, and pulling it in would reopen "the band reads as a card floating over the answer".

## Tests

All mutation-checked — I reintroduced each bug and watched the check fail:

| test | pins | fails when |
|---|---|---|
| `exchanges.test.mjs` | `answerAt` is 1 for an answer followed by two tool calls; the steps split correctly either side; `undefined` when the answer really is last | `x.answerAt = i` is removed |
| `pendingExchange.test.mjs` | the component renders two `.cx-steps` runs around the answer, one when it is last, and the padded turn cell (`turn␇9/10` as wide as `turn 10/10`) | the render ignores `answerAt`, or the padding is dropped |

The render test bundles the component with markdown stubbed — sanitising wants a DOM, and what these check is which order the blocks come out in.

## Collisions

Trial-merged both open PRs against this branch: **#76 clean, #78 clean.**

One semantic note for whoever lands second: **#78 renders `InputRequiredNotice` at the very bottom of the reading column** (after the pending echo) — the same "last thing in the reader" slot this PR assigns to the working line. No textual conflict, and the ordering is probably right as-is (a blocked prompt outranks "working"), but the two are adjacent and it is worth deciding whether `working` should be suppressed while an input-required notice is showing, since the agent is waiting rather than working.

I did not touch `.cxv-bar` — agent-manager-1 owns it this round.

## Suites

Web green (including the two new files). Server: `resize.test.mjs` and `migration.test.mjs` each failed once during another agent's concurrent run and pass in isolation — the known fixed-port contention.

### DIFF
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -268,7 +268,9 @@ pane with nothing to render (a shell) simply stays a terminal.
   above. Only the last exchange shows it, and while an optimistic echo is pending the echo owns
   the spot: the agent is one process working on one thing, and two `working` lines stacked (the
   live turn's and the echo's) is what read as the indicator being "sometimes below and above".
-  The card has had this guard since it grew an echo; the reader had not.
+  The card has had this guard since it grew an echo; the reader had not. The cost of choosing the
+  echo's line is that it is the bare one — the live turn's carries `· what it is doing now`, and
+  that detail is gone for the second or two until the transcript catches up and the echo clears.
 - **The turn column holds still.** `turn 9/10` and `turn 10/10` are different widths, so the whole
   right-hand cluster stepped sideways the moment a session passed nine turns. The number is padded
   to the width of the total with figure spaces (U+2007) — the row is mono and tabular, so a figure
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -253,6 +253,26 @@ pane with nothing to render (a shell) simply stays a terminal.
   tool call is unusable while a task runs; one that never moves makes you chase it.
 - **The prompt band sticks to the top** while you read a long turn. What you want overhead deep
   in someone's 67-step answer is the question it is answering — not a row of numbers.
+- **An exchange reads in the order it happened.** Usually the answer is the last thing the agent
+  said, so it goes under the work. But an agent that answers and then keeps going — a resumed
+  task, a reply followed by more tool calls — leaves its answer in the *middle* of the turn, and
+  the reader used to lift it out and print it under work it predates. With the steps expanded that
+  reads as nonsense: "I will add the index and re-run" sitting below the edit and the re-run.
+  `answerAt` (exchanges.ts) remembers the index the answer was lifted from, and the work renders as
+  two runs of steps with the answer between them. Grouping runs over each side separately, so two
+  `Read`s either side of the reply stay two rows rather than collapsing into `Read ×2` and erasing
+  the sequence.
+- **One working line, and it is the last thing in the reader.** It carries what the agent is doing
+  *now* (`working · Bash …`), so it belongs at the end of the rail it continues — putting it above
+  the steps would report the present before the past, which is the same disorder as the bullet
+  above. Only the last exchange shows it, and while an optimistic echo is pending the echo owns
+  the spot: the agent is one process working on one thing, and two `working` lines stacked (the
+  live turn's and the echo's) is what read as the indicator being "sometimes below and above".
+  The card has had this guard since it grew an echo; the reader had not.
+- **The turn column holds still.** `turn 9/10` and `turn 10/10` are different widths, so the whole
+  right-hand cluster stepped sideways the moment a session passed nine turns. The number is padded
+  to the width of the total with figure spaces (U+2007) — the row is mono and tabular, so a figure
+  space is exactly a digit — and the clock beside it was already stable.
 - **The prompt band spans the pane.** Reaching into the left gutter but stopping at the text
   column on the right made it read as a card floating over the answer rather than as the head of
   it. Full bleed both sides; the meta row sits tight under the band it belongs to, and one
--- a/web/src/components/conversation/ConversationView.tsx
+++ b/web/src/components/conversation/ConversationView.tsx
@@ -497,7 +497,13 @@ export default function ConversationView({ session, paused, isMobile, readOnly, 
                 total={exchanges.length}
                 q={q || undefined}
                 baseModel={head.model || undefined}
-                running={live && x === exchanges[exchanges.length - 1]}
+                // One working line in the reader, and it is the last thing on
+                // the page. While an echo is pending it owns that spot — the
+                // agent is one process working on one thing, and two `working`
+                // lines stacked (the live turn's and the echo's) is what read
+                // as the widget being "sometimes below and above". The card has
+                // had this guard since it grew an echo; the reader had not.
+                running={live && x === exchanges[exchanges.length - 1] && !sent}
               />
             </div>
           ))}
--- a/web/src/components/conversation/Exchange.tsx
+++ b/web/src/components/conversation/Exchange.tsx
@@ -22,6 +22,10 @@ const textOf = (t: TraceTurn | TraceTurn[] | null) =>
 const moreOf = (t: TraceTurn | TraceTurn[] | null) =>
   blocksOf(t).reduce((n, b) => n + (('more' in b && b.more) || 0), 0);
 
+/** `9` under `10` — a figure space (U+2007) is a digit wide in tabular figures. */
+const padTurn = (n: number, total?: number) =>
+  String(n).padStart(String(total ?? n).length, '\u2007');
+
 const moreLabel = (more?: number) =>
   (more ? `+${more > 1024 ? `${Math.round(more / 1024)} KB` : `${more} chars`} not retained` : '');
 
@@ -169,7 +173,14 @@ export function ExchangeView({
   const [selfOpen, setSelfOpen] = useState(false);
   const toggle = onToggle ?? (() => setSelfOpen((o) => !o));
 
-  const steps = useMemo(() => stepsOf(x.steps), [x.steps]);
+  // The work, split where the answer belongs. Grouping runs over each side
+  // separately on purpose: two Reads either side of the agent's reply are two
+  // things it did, before and after saying it, and collapsing them into one
+  // `Read ×2` row would erase exactly the sequence this is here to keep.
+  const at = x.answerAt;
+  const stepsBefore = useMemo(() => stepsOf(at == null ? x.steps : x.steps.slice(0, at)), [x.steps, at]);
+  const stepsAfter = useMemo(() => (at == null ? [] : stepsOf(x.steps.slice(at))), [x.steps, at]);
+  const steps = useMemo(() => [...stepsBefore, ...stepsAfter], [stepsBefore, stepsAfter]);
   // A turn whose match is buried in the work unfolds it, or the search would
   // report a hit with nothing on screen to look at.
   const hitInWork = useMemo(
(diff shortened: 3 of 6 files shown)


lvwerra (commented):
**Verdict: approve with one ask** — add a fixture for `answerAt === 0`. Everything else here holds up, including under mutation. Not requesting changes: the shipped behaviour is correct, the gap is in what the tests pin.

I re-measured all three items on both builds rather than reading the diff, using a mocked ten-turn trace with a live last turn that answers in the middle of its work.

## Axis 1 — correctness

**The ordering change is right, and the boundaries hold.** I ran `splitExchanges` over the awkward shapes:

| shape | result |
|---|---|
| answer, then more work | `answerAt: 0` — reachable, see below |
| turn ends on a tool, no text at all | `answerAt: undefined`, nothing lifted |
| text → tool → text(final) → tool | `answerAt: 2`; the earlier plain text stays a step, the final is lifted |
| prompt only, no steps | no answer, no crash |
| two `final` turns in one exchange | lifts the **last**, earlier one stays a step |
| `final` whose text is blank | `saidSomething` false, nothing lifted |

**Your claimed pins are real** — I reintroduced two bugs myself rather than taking the table's word: dropping `x.answerAt = i` (`exchanges.ts:99`) fails `exchanges.test.mjs`, and moving the `stepsAfter` block above the answer (`Exchange.tsx:244`) fails `pendingExchange.test.mjs` at *the work it promised is below the answer*.

### The ask: nothing pins `answerAt === 0`

`answerAt: 0` is reachable — an agent that replies **before** doing the work and never speaks again, which `markFinalTurns` marks at index 0 of the steps:

```
splitExchanges([user('add an index'), agent([text('I will add one now.')], 'final'),
                agent([Edit …]), agent([Bash …])])
  → { answerAt: 0, steps: 2 }
```

`Exchange.tsx:176-178` handles it correctly *because* it tests `at == null`. Swap that for the idiom almost everyone writes — `!at` — and:

- every test in `web/` still passes, both files, no failures;
- and the rendered turn puts the answer back **under both tool rows**, i.e. exactly the bug this PR exists to fix. I rendered it: `{"answerBeforeEdit": true}` as written, `{"answerBeforeEdit": false}` with `!at`.

So the headline fix is one token away from silently regressing, with a green suite. A third case in `exchanges.test.mjs` (assert `answerAt === 0`) plus a render case with `answerAt: 0` closes it.

### Minor: the collapsed summary's step count now depends on where the answer was said

`stepSummary(x, steps)` (`Exchange.tsx:194`) counts **grouped rows**, and grouping now runs per side. Same work, same tools, two different summaries:

```
4 Reads, answer last          →  "1 step · 4 tools · 1m 0s"
4 Reads, answer at index 2    →  "2 steps · 4 tools · 1m 0s"
```

Cosmetic, and defensible — the count matches the rows you see when you expand — but two turns that did the identical thing now report different amounts of work while collapsed. `4 tools` stays stable, which is the more meaningful half. If you'd rather it not move: `stepSummary(x, stepsOf(x.steps))` counts the whole turn while the render keeps the split.

### One correction to the PR body

> `unmarkTrailingFinal` means this only bites *completed* exchanges, which is why it shows up when scrolling back rather than while watching.

`unmarkTrailingFinal` only runs when the window does **not** reach the end (`traces.js:1720`, `if (!atEnd && openEdge)`). On a live tail the trailing `final` *is* marked, so a turn that answers and then starts another tool gets `answerAt` while you are watching it. My probe's last turn was exactly that shape and rendered split, live. The behaviour is right — the sentence just understates the blast radius, which matters for anyone judging the risk of this change.

**The `!sent` guard is sound.** I traced the lifecycle rather than assuming: `sent` clears when the trace catches up (`ConversationView.tsx:171`) and on a failed send (`:191`, which also restores the draft), and `PendingExchange` always renders its own `working`. There is no window with zero working lines.

## Axis 2 — does it solve the operator's problem

Measured in a rendered reader, `main` vs this branch, same fixture:

| item | main | #85 |
|---|---|---|
| answer vs the work that followed it | `answerBeforeLaterWork: false`, 1 step run | **`true`, 2 step runs** |
| `turn N/M` left edges over 10 turns | **2 distinct** | **1** |
| clock right edges | 1 | 1 (already fine — your measurement was right) |
| working lines with an echo pending | **2** (`working · result created`, `working`) | **1** |

All three fixed, and the working line's rule reads as one place: last in the exchange, last on the page.

**One consequence worth stating:** the line that survives is the echo's bare `working`. On main the live turn's line carried `working · result created`; while an echo is pending you now get no "what it is doing now" detail until the trace catches up. Transient, and better than two lines — but it is a real loss in that window, and if the echo ever sticks (a prompt the trace never matches) it is not transient.

## Collisions

Four trial-merges, both directions, all **clean**: #85↔#84, #85↔#78.

I went further on #84 since it is mine and we both edit `ConversationView.tsx`: merged the two, typechecked, ran both web suites and my `reader-info` browser suite, then re-ran the three measurements above on the merged build. All of #85's fixes hold and my suite passes. **Neither of us needs to rebase** — your `running=` line and my `.cxv-bar` rewrite are different regions of the same file.

On your #78 note: I agree the ordering is right as-is, and I would go further — `working` should probably be suppressed while an input-required notice is showing, because the agent is blocked rather than working, and two "state of the agent" lines stacked is the same complaint this PR is fixing. Worth deciding rather than inheriting, whichever of you lands second.

lvwerra:
Ask taken, plus the two smaller points. Pushed as `fedb89a`.

## `answerAt === 0`, in both layers

You were right that it is reachable and unpinned, and right about *why* it matters: zero is the value a falsy guard drops.

- `exchanges.test.mjs` — your fixture, asserting `answerAt === 0`, nothing above it, and both tool rows below.
- `pendingExchange.test.mjs` — a render case with `answerAt: 0`, asserting the answer precedes the first tool row and that there is exactly one run of steps (nothing above it to run).

I ran your mutation to check the case sits where the regression is: with `at == null` swapped for `!at`, **the new render case fails and nothing else does** — `exchanges.test.mjs` stays green, correctly, because the splitter is not what breaks. So the hole is closed at the layer it lives in.

## The summary's step count — keeping it, with the reasoning written down

I went with counting over the split, and put the trade-off at the call site rather than leaving it for the next person to rediscover:

```
// Counted over the split, deliberately: the fold control says how many rows
// are behind it, so "2 steps" has to mean two rows when you open it. The cost
// is that a turn whose answer landed mid-work can report one row more than an
// identical turn whose answer came last — same tools, same time, different
// grouping. `stepSummary(x, stepsOf(x.steps))` would count the whole turn
// instead and disagree with what unfolds; between the two, the number that
// matches what you are about to see wins.
```

Your alternative is the better one if the summary is read as "how much work was this"; I think it is read as "how much is behind this triangle", and a fold control that lies about its own contents is the worse of the two. Happy to flip it if you disagree — it is one argument.

## The `unmarkTrailingFinal` correction

You are right and I have struck that sentence in the PR body. `unmarkTrailingFinal` only runs when the window does not reach the end (`traces.js:1720`, `if (!atEnd && openEdge)`), so on a live tail the trailing `final` *is* marked and a turn that answers and then starts another tool gets `answerAt` while you are watching. The behaviour is the same; my sentence understated who sees it, which is the part that matters for judging the change.

## The working line's lost detail

Named in the docs rather than fixed, deliberately:

> The cost of choosing the echo's line is that it is the bare one — the live turn's carries `· what it is doing now`, and that detail is gone for the second or two until the transcript catches up and the echo clears.

Threading the live turn's `nowDoing()` into the echo would put the *previous* turn's activity under the prompt you just sent, which is accurate but reads as if the new prompt were already running. Your sticking case is real but pre-existing: an echo the trace never matches held two lines before this change and holds one now, so it is strictly less wrong. If it turns out to stick in practice, the fix is to expire `sent` on a timeout, which is its own change.

## On #78

Agreed, and I would take your stronger version: `working` should be suppressed while an input-required notice is showing, since the agent is blocked rather than working, and two state-of-the-agent lines stacked is the complaint this PR is fixing. Whoever lands second should do it.

## What to expect merging this with #84

Re-checked after today's commit — **still clean both directions**, and here is why, so the merger knows what they are looking at:

| file | #85 touches | #84 touches |
|---|---|---|
| `ConversationView.tsx` | one line, the `running=` prop inside the exchange map (~500) | imports (~30), the props signature + `onShare`, `infoOpen` state and its effect (~100), the bar markup (~410–440) |
| `Exchange.tsx`, `exchanges.ts` | all of the ordering and the turn padding | not touched |
| `conversation.css` | not touched | the panel |

No semantic overlap either: nothing in #85 reads the bar, and the turn padding lives in the meta row (`.cx-n`), which the panel does not replace — after both land the panel holds the conversation-level facts and the meta row keeps the per-turn ones.

**One thing that does need resolving, and it is mine to clean up:** `docs/conversation-view.md` ends up self-contradictory. Your §3.3 says the facts "are plain text in the panel"; the paragraph #77 left on main still says the date "is text in `.cxv-bar` beside the other conversation-level facts" (line 342 in the merged file, in the *Below the composer* bullet). Git merges them without complaint because they are far apart. If #84 lands first I will fix that line as part of rebasing; if #85 lands first, it is one sentence in your §3.3 neighbourhood — either way it should not survive both merges as written.

PR #86, opened 2026-08-16
One button on a session row, and archive as the way out

lvwerra:
> **Depends on #84 and must merge after it.** The base of this PR is `feat/trace-share-and-reader-info`, not main — #84 works in the same row.

The sidebar row carried up to four controls. It keeps one: `×`, which archives.

| row | before | after |
| --- | --- | --- |
| a claude session | Start · Read trace · Share · Delete | **Archive** |
| a shell | Start/Stop · Delete | **Archive** |
| archived, retired on purpose | — | **Restore · Delete** |
| archived, merely quiet | — | **Archive** |
| a remote agent | Reconnect/Disconnect · … | **Reconnect/Disconnect · Archive** |

## What went, and why it did not need a new home

**Start** was the row's own `onClick` spelled a second time — both call `onOpenSession(...)`. **Stop** went on the operator's reasoning, which I agree with: an idle CLI costs nothing, so "pause this" is not a real case; the real one is a runaway agent, and in that moment you open the pane, where Ctrl-C interrupts with the CLI's own semantics rather than a button's. `onStopSession` loses its last UI consumer, but `stop()` stays — archiving calls it.

**The remote reconnect/disconnect pair is untouched**, as asked, and I agree it should be: it is a line to another machine, not a local process. Group rows are untouched.

## A correction to the brief: #84 did not remove the trace buttons

The brief said read-trace, share-trace and share-session would already be gone on this base. They are not — #84's own body says *"Per-row trace actions are untouched: read-trace, share, handover, and the trace-pane rows all still work."* What #84 removed was the sidebar's trace *widget*, not the row's buttons. So this PR removes them, and had to find homes:

- **Share a session** — already has one: #84 put Share in the reader's `i` panel. Nothing to do.
- **Read a session's own trace** — that is what reader mode is, and #84's panel now hands over the file too. Its only trigger was this button, so `openTrace` and the pane-spawning path it owned are dead code and go with it.
- **Share a trace / continue from a trace** — these had **no** other home. The brief says to put handover in the `i` panel, but that panel is `ConversationView`'s and a **trace pane does not use ConversationView** — it renders its own view. So both now sit on the trace pane's header, which is the thing they act on. The prefilled create panel still belongs to the sidebar, so the request travels there as an id rather than the panel moving.

## Archiving is stored now, and it stops the agent

It had to become stored: *"I am finished with this one"* cannot be expressed as an absence of activity, which is all the idle window measures — an agent you retire the moment it answers is as active as it will ever be. `archivedAt` lives on the session record, so it survives a reload and reads the same on every device.

`POST /api/sessions/:id/archive` stops the agent and stamps it. `POST …/unarchive` clears the stamp and deliberately does **not** restart: unarchiving says "let me see this again", and starting is what opening the pane does.

**Delete is gated server-side** (409, `code: not-archived`), not just hidden in the UI, so the rule holds for any caller. The refusal is a sentence worth showing, and the toast shows it.

### What archiving a running agent does

It kills the PTY mid-turn (`host.pty.kill()`), and the honest account of what that costs is:

- **Completed turns survive.** The harness writes its own transcript as it goes, so everything already said is on disk and is what the reader shows afterwards.
- **The in-flight answer is lost.** A reply being generated when you archive is gone; nothing replays it.
- **Files it already wrote stay written.** An agent edits the workspace directly, so a half-finished refactor is half-finished on disk, not rolled back.
- **Reopening resumes.** Restore, open the pane, and the harness `--resume`s its pinned conversation — the session picks up its thread rather than starting blank.

So the cost is bounded and recoverable except for one in-flight answer. That is why I did **not** put a confirmation in front of archive. But it is the one place in this design where a single unconfirmed click can lose work someone is waiting on, so if you want a guard, the right one is narrow: confirm only when `state === 'working'`, not on every archive. Say the word and I will add it.

### Auto-archived sessions becoming deletable — I did not do this, and here is why

This is the part you said you were least sure of, and I think the worry is right, so the two roads lead to the same *view* but not the same *powers*:

- a session **you archived** — restore, or delete
- a session **merely quiet** for longer than the window — offered the archive, and nothing else

Three reasons. First, the window's verdict is not a decision: it is a statement about the clock that flips back the moment the setting changes, and letting a config change hand out delete rights to sessions nobody considered is a strange amount of power for a dropdown. Second, it keeps one sentence true — *only a thing you retired can be removed* — where the union version needs a footnote about a month. Third, it costs one click in the rare case and removes a whole class of accident in the common one.

The cost of being wrong here is smaller than it looks in one respect worth stating: `store.remove()` deliberately leaves the workspace folder on disk, so deleting removes the session record and its place in the list, never the work. The button says so.

If you would rather have the simpler rule — everything in the archived view is deletable — it is a one-line change (drop the `retired` split) and I will make it.

### And the no-dialog question

Agreed, no confirmation for delete. The archived view is somewhere you went on purpose, the row is already out of your way, and the files survive. A dialog there would be the third time the system asked you whether you meant it.

## `stopped` stays a state

Untouched, and it must be: it is what every session lands in after a Space restart, and the archive route asserts it explicitly.

## Verified

`server/test/archive.test.mjs`, **16 checks** against a real server rather than mocks:

- a live session refuses to delete — 409, `code: not-archived`, and a sentence with the word *archive* in it; the session is still there afterwards
- an archived one deletes and leaves the tree
- the flag is **on disk** in `sessions.json`, which is the entire reason it is stored
- restore clears it and makes the session undeletable again
- **archiving a running agent stops it** — driven through a real terminal socket, `running: true` → `running: false`
- `stopped` is still the state it lands in
- archiving an unknown id is a 404

Also: web typecheck, web suite, the full server suite (12 files, no retries), and before/after screenshots of the row and the archived view against a live instance, with the button inventory read out of the DOM rather than eyeballed.

**[Screenshots and the button inventory →](https://claude.ai/code/artifact/0172c34d-cd7f-4b87-99e3-aaebd866a755)**

Not verified: how this reads on a phone (checked at desktop width only), and I have not exercised archive against a *remote* agent with a live peer attached — it takes the no-process branch, so it files the session away and leaves the connection alone, which is what the brief asked for.

## Conflicts

- **#85** — clean.
- **#76** — two files, both mechanical, and one of them is the tie-in you flagged:
  - **`server/src/index.js`, 1 hunk.** #76 adds its `?ifNeverStarted=1` guard in exactly the spot this PR puts the archive guard — my comment there predicts it by name. They compose rather than compete:
    ```js
    if (req.query.ifNeverStarted === '1') {
      if (s.everStarted) return res.status(409).json({ error: 'session has already started' });
    } else if (!s.archivedAt) {
      return res.status(409).json({ error: 'archive this session before deleting it', code: 'not-archived' });
    }
    ```
    That gives the wrong-CLI mistake its one-step delete without opening a hole in the rule for anything that has ever run.
  - **`web/src/components/Sidebar.tsx`, 3 hunks**, all in the props destructure and its type. #76 adds `onPrepareQuickStart`/`onAbandonQuickStart`; this removes `onStopSession`, `onShareSession`, `onShareTrace`, `onOpenTrace` and adds `onArchiveSession`, `onUnarchiveSession`, `retired`, `handoverFor`, `onHandoverHandled`. Take both sides. (#76 still carries `onOpenSharedTrace`, which #84 removed — that resolves when #76 rebases onto #84.)

### DIFF
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -474,14 +474,17 @@ pane with nothing to render (a shell) simply stays a terminal.
 
 ### 3.4 What leaves the sidebar
 
-| Today | Tomorrow |
+| Was on the row | Where it is now |
 |---|---|
-| `Read this session's trace` on every agent row (`Sidebar.tsx:245`) | bottom bar → **reader** |
-| `Share this session` on every agent row (`Sidebar.tsx:246`) | pane header → **share** |
-| A `trace` **session row per read** (`App.tsx:openTrace`) | **gone** — no duplicate rows |
-| Trace row's `Share` (`Sidebar.tsx:237`) | trace pane header (imported traces keep a pane) |
-| Trace row's `Handover` (`Sidebar.tsx:238`) | reader / trace-pane footer |
-| Quick-add **Trace** = open a shared dataset (`Sidebar.tsx:482`) | **Settings → Open a shared trace** (2026-08-16) |
+| `Read this session's trace` on every agent row | **reader mode** — a session's own history is a mode of its pane (done, #86) |
+| `Share this session` on every agent row | **pane header → `i` → Share** (done, #86) |
+| A `trace` **session row per read** (`App.tsx:openTrace`) | **gone**, with `openTrace` itself — no duplicate rows (done, #86) |
+| Trace row's `Share` | **trace pane header** — imported traces keep a pane (done, #86) |
+| Trace row's `Handover` | **trace pane header**, beside Share. Not the reader's `i`: that panel is `TraceInfo`, which a trace pane does not use (done, #86) |
+| Quick-add **Trace** = open a shared dataset | **Settings → Open a shared trace** (2026-08-16) |
+| `Start` | the row's own click already opened the pane (done, #86) |
+| `Stop` | **nowhere, deliberately.** An idle CLI costs nothing; a runaway one is interrupted in its pane, where Ctrl-C carries the CLI's own semantics (done, #86) |
+| `Delete` | **the archived view only** — archive a session, then remove it; the server enforces it (done, #86) |
 
 The last row was `stays` until the operator asked for the sidebar's trace widget to go
 (improv.md, iteration two). It moved rather than went away, because it is the one trace
@@ -493,8 +496,10 @@ dataset id, which is furniture for something you do when a link arrives.
 What disappears is the *local* trace pane — a session's own history is now a mode of its own
 pane, not a second entity.
 
-Agent rows keep stop/play and delete. Three glyphs less per row, which is most visible exactly
-where the sidebar is worst: on a phone.
+An agent row is now **one** control: `×`, which archives — it stops the agent and files it
+away, and it is the only route to deleting one. A remote agent keeps its reconnect/disconnect
+pair beside it, because that is a line to another machine rather than a local process. Up to
+four glyphs less per row, which is most visible exactly where the sidebar is worst: on a phone.
 
 ---
 
@@ -585,7 +590,7 @@ the block model already supports.
 - One component to style, so "the overview is more pleasant" becomes true everywhere at once.
 - The card answers "what did it actually do?" without leaving the Overview.
 - A session's history is reachable from the session, not from a second sidebar row.
-- Three glyphs less per sidebar row, and a full-screen card on a phone.
+- A sidebar row down to one control, and a full-screen card on a phone.
 - Removed code: the `turnsLog` stepper in the card, `openTrace`'s pane-creation path, two
   sidebar buttons, `.tv-badge` and the terminal-styled viewer chrome.
 
@@ -668,11 +673,13 @@ Built (this branch):
 5. Tests for the one piece of judgement in the renderer — what counts as the answer, and what
    stays in the work (`web/test/exchanges.test.mjs`, `npm test` in `web/`).
 
+7. The sidebar lost its trace buttons and `openTrace`, and the row came down to one control
+   (#86, §3.4). Share moved into the pane header's `i` with #84.
+
 Not yet, in the order I would do it:
 
 6. `head.prompts[]` (§5) — index, first line and timestamp per prompt, so a surface can draw the
    skeleton and label "show previous turn" before fetching the page that holds it.
-7. The sidebar loses its trace buttons and `openTrace`; share moves into the pane header (§3.4).
 8. Windowing by exchange in reader mode: a collapsed turn is 2–3 rows, so the DOM stays small,
    `head.prompts[]` gives the skeleton up front, and only an opened turn needs its page. The
    measured-height machinery in `TraceView` is reused as-is — what changes is what a "row" means.
--- a/server/package.json
+++ b/server/package.json
@@ -16,7 +16,7 @@
     "test:ui": "node terminal-ui.test.mjs && node screenshot-input.test.mjs && node reader-info.test.mjs",
     "test:screenshots": "node screenshot-input.test.mjs",
     "test:mobile": "node mobile.test.mjs",
-    "test": "node test/trace-download.test.mjs && node test/attachments.test.mjs && node state-checkpoint.test.mjs && node test/usage.test.mjs && node test/operations.test.mjs && node test/hidden.test.mjs && node test/slowfs.test.mjs && node test/spawn-group.test.mjs && node test/revive.test.mjs && node test/repin.test.mjs && node test/codex-repin.test.mjs && node test/opencode-resume.test.mjs && node test/terminal-modes.test.mjs && node test/trace-tail.test.mjs && node test/trace-window.test.mjs && node migration.test.mjs && node resize.test.mjs"
+    "test": "node test/archive.test.mjs && node test/trace-download.test.mjs && node test/attachments.test.mjs && node state-checkpoint.test.mjs && node test/usage.test.mjs && node test/operations.test.mjs && node test/hidden.test.mjs && node test/slowfs.test.mjs && node test/spawn-group.test.mjs && node test/revive.test.mjs && node test/repin.test.mjs && node test/codex-repin.test.mjs && node test/opencode-resume.test.mjs && node test/terminal-modes.test.mjs && node test/trace-tail.test.mjs && node test/trace-window.test.mjs && node migration.test.mjs && node resize.test.mjs"
   },
   "engines": {
     "node": ">=20.19"
--- a/server/reader-info.test.mjs
+++ b/server/reader-info.test.mjs
@@ -6,7 +6,7 @@
  *     including the two that were title attributes a phone cannot open;
  *   - it closes on Escape and on a press outside it;
  *   - it offers the transcript as a file, and that URL really answers;
- *   - it offers Share, which is the same dialog the sidebar opens;
+ *   - it offers Share, which is the only place that dialog opens from now;
  *   - the sidebar no longer carries a trace widget, and Settings does.
  *
  * Set READER_INFO_PUBLIC_DIR to a prebuilt web/dist to skip the build, and
@@ -310,7 +310,8 @@ try {
   check('a press outside closes the panel',
     await waitFor(async () => await page.locator('.tinfo').count() === 0, 2_000));
 
-  // 6. Share, from the reader — the same dialog the sidebar row opens.
+  // 6. Share, from the reader. The sidebar row used to open this dialog too;
+  //    since #86 the panel is the one place it comes from.
   await page.locator('.tinfo-btn').tap();
   await page.locator('.tinfo-actions button', { hasText: 'Share' }).tap();
   const shareOpen = await waitFor(async () => await page.locator('.share-card').isVisible(), 5_000);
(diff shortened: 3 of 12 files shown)


lvwerra:
Rebased onto `90c04f1` and force-pushed. **Base retargeted to `main`** now that #84 has merged. Design is unchanged — one `×` meaning archive, archive stops the agent, delete only from the archived view, no stop button, `stopped` still a state.

## The rebase was smaller than it looked

The eight-file conflict list is an artifact of *how* the rebase is run, not of the change. `#84` was rewritten when it merged, so my branch point (`7b9fec1`) is no longer an ancestor of main — a plain `git rebase origin/main` therefore tries to replay **#84's superseded commits** as well as mine, and those collide with the merged versions of themselves. Replaying only my own commit:

```
git rebase --onto origin/main 7b9fec1 feat/archive-mechanism
```

→ **zero conflicts.** Files like `TerminalPane.tsx`, `ConversationView.tsx`, `conversation.css` are not in my diff at all; they were only in the list because #84's old commits were being replayed over their own successors.

## What I checked by hand, since a clean rebase is not a correct one

- **The row is still the right target.** Main's `row-actions` is byte-identical to what I branched from — #84 as merged still left the trace buttons on the row, exactly as its body said it would. So my removals still remove what they were written to remove.
- **`TraceInfo` is not a new home for handover.** #84's final shape extracted the `i` panel into `components/TraceInfo.tsx`, which is a nice improvement — but it is session-shaped (it fetches that session's trace summary) and only `TerminalPane` renders it. A trace pane still does not use it, so share and handover stay on the trace pane's own header. Giving trace panes an `i` of their own is a reasonable follow-up; it is not a rebase.
- **#87's status marks** sit in the same row and render correctly beside the single control (screenshot below).
- Typecheck, `npm test` in `web/`, the full server suite (18 files, no retries), and `reader-info.test.mjs` all green.

## `reader-info.test.mjs` — extended, not fought

Its existing assertion is about the trace **widget** (`.quick-add` / `.open-trace`), which I do not touch, so it passed unchanged. I added two checks to that same block, since "what the sidebar no longer carries" is precisely what this PR continues:

```
PASS  the sidebar has no trace widget left
PASS  the session row carries exactly one control  titles: ["Archive — stops the agent and files it away"]
PASS  and that control archives
```

Driven through the real browser the suite already boots, so it is the rendered row rather than the source.

I also corrected two comments in that file that my change makes false — it described the panel's Share as "the same dialog the sidebar row opens", and after this the panel is the only place it opens from.

## Docs

This turns out to be **step 7 of `docs/conversation-view.md`'s own sequence** — *"The sidebar loses its trace buttons and `openTrace`; share moves into the pane header (§3.4)"* — so I moved it out of "Not yet" and into the record.

§3.4's table was a `Today | Tomorrow` plan; it now says where each control actually went, including the two the plan did not cover (`Start`, `Stop`) and the one whose destination changed (`Handover` → the trace pane header, not the reader's `i`, for the `TraceInfo` reason above). Two lines that flatly contradicted this PR are fixed: *"Agent rows keep stop/play and delete"* and *"Three glyphs less per sidebar row"*.

## Verified on the rebased build

Booted the rebased build and drove it, rather than trusting the diff:

| | controls |
| --- | --- |
| working list | `Archive` |
| archived, retired on purpose | `Restore` · `Delete` |
| archived, merely quiet | `Archive` |

(Port checked as free before binding, and the served bundle hash checked against `dist/index.html` — I measured another agent's app that way once already this week.)

lvwerra:
Merged `fix/immediate-upload-progress` (#76) into this branch, so the combined result lives here and #76 stays untouched. **#86 now depends on #76 and must merge after it.**

Both conflicts resolved, plus a third thing the merge turned up that is worth reading — it is a consequence of this PR's contract, not of the merge.

## 1. The delete route — both guards, because they answer different questions

Implemented as the comment promised, with #76's own 409 intact:

```js
const claimedNeverStarted = req.query.ifNeverStarted === '1';
if (claimedNeverStarted && s.everStarted) {
  return res.status(409).json({ error: 'session has already started' });
}
if (!claimedNeverStarted && !s.archivedAt) {
  return res.status(409).json({ error: 'archive this session before deleting it', code: 'not-archived' });
}
```

`?ifNeverStarted=1` is the **caller's** precondition — "only if this never ran" — so a started session is refused in #76's words even where the archive rule would have allowed it (an archived session that has run, asked for with the flag, is still refused: the caller asked the other question). Without the flag, delete stays archived-only. They compose into the exemption this guard was written expecting.

Six new checks in `archive.test.mjs` (22 total) pin the seam:

```
PASS  a session that never ran deletes in one step, no archiving first  status 200
PASS  a started session is refused for ifNeverStarted  status 409
PASS  and in #76's words rather than the archive guard's  {"error":"session has already started"}
PASS  without the flag, the archive rule still answers  {"code":"not-archived"}
PASS  so the session survives both refusals
```

## 2. The sidebar row — one row, not two prop lists

**Nothing of #76's needed reinstating, and I checked rather than assumed.** Every reference #76 has to `onStopSession`, `onShareSession`, `onOpenTrace`, `onShareTrace` and `isShareable` is the *pre-existing* row markup this branch deletes — its diff against main adds no new line touching any of them. So the row keeps its single `×`, and `onPrepareQuickStart` / `onAbandonQuickStart` survive untouched, since they are about creating a session to hang an attachment on rather than about the row's buttons.

`isShareable` and `hiddenSessionIds` lose their last users and go with the buttons — the second was already dead on `main`.

## 3. What the merge surfaced: DELETE is a contract, not a UI rule

`screenshot-input.test.mjs` clears the session list twice with a bare `DELETE`. Since this branch the server refuses that for anything unarchived, so both loops silently 409'd and left their fixtures behind — which is what actually failed:

```
FAIL  fresh-visit attachment creates only a stopped target …
      [{"id":"screenshot-e2e-…","running":true,"everStarted":true}, …]
FAIL  Escape discards the unlaunched target and its upload across reload
```

One cause, two symptoms. Both teardowns are of sessions that **have** started, so `?ifNeverStarted=1` is not their route — they archive first, exactly as the sidebar does. Back to 31/31.

**The general point for the dev deploy:** this guard applies to every caller, not just the sidebar, so anything deleting a session programmatically now archives first. That is the intended design — enforcing it server-side is why the rule holds for a caller that is not the UI — but it is a real API change and this is the first thing to trip on it. `migration.test.mjs` has three more such cleanup calls; they are `.catch(() => {})` with nothing asserted after them, so they no-op harmlessly in a throwaway `DATA_DIR` and that suite stays green. I left them rather than churn an unrelated file, but they are now no-ops and someone should know.

## Verified

| suite | result |
| --- | --- |
| `server npm test` (18 files, incl. `archive.test.mjs` 22/22) | green |
| `server/screenshot-input.test.mjs` | **31/31**, same as #76 scores alone |
| `server/reader-info.test.mjs` | green, incl. this PR's two row checks |
| `terminal-ui.test.mjs` | green |
| `web npm test`, typecheck, build | green |

A process note, since it nearly fooled me: my first "pass" of `screenshot-input` was run against a **stale dist** left over from building #76 for the baseline, and `reader-info` then failed showing the old four-button row — a build artifact, not a regression. Every result above is from a dist rebuilt on this branch and verified to contain this PR's own strings before the run.

lvwerra:
Done — the archived view's delete is a trash can now (`0e798da`).

The reasoning is worth keeping in the record, because it is the whole justification: on a working row `×` means **archive**, which is reversible. Reusing that mark for **delete**, which is not, dresses the permanent act as the recoverable one. Two acts, two marks. It is the `TrashGlyph` that `FilesPane` and `SkillsEditor` already use, so destructive looks the same wherever it turns up.

Checking the rest of the row read correctly, as asked, turned up two more things — both mine, both from the original PR:

**Restore was a ▷, and that glyph is already taken on this row.** `PlayGlyph` means *Reconnect* for a remote agent, so an archived remote showed **two identical triangles meaning different things**, side by side. It also promised a start that unarchiving deliberately does not do — the server route says so explicitly. Restore is an arrow back up into the list now, which cannot be confused with the reconnect beside it.

**`.danger-hover` was never defined.** I put that class on the delete button when I wrote this PR and never added a rule for it, so the one destructive control in the app's list had no destructive affordance whatsoever. It now goes red under the pointer, mirroring `.tw-act.danger` — the idiom the Files pane's row trash already uses — rather than `.mini-btn.danger`, which is the filled red button meant for a surface whose whole point is the action, and too loud for one glyph among others in a row.

Every control in the row now carries an `aria-label` naming the act. Read out of the live DOM:

| view | controls |
| --- | --- |
| working row | `Archive` — `×` |
| archived, retired | `Restore` — `↑` · `Delete` — 🗑 (`danger-hover`) |
| archived, merely quiet | `Archive` — `×` |

Nothing says "close" any more.

**Verified:** `archive.test.mjs` 22/22, `web npm test` green, typecheck and build green, and `reader-info.test.mjs` green — including its two row assertions, which still read `["Archive — stops the agent and files it away"]` on the working row. No test asserted on the archived view's `×`, so nothing needed updating there. The glyphs were confirmed by reading the rendered `path` data and `aria-label`s out of the DOM, not by eye, and then looked at.

Presentation and labelling only: archive is still the row's one control, delete still lives only in the archived view, never-started sessions still delete in one step.

lvwerra:
Follow-up: main moved while I was pushing the trash-can change — **#76 and #78 both landed** — so this branch went red for a few minutes. Merged main in; back to MERGEABLE / CLEAN.

Both conflicts were the two you said you had already handled on your side, and both are unions rather than choices, so I resolved them here as well rather than leave the PR red for the assembly:

- **`web/src/types.ts`** — `archivedAt` (this branch) and `inputRequired` (#78) are independent optional fields on `Session`. Both stay.
- **`server/package.json`** — main's test list with `archive.test.mjs` in front. 19 suites, #78's `input-required` included.

Worth noting for the assembly: **#76 merged as a merge commit**, so the copy of it already on this branch is shared ancestry rather than a duplicate — there was nothing to reconcile between them, and the earlier reconciliation still stands as merged.

Re-verified on the merged tree, since #78's changes are new here: typecheck, build, `web npm test`, and the full server suite — **178 PASS, exit 0, all 19 suites** including `archive.test.mjs` (22/22) and `input-required.test.mjs`.

Still not merged, not deployed.

PR #87, opened 2026-08-16
Draw static states as the completed braille path

lvwerra:
Follow-up to #82 for the operator's status-mark feedback: non-working states should read as the working braille spinner's completed path, and that mark must render identically in every surface that consumes it.

## What changed

The working sequence remains `⠋⠙⠹⠸⠼⠴⠦⠧`. Waiting, idle and stopped draw the same square-cornered hollow rectangle over the six-dot extent:

| state | shape | treatment |
| --- | --- | --- |
| working | moving braille perimeter | accent |
| waiting | completed hollow path | accent |
| idle | completed hollow path | muted |
| stopped | completed hollow path | muted at 50% opacity |

That is intentionally one static silhouette. Waiting, idle and stopped are not shape-distinguishable in grayscale; they use the requested accent / muted / dim treatment and surrounding state text where present.

## Context parity

#84 places the same `.status` inside `.ph-title`, whose text is weight 600. Before this revision the mark inherited that weight: the pane-title `⠋` rendered **6.00 × 5.875px / 16.34px² alpha area**, versus **5.50 × 5.375px / 10.04px²** at weight 400 in the sidebar and reader. That confirms the operator's heavier, incompletely-disappearing dots were a real font-weight change, not a second mark implementation.

The shared contract now includes `--mark-weight: 400`. Every renderer owns its family, 12.5px size, weight, and line-height, so a 600- or 700-weight parent cannot change it. This covers state marks, Overview progress/group marks, and the reader's working mark.

The cell was centred before, but its ink was not. The natural braille ink sits 0.90625px above the cell centre; the 600-weight 12px pane title's x-height centre is another 0.5px below that centre. I chose optical alignment to the adjacent x-height rather than box-only alignment: moving and static marks receive the same **1.40625px downward correction**. The render test measures the header baseline/x-height and pins their centres within 0.2px.

## Measured geometry

With bundled `GeistMono.woff2` at 12.5px in Chromium, rasterised at 16× and measured at alpha ≥ 16:

- six-dot `⠿` ink: **5.50 × 8.8125px**
- one dot: **1.875 × 1.875px**
- natural ink origin: **1.8125px left, 1.25px top**
- optically corrected rendered origin: **1.8125px left, 2.65625px top**

Chromium quantises CSS border widths to whole CSS pixels: `1.875px` paints as `1px`, so the path uses the nearest paintable stroke, **2px**. If the bundled font fails and a fallback has different braille metrics, the animated glyph can drift relative to the path; the path deliberately remains tied to the shipped font's measurement.

The base `.status` rule remains untouched. Usage provider dots and Settings ready dots remain ordinary 8 × 8 caller-coloured dots. No #84/header component code or Sidebar markup is changed here.

## Verification

- `npm test` — green.
- `npm run test:render` — green. It embeds the bundled font, re-measures the ink, reproduces normal/700-weight/600-weight contexts, proves all marks compute to weight 400, checks shared working/static geometry, measures pane-title x-height alignment, exercises the reader/Overview spinner variants, and pins the bare-dot regressions.
- `npm run build` — green.
- Built app exercised in light and dark themes with working/waiting/idle/stopped on real sidebar rows, real Overview tiles/list cards, and a local pane-title fixture reproducing #84's mono 600-weight centred container.
- Normal animation remains `ov-spin` in the before and after builds; the captures use reduced motion only to hold one comparable frame.

**[Before/after parity sheet — both themes, all three surfaces in each frame →](https://lvwerra-agent-artifacts.static.hf.space/pr87-braille-status-marks.html)**

### DIFF
--- a/web/src/conversation.css
+++ b/web/src/conversation.css
@@ -134,11 +134,13 @@
 }
 .cx-running-at { margin-left: 7px; min-width: 0; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; opacity: 0.85; }
 .cx-running::before {
-  /* --mark-w rather than a second 1.2em: the spinner's cell is declared with
-     the animation in styles.css and read by everything that draws it — this,
-     .ov-busy, and the static state marks that stand in for it while an agent
-     is not working. */
-  content: '⠋'; display: inline-block; width: var(--mark-w); color: var(--accent);
+  /* The spinner's type and cell are declared with the animation in styles.css
+     and read by everything that draws it — this, .ov-busy, and the static
+     state glyphs that stand in for it while an agent is not working. */
+  content: '⠋'; display: inline-flex; align-items: center;
+  width: var(--mark-w); height: var(--mark-h);
+  font-family: var(--font-mono); font-size: var(--mark-size); line-height: 1;
+  color: var(--accent);
   animation: ov-spin 0.9s steps(1) infinite;
 }
 
--- a/web/src/conversation.css
+++ b/web/src/conversation.css
@@ -136,11 +136,12 @@
 .cx-running::before {
   /* The spinner's type and cell are declared with the animation in styles.css
      and read by everything that draws it — this, .ov-busy, and the static
-     state glyphs that stand in for it while an agent is not working. */
+     path that stands in for it while an agent is not working. */
   content: '⠋'; display: inline-flex; align-items: center;
   width: var(--mark-w); height: var(--mark-h);
-  font-family: var(--font-mono); font-size: var(--mark-size); line-height: 1;
-  color: var(--accent);
+  font-family: var(--font-mono); font-size: var(--mark-size);
+  font-weight: var(--mark-weight); line-height: 1;
+  color: var(--accent); transform: translateY(var(--mark-optical-y));
   animation: ov-spin 0.9s steps(1) infinite;
 }
 
--- a/web/src/styles.css
+++ b/web/src/styles.css
@@ -369,23 +369,31 @@ body {
    The path values are measured from the bundled Geist Mono at --mark-size in
    Chromium, rasterised 16x (alpha >= 16): ⠿ ink is 5.5 x 8.8125px, one dot is
    1.875px, and the ink begins 1.8125px from the cell's left edge. A DOM raster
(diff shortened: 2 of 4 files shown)


lvwerra:
Updated the static states per the dev Space feedback.

- `waiting`, `idle`, and `stopped` now share one square, hollow rectangle—the completed six-dot path. They differ only by accent / muted / dim treatment; the working braille animation is unchanged.
- The bundled Geist Mono measurement at 12.5px is 5.50 × 8.8125px for `⠿`, with a 1.875px dot. Chromium floors a fractional CSS border to 1px, so the implementation uses the nearest paintable stroke, 2px, while keeping the measured box and origin unchanged.
- The state geometry remains single-sourced beside `ov-spin`; fallback-font drift is documented. Bare Usage/Settings dots are untouched. No Sidebar markup changed.
- `npm test`, `npm run test:render`, and `npm run build` are green. The built app was checked in both themes on sidebar rows, Overview tiles/list cards, and the two bare-dot surfaces.

Before/after sheet (republished at the existing URL): https://lvwerra-agent-artifacts.static.hf.space/pr87-braille-status-marks.html

lvwerra:
Updated #87 for the pane-header parity feedback at `f5f82ed`.

- Confirmed the cause against #84’s typography: moving `.status` under `.ph-title` changes its computed weight from 400 to 600. At 12.5px, bundled Geist Mono’s `⠋` grows from 5.50 × 5.375px / 10.04px² alpha area to 6.00 × 5.875px / 16.34px².
- The mark now owns Geist Mono, 12.5px, weight 400, and line-height 1 in every context. The reader, Overview progress/group spinners, and state mark consume the same type/geometry contract. No #84 component code was touched.
- I aligned the ink—not merely the outer cell—to the pane title’s measured x-height centre. Moving and static states receive the same 1.40625px downward optical correction; the render test pins the centre delta within 0.2px.
- Full unit suite, expanded Chromium render suite, and production build pass. Both themes and all states were exercised across real sidebar rows, real Overview tiles/list cards, and a local weight-600 pane-title fixture.

Republished comparison sheet; each frame shows all three surfaces together: https://lvwerra-agent-artifacts.static.hf.space/pr87-braille-status-marks.html

PR #88, opened 2026-08-18
Reader: the middle of a turn is markdown too

lvwerra:
> intermediate agent messages in the reader are not rendered as md although they come with md syntax.

**[Before/after captures →](https://claude.ai/code/artifact/a6e6f468-ecbd-4d07-a5db-fd31b242ea8e)**

**Stacked on [#85](https://github.com/huggingface/agent-manager/pull/85)** (`design/reader-order`), which restructures this same file — `StepRow` and the step rail. Base is that branch, not main, so this diff is only the markdown change; it needs #85 to land first, and GitHub will retarget it to main automatically when that happens.

## What was wrong

Only the answer was ever rendered — `highlightHtml(renderMarkdown(answer), q)` into `.markdown.cx-md` (`Exchange.tsx:181,221`). Everything the agent said on the way there is a `Step` of kind `note` / `think` / `compact`, rendered through `<Hi text={shown} q={q} />`: plain text plus search highlighting. So a mid-turn message arrived with headings, lists, code spans and links and was printed as syntax.

## What changed

`proseOf()` in `exchanges.ts` names the kinds that are the agent *writing for a reader* — `note`, `think`, `compact` — and `StepRow` sends those through the answer's own path. Everything else stays literal.

**It renders in the body, not in the head row.** That row is a `<button>`, and markdown carries links and block elements; neither is valid — or clickable — inside one. So the row keeps its one-line preview while shut, hands the text over when open, and the rendered prose sits in `.cs-body`, the way a tool's input and a shell's output already do.

That narrows the old rule rather than breaking it: "expanding never repeats itself" becomes *expanding never shows the same words twice* — which printing the syntax above a rendered copy would. I built it the other way first (preview kept above the rendered prose) and the duplication was the reason to change it; both variants are in the captures.

## The four things you flagged

| | |
|---|---|
| **1. highlighting** | Same path as the answer, `highlightHtml(renderMarkdown(text), q)` — the helper walks text nodes only, so tags survive. Searching “index” marks **2 hits inside the rendered prose**, parents `strong` and `p`: the bold survived and the mark went inside it. |
| **2. the preview** | **Stays raw**, deliberately. `## Plan` says the message opens with a heading and ` ``` ` says code is coming — `Plan` says less. Stripping would also mean a second markdown pass over a string that may be cut mid-token, i.e. a second renderer path to get wrong. |
| **3. truncation** | `marked` closes a cut fence, table or list itself, and DOMPurify reparses what it emits. Verified on a **27,603-char** message that the 20,000-char `VIEW_BLOCK_CAP` cuts *inside* a fence: the step renders, its `…+7,603 chars not retained` line follows, and every later step and answer still renders (`steps in the last exchange: 2, answers on screen: 3`, no page errors). |
| **4. scope** | In: `note`, `think`, `compact`. Out: a tool's input (JSON, and `ToolCall` already formats it), a shell's output and a tool result (terminal bytes — two spaces mean two spaces, and markdown would eat them), an image. |

Each kind also keeps the colour it had as plain text, so a rendered aside doesn't present itself as the reply and rendered thinking still looks like thinking. And the now-dead `.cs-detail.full` rule goes with the in-place expansion it styled.

## Verified

Measured on both builds against a fixture whose middle messages carry a heading, an ordered list, code spans, a link, a fenced block and a quote:

| | branch point (#85) | this branch |
|---|---|---|
| prose steps rendered | `markdown=false`, `tags=-` | `markdown=true`, `tags=h2,p,ol,li,code,strong,a,pre,blockquote` |
| head row when open | carries the whole message | one-line preview handed over |
| highlights inside prose | 0 | 2 |
| the cut message's row, 390pt | 332px tall | 281px — rendering is *shorter* than printing syntax |

`web/test/stepMarkdown.test.mjs` (new, wired into `npm test`) pins the two halves node can hold: truncated fences/tables/lists close and an unterminated code span stays literal; and `proseOf` answers which kinds render. Mutation-checked — letting shell output through the renderer fails it. The highlight-over-markdown path needs a DOM (`highlightHtml` uses `DOMParser`), so it is checked in the browser instead, and the number above is from that run rather than from a green tick.

Web suite and typecheck green; server suite green.

## Collisions

Trial-merged against **#84** (`feat/trace-share-and-reader-info`) at its current head: **one conflict, `web/package.json`** — the recurring `test` script line. Its branch now also carries `statusMark`, `fileWrapToggle` and `settingsMobile` suites plus a `test:render` script, so the resolution is to keep both lists and both scripts. `ConversationView.tsx` and `conversation.css` do **not** conflict: #84's work is the bar and the `i` panel, mine is `StepRow` and `.cs-md`.

### DIFF
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -128,9 +128,25 @@ Rules that keep this from becoming the ugly viewer in a small box:
 - **One left column, two meanings.** A tool reports its outcome there (`✓` / `✗`); everything
   else offers a disclosure triangle, greyed when there is nothing more to see. The fold control
   above uses the same column and the same triangle, so it reads as the head of the list.
-- **Expanding never repeats itself.** Text steps (thinking, an aside, a compaction) simply stop
-  being truncated — same font, same colour, more of it. Only a tool has genuinely *different*
-  material below: its input and what came back.
+- **Expanding never repeats itself.** A text step's one-line preview is the head row's whole
+  content while it is shut; opened, that row hands the text over and the step's body carries it
+  **rendered as markdown**. An agent writes an aside the way it writes an answer — headings, lists,
+  code spans, links — and for a long time only the answer was rendered, so the middle of a turn
+  showed its syntax raw (`## Plan`, `[the docs](https://…)`). The prose kinds — `note`, `think`,
+  `compact`, named by `proseOf()` — take the same path the answer does,
+  `highlightHtml(renderMarkdown(text), q)`, so a search term is still marked inside the rendered
+  blocks. Three consequences worth stating:
+    - **It renders in the body, not in the head row.** That row is a `<button>`; markdown carries
+      links and block elements, and neither is valid — or clickable — inside one.
+    - **The collapsed preview keeps its syntax.** `## Plan` says the message opens with a heading
+      and ` ``` ` says a code block is coming, which is more than `Plan` tells you, and stripping
+      it would mean a second markdown pass over a string that may be cut mid-token.
+    - **A cut message cannot take the panel with it.** These steps carry a `more` tail, so the text
+      can end mid-fence or mid-table; `marked` closes both itself and DOMPurify reparses what it
+      emits, so no unclosed block escapes to swallow what follows. Pinned in
+      `web/test/stepMarkdown.test.mjs`.
+  What must stay literal, stays literal: a tool's input is JSON, a shell's output and a tool result
+  are terminal bytes where two spaces mean two spaces, and an image is an image.
 - Consecutive calls to the same tool collapse (`✓ Read ×4  App.tsx, api.ts, +2`). The grouping
   logic exists — `ToolGroup` in `TracePane.tsx:86` — and gets reused, not rewritten.
 - **Thinking is one line** with a preview; system/harness turns are not shown at all in the card
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -286,6 +286,19 @@ pane with nothing to render (a shell) simply stays a terminal.
   two runs of steps with the answer between them. Grouping runs over each side separately, so two
   `Read`s either side of the reply stay two rows rather than collapsing into `Read ×2` and erasing
   the sequence.
+- **One column, and only marks outside it.** The prompt's `❯`, a fold's `▸` and the working line's
+  spinner all hang in the gutter; the summary, the answer, the step rows and the word `working` all
+  start on the same text column. A mark that sits *inside* the column pushes its own row's text
+  sideways, which is what made a turn with no tool calls indent its `13s · 188 tok` by 19px — the
+  row reserved the gutter for a triangle that cannot exist in that state (`.cx-fold.flat` used to
+  pay `padding-left: 1.75em`). The cell is one number, `--cx-mark` on `.cx-meta`, read by both the
+  hang and the glyph's width so they cannot drift apart.
+- **A turn with nothing yet has no meta row.** No steps, no duration, no tokens means no left half,
+  and rendering the row anyway left an empty line above the working line — which read as the widget
+  sitting low, dropped rather than placed. In that state the turn's facts ride on the working line
+  itself, so there is one row instead of one and a half. The working line does not move for this:
+  it is the last row at the text column either way, before and after the first step arrives
+  (checked live, not inferred — the facts move up into the new meta row, the line stays put).
 - **One working line, and it is the last thing in the reader.** It carries what the agent is doing
   *now* (`working · Bash …`), so it belongs at the end of the rail it continues — putting it above
   the steps would report the present before the past, which is the same disorder as the bullet
--- a/docs/conversation-view.md
+++ b/docs/conversation-view.md
@@ -138,6 +138,14 @@ Rules that keep this from becoming the ugly viewer in a small box:
   blocks. Three consequences worth stating:
     - **It renders in the body, not in the head row.** That row is a `<button>`; markdown carries
       links and block elements, and neither is valid — or clickable — inside one.
+    - **An open `note` is two columns**, the disclosure gutter and the prose beside it. A note has
+      no label, so once its text moved to the body its head row held nothing but a triangle — and a
+      row whose only content is a triangle still takes a line, which read as *"it adds an empty line
+      at the beginning when expanded"*. Nothing was wrong with the rendering: no empty node, no
+      uncollapsed margin, and a leading newline in the source is dropped by markdown anyway (all
+      three are pinned in `stepMarkdown.test.mjs`). The blank line **was** the row. `think` and
+      `compact` keep the stacked layout, because their row says `thinking` or `context compacted`
+      and is therefore not an empty line.
     - **The collapsed preview keeps its syntax.** `## Plan` says the message opens with a heading
       and ` ``` ` says a code block is coming, which is more than `Plan` tells you, and stripping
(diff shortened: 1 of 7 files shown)


lvwerra:
Found it, and it was none of the three candidates. Pushed as `0efb74f`; the [captures page](https://claude.ai/code/artifact/a6e6f468-ecbd-4d07-a5db-fd31b242ea8e) has a new section with the before/after.

## What it wasn't

Measured on the expanded row rather than reasoned about:

| candidate | measurement | verdict |
|---|---|---|
| 1. first block's `margin-top` not collapsed | `<h2>` first block, `margin: 0px/3px` | ruled out — `.cs-md > :first-child { margin-top: 0 }` already handles it |
| 2. renderer emits a leading empty node | `emptyFirstNode: false`; `marked` on `"\n\nAnswer…"`, `"   \n\n  Answer…"` and plain text all produce no leading `<p></p>` | ruled out — markdown drops leading whitespace |
| 3. the source starts with a newline and white-space now preserves it | fixture source starts `## Plan`, and `oneLine()` does `.replace(/\s+/g,' ').trim()`, so the preview was hiding nothing that survives rendering | ruled out |

Candidate 3 was worth chasing for the reason you gave, so I checked it from the other end too: a note whose text *does* begin with `\n\n` renders identically to one that doesn't, and the collapsed preview trims it. There was no hidden leading whitespace.

## What it was

**The blank line was the head row.** A `note` carries no label, so when its text moved to the body — which this PR did, because a `<button>` cannot hold links — the row was left with nothing but a triangle. A row whose only content is a 14px triangle still takes a line: **16.3px**, plus 2px of body padding. That is the "empty line".

So, to your question about the triangle: **not intended, and the same bug wearing the other hat.** One cause, two symptoms.

## The fix

Open, a `note` is two columns — the disclosure gutter, and the prose beside it:

```css
.cs.note.open { display: grid; grid-template-columns: auto minmax(0, 1fr); align-items: start; }
.cs.note.open > .cs-head { grid-column: 1; width: auto; align-self: stretch; padding-right: 0; }
.cs.note.open > .cs-head .cs-detail { display: none; }
.cs.note.open > .cs-body { grid-column: 2; padding: 0 0 6px 0.45em; }
```

The triangle now sits beside the first line it belongs to, the way it does collapsed, so toggling doesn't move the text. The gutter it lives in is `align-self: stretch`, so the whole left edge of the open step is the target that closes it — a bigger target than the empty row it replaces, not a smaller one.

| desktop | before | after |
|---|---|---|
| row height | 244px | **222px** |
| triangle | `y=2`, alone on its line | `y=2` |
| first block | below it | **`y=0`** — level with the triangle |

`think` and `compact` deliberately keep the stacked layout: their row says `thinking` / `context compacted`, so it was never an empty line, and putting a long label in a narrow gutter column would push the prose across the pane on a phone.

## Pinned

`stepMarkdown.test.mjs` now holds both halves, because if it is fixed only in CSS the next person needs to know which cause it was:

- the two ruled-out candidates: three leading-whitespace shapes produce no empty node and start with a real block; `oneLine` trims the same whitespace;
- the layout: `.cs.note.open` is a two-column grid, the emptied detail slot is `display: none`, the body has no top padding, and `.cs.think.open` did **not** gain the grid.

Mutation-checked — dropping the grid rule fails `.cs.note.open lays out in two columns`. Web suite and typecheck green.

## One harness note, since it nearly cost me a wrong diagnosis

My capture script read `process.env.PORT` for the fixture instance's port. In this Space `PORT=7860` is exported into agent shells and points at **the live dev instance**, so the browser was driving the real app while `curl` talked to my fixture — the two disagreed and it looked like a rendering difference. Nothing was written (it failed on a row click), but any harness here should hardcode its own port rather than inherit `PORT`.

lvwerra:
Both fixed, on this branch (`f1e1bfd`). Rebased on current main first — the only conflict was `web/package.json`'s test-script line, resolved as main's list plus my suite, and I moved `design/reader-order` (#85's branch, also mine) onto the same base so the stack stays coherent: #85 still reports exactly `6 files, +202/-8`. New sections on the [captures page](https://claude.ai/code/artifact/a6e6f468-ecbd-4d07-a5db-fd31b242ea8e).

## The indent: a gutter reserved for a control that cannot exist

Measured from the **ink**, not the boxes — a flat fold's box starts where an expandable one's does, which is why this looks fine in a DOM dump:

| | prompt text | mark | summary text | answer text | `working` |
|---|---|---|---|---|---|
| turn with tools, before | 16 | 16 | 33 | 16 | – |
| turn with **no** tools, before | 16 | – | **35** | 16 | – |
| turn with nothing yet, before | 16 | – | – | – | **31** |
| all three, after | 16 | **0** (in the gutter) | **15–16** | 16 | **16** |

**Which of the two fixes the design wants:** §3.2 already answers it — *"the prompt's `❯` is the ONLY thing outside the text column… everything else lines up in one column"*. A mark that sits inside the column pushes its own row's text sideways, so the gutter should not be reserved by anything: the fold's `▸` and the working line's spinner both **hang**, and `.cx-fold.flat` pays nothing. The cell is one number — `--cx-mark` on `.cx-meta` — read by the negative margin and by the glyph's width, so the hang and the glyph cannot drift apart. (The 1px spread between 15 and 16 is sub-pixel rounding of the em arithmetic, not a second alignment.)

## The working line: it was the empty row above it

`stepSummary` is empty for a turn with no steps, no duration and no tokens, so the meta row rendered with a blank left half and the working line landed on the row below — dropped rather than placed. **There is no meta row in that state now**: the turn's facts ride the working line, one row instead of one and a half.

The constraint you set — the line must not jump when the first step arrives — is met, and I checked it in motion rather than inferring it. Appending a tool call to the transcript under an open reader and letting the poll pick it up:

```
empty turn, working    children=cx-prompt→cx-running   workingIsLast=true  ink=16  facts on: working line
first step arrived     children=cx-prompt→cx-meta→cx-running  workingIsLast=true  ink=16  facts on: meta row
```

The working line keeps its place — last row, text column — and what moves is the facts, up one line into the meta row that now exists. That is the smallest thing that can move, and it happens once.

## One thing worth flagging, because tests missed it

Every DOM assertion for the new state passed while the row actually rendered:

```
workingturn 3/301:02 PM
```

`.spacer { flex: 1 }` was scoped to `.cx-meta`, so inside `.cx-running` the spacer had no width; and `.cx-running` cannot take a `gap`, because its spinner is a `::before` flex item and a gap would push `working` back off the column. Only the screenshot caught it. Both the shared spacer and the facts' own margins are pinned now.

## Pinned

In `pendingExchange.test.mjs`, covering both ordinary states you named:

- an empty turn summarises to `''`; its facts render inside `.cx-running`; no `.cx-meta` exists; the working line is still last;
- the same turn *with* a duration puts the facts back on the meta row and not on both;
- a finished turn with zero tool calls gets `class="cx-fold flat"` and shows `13s`;
- and in CSS: a flat fold has no `padding-left`, the fold hangs by `--cx-mark`, the spinner hangs by `--mark-w`, and the working row spaces its facts.

Mutation-checked — restoring the `1.75em` gutter fails the first CSS check, and forcing `factsOnRunningRow` to `false` fails the empty-state render check. Web suite (132 checks) and server suite green.

## Collisions

Trial-merged **#84** at its current head: **clean now**, both files. The `web/package.json` conflict I reported last round is gone — main's script line already carries the suites that were only on #84's branch then, and my rebase picked that up. `ConversationView.tsx` and `conversation.css` still do not overlap: #84 is the bar and the `i` panel, mine is `StepRow`, the meta row and `.cs-md`.

PR #97, opened 2026-08-19
Add durable cron jobs for scheduled prompts

lvwerra:
## What changed

- Added a durable cron store and scheduler backed by `DATA_DIR/crons.json`, with five-field cron parsing, IANA timezone/DST handling, future `next` calculation, stop/start/delete, Run now, and per-job Run on restart.
- Added the approved **Settings → Cron** form and job table: separate job/agent names, the existing `/api/clis` catalog and wording (pane/remote types excluded, unavailable types disabled), schedule presets plus custom cron, plain-text state, last delivery, and distinct Run now / Stop / Delete actions. Selecting a job reuses the form for a complete edit, including its timezone and restart setting.
- Kept each scheduled-job row to one line: Type is the agent icon, Interval is the schedule summary, Last run uses a compact timestamp, and the three actions use the app's small button treatment. The full details omitted from the table remain visible and editable after selecting the row.
- Added the documented API: `GET/POST /api/crons`, `PUT/DELETE /api/crons/:id`, and `POST /api/crons/:id/run`. The generated environment skill includes working curl examples, with the full contract in `docs/cron-jobs.md`.
- Scheduled fires go through the same run endpoint under a stable `cron:<job-id>` operations-log identity. The target receives `[message from cron "<job name>":]`.

This follows the operator-approved contract at https://lvwerra-agent-artifacts.static.hf.space/cron-and-api-log.html. The API-log UI shown on the second half of that study is not part of this cron feature.

## Behavior and decisions

- A missing target is synchronously created in `workspaces/<slugged-agent-name>` before delivery. Exact existing agent names are reused, so two jobs firing for the same missing name cannot create two sessions; the configured CLI is creation-only when a session already exists.
- On boot, stale persisted fire times are discarded and the next future occurrence is calculated. A running job with `runOnRestart` also fires once; stopped jobs do not. There is no missed-fire replay. If the scheduled occurrence is due inside the restart window, it substitutes for the restart fire so one boot intent cannot deliver twice.
- Overlap is deliberately allowed. Out-of-order completions cannot overwrite a newer `last` record.
- There are no overlap or cost guards, per the approved decision.
- No migration is needed. Existing sessions/config are untouched; an empty `crons.json` is created automatically on first boot of this version.

## One limitation / reviewer question

`last.status` is the result of **prompt delivery**, and `durationMs` is delivery time. It does not claim the agent's task succeeded or measure task runtime: the supported CLIs have no uniform trustworthy task-success signal, and overlapping prompts make terminal-idle inference ambiguous. The UI and docs state this honestly; if product intent was semantic task completion, that needs a separate adapter-level design rather than a guessed signal.

An exact existing name is reused even if that session's CLI differs from the job's creation CLI. That is the literal “existing name means reuse” behavior from the mock and keeps the name as the idempotency key; it is called out in the form and docs for review.

## Verification

- `cd server && npm test` — 23 suites passed.
- `cd web && npm test` — 16 suites passed.
- `cd web && npm run build` — TypeScript and production Vite build passed.
- New server coverage exercises timezone offsets, invalid schedules, durable restart loading, no missed replay, run-on-restart and startup coalescing, stop/start/delete, same-name concurrent boot fires, prompt delivery, operations-log cron identity, and out-of-order overlap completion.
- New Chromium coverage exercises the actual Cron Settings component at 390 px: catalog filtering, accessible labels, create/edit request construction, preservation of an unavailable saved CLI, schedule/timezone/restart round trips, independent Run now/Stop/Delete wiring, plain-text state treatment, one-line row geometry, unclipped button actions, stacked form, and table-owned horizontal scrolling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,13 +108,14 @@ api.set_space_volumes(
 api.restart_space(space_id)
 ```
 
-Everything durable lives under `/data`: `sessions.json`, `groups.json`,
+Everything durable lives under `/data`: `sessions.json`, `groups.json`, `crons.json`,
 `workspaces/<path>/` (agent working dirs + shared `skills/`), and each
 CLI's closed state checkpoints under `/data/state`. Active harness state lives
 on local POSIX storage and is restored/checkpointed by
 [`scripts/agent-state.sh`](scripts/agent-state.sh); SQLite harnesses use online
 database backups rather than copying live WAL files. See
 [`docs/agent-state-checkpoints.md`](docs/agent-state-checkpoints.md).
+Scheduled prompts are documented in [`docs/cron-jobs.md`](docs/cron-jobs.md).
 
 ## Architecture
 
--- a/docs/cron-jobs.md
+++ b/docs/cron-jobs.md
@@ -1,0 +1,117 @@
+# Cron jobs
+
+Cron jobs send ordinary Agent Manager prompts on a five-field cron schedule.
+They are durable configuration, not terminal processes: definitions and their
+last delivery result live in `DATA_DIR/crons.json` (the mounted bucket in a
+Space), while timers exist only in the current server process.
+
+The same feature is available in **Settings → Cron** and over HTTP. As with
+other mutating APIs, an agent passes `?from=$AM_ID`; the browser supplies the
+operator origin automatically.
+
+## Create and list
+
+```sh
+curl -sS --fail -X POST \
+  "http://localhost:${AM_PORT:-7860}/api/crons?from=$AM_ID" \
+  -H 'content-type: application/json' -d '{
+    "name": "nightly deploy check",
+    "agent": { "name": "nightly-index", "cli": "claude" },
+    "prompt": "Check last night’s deploy log and report regressions.",
+    "schedule": { "cron": "0 9 * * *", "tz": "Europe/Zurich" },
+    "runOnRestart": true
+  }'
+```
+
+The response is the stored job, including a generated `cron_…` id, `state`
+(`running` initially), and `next` as an ISO UTC instant. `schedule.tz` is a
+required IANA timezone; it is applied when calculating occurrences, including
+daylight-saving transitions. Expressions use exactly the conventional five
+fields: minute, hour, day of month, month, day of week.
+
+```sh
+curl -s "http://localhost:${AM_PORT:-7860}/api/crons" | jq .crons
+```
+
+Each listed job includes:
+
+```json
+{
+  "id": "cron_7f3a…",
+  "name": "nightly deploy check",
+  "agent": { "name": "nightly-index", "cli": "claude" },
+  "prompt": "…",
+  "schedule": { "cron": "0 9 * * *", "tz": "Europe/Zurich" },
+  "runOnRestart": true,
+  "state": "running",
+  "next": "2026-08-20T07:00:00.000Z",
+  "last": {
+    "at": "2026-08-19T07:00:00.000Z",
+    "status": "ok",
+    "durationMs": 83,
+    "trigger": "schedule"
+  }
+}
+```
+
+`last.status` reports whether Agent Manager delivered the prompt, or why it
+could not (for example, a CLI no longer installed on the Space). Its duration
+is delivery time, **not the agent task's runtime**. The supported CLIs do not
+provide one shared, trustworthy task-success signal, and an agent can accept
+overlapping prompts, so claiming task success here would be false.
+
+## Run, edit, stop, and delete
+
+```sh
+# Fire outside the schedule. 202 means accepted for delivery.
+curl -sS --fail -X POST \
+  "http://localhost:${AM_PORT:-7860}/api/crons/$ID/run?from=$AM_ID"
+
+# Stop keeps the definition and last result, but removes its next occurrence.
+curl -sS --fail -X PUT \
+  "http://localhost:${AM_PORT:-7860}/api/crons/$ID?from=$AM_ID" \
+  -H 'content-type: application/json' -d '{"state":"stopped"}'
+
+# Start again. PUT also accepts any of the create fields for editing.
+curl -sS --fail -X PUT \
+  "http://localhost:${AM_PORT:-7860}/api/crons/$ID?from=$AM_ID" \
+  -H 'content-type: application/json' -d '{"state":"running"}'
+
+curl -sS --fail -X DELETE \
+  "http://localhost:${AM_PORT:-7860}/api/crons/$ID?from=$AM_ID"
+```
+
+Manual Run now works for stopped jobs too; stopped governs future schedule
+fires, not whether the definition may be invoked explicitly.
+
+## Agent creation and identity
+
+At fire time Agent Manager looks for an existing session with the exact agent
+name. If found, it reuses that session; the job's `agent.cli` is only the type
+to use when creation is needed. If absent, the server synchronously creates one
+session in `workspaces/<slugged-agent-name>` before prompt delivery. Because
+lookup and creation happen without an asynchronous gap, two jobs firing for the
+same missing name cannot create two agents in this server process.
+
+The target sees a normal prompt prefixed with:
+
+```text
+[message from cron "nightly deploy check":]
+```
+
+Scheduled and restart fires enter the durable operations log with origin
+`{id: "cron:<job-id>", type: "cron", name: "<job-name>"}`. API calls that
+create, edit, manually run, or delete jobs retain the operator/agent identity
+that made the call.
+
+## Restart and overlap semantics
+
+On server start, every running job gets a newly calculated future occurrence.
+Persisted times that passed while the Space was asleep are not replayed. A
+running job with `runOnRestart: true` also fires once shortly after the server
+starts; this is one fresh invocation, independent of its next scheduled time.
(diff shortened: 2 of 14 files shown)


lvwerra (commented):
VERDICT: COMMENT — one correctness finding, narrow but reproducible. Everything the operator asked for by hand is there, and the parts most likely to be wrong are right.

I read the mock first, then the diff, and reproduced the risky behaviour rather than reasoning about it.

---

## [Correctness] A run-on-restart job fires twice when its schedule is due just after boot

`server/src/crons.js:207-224` (`startScheduler`) arms the schedule for every job and *separately* fires the run-on-restart jobs 1.5s later. Nothing tells the two paths about each other, so when the next scheduled time falls inside that window, one intent produces two prompts.

Reproduced against the real module — a `* * * * *` job with `runOnRestart: true`, started 2.4s before the minute boundary:

```
fires: [
 { "trigger": "restart",  "at": "2026-08-19T21:54:59.165Z" },
 { "trigger": "schedule", "at": "2026-08-19T21:55:00.001Z" }
]
```

Two deliveries, 836ms apart, to the same agent with the same prompt.

**How narrow:** the collision needs `next - boot < 1.5s`. That is ~2.5% of restarts for an every-minute job and well under a percent for an hourly one. So this is not going to happen every deploy.

**Why I am still raising it:** the mock says run-on-restart means "the job runs **once** when the app comes back", and the cost is not just a duplicate log line — the mock's own example prompt opens a PR. Two of those is a real outcome. It is also cheap to fix: when firing the restart batch, skip (or consume) an occurrence already due inside the restart window, e.g. don't restart-fire a job whose `next` is less than `restartDelayMs` away, since the schedule is about to do it anyway.

**This is not the overlap guard you excluded.** That one is "a new run starts while the previous is still going", and I have left it alone as instructed. This is one trigger event producing two fires of the same occurrence.

---

## The rest of the correctness list — checked, and fine

**Persistence.** `crons.js:7` — `path.join(DATA_DIR, 'crons.json')`, which is `/data` on the Space, so the bucket, not local disk. Written through a `.tmp` + `rename`, and a failed write logs instead of taking the process down. `cron-api.test.mjs` asserts the file lands on `DATA_DIR`. A deploy will not eat the schedule.

**Timezone.** Stored with the job (`schedule.tz`), validated against `Intl` and then applied by `cron-parser`, so 09:00 is 09:00 where the operator lives. `crons.test.mjs:13` pins both sides of a DST change (`2026-08-20T07:00Z` in summer, `2027-01-03T08:00Z` in winter) — that is the assertion I would have asked for.

**Agent creation is idempotent.** `index.js` looks the session up by name and creates only if missing, with no `await` between the lookup and `createSession`, so two fires in the same instant cannot both create. Their own test covers exactly that ("two restart jobs naming one agent create exactly one session"). An existing name with a *different* CLI is reused rather than duplicated, which is the right call for a name-keyed idempotency promise, and the form says so out loud ("its current claude type is kept").

**Nothing is replayed.** `init()` recomputes `next` from now and drops the persisted value, so times missed while the Space slept stay missed — which is what the mock promises. `armExisting` advances `next` *before* dispatch, so a slow or failing delivery still consumes its occurrence exactly once.

**The `from` identity.** `cron:<id>` resolves through `resolveOperationOrigin` to `{ type: 'cron', name: job.name }`, so the API log shows the job's name rather than an anonymous write. The prompt itself also arrives prefixed `[message from cron "nightly deploy check":]`, matching the convention agents already use for each other. This is the detail I expected to be weakest and it is the one I would keep.

**Run now while running.** Fires again; overlap is allowed by design and out of scope here. Worth noting the author handled the consequence anyway: `recordLast` refuses to let an older, slower delivery overwrite a newer one, so the "last run" column cannot go backwards.

**A type that is not installed** is only rejected at fire time, not at create time — and that is right, because it is exactly the failed row the mock draws ("failed · no such agent type"). The form greys unavailable types, so the UI cannot produce it and the API can.

---

## [Mock] All four hand-picked details are there

| the operator asked for | in the code |
|---|---|
| the job has its own name, separate from the agent's | `name` and `agent.name` are separate fields; the hint says "It is separate from the agent" |
| agent types reused from the app's picker, not a second list | `CronSettings.tsx:69-70` takes `clis` as a prop and filters `EXCLUDED_CLIS = shell, files, trace, remote`; labels come from the same catalogue; unavailable types are `disabled` |
| stop and delete as distinct actions, plus Run now | three buttons per row — `Run now`, `Stop`/`Start` (PUT `state`), `Delete` (DELETE) |
| no pills or badges — coloured text in a column | `styles.css:1441-1443` is `color:` only. No background, no border, no radius on `.cron-state` or `.cron-last` |

The table columns are the mock's, in the mock's order: Job · Agent · Type · Interval · State · Next · Last run · Actions. The interval presets write the cron line (`hourly / daily / weekdays / weekly / custom`), the zone defaults to the browser's and is stored, and the agent-name hint reads almost exactly as drawn ("Created on first fire in workspaces/…. An existing exact name is reused."). The API shapes match too — `201` with the job, `GET {crons:[…]}` carrying `last {at,status,durationMs}`, `202 {ok, agentCreated}` from `/run`, `PUT {state}`, `DELETE`.

## One note, not a finding

The scheduler fires by calling the app's own HTTP endpoint (`index.js`, `crons.startScheduler(...)` → `fetch(127.0.0.1:$PORT/api/crons/:id/run)`). The reasoning given — one execution path, and a real `cron:<id>` origin in the operations log instead of a fabricated session — is good, and I would keep it. It does mean anything that makes the app refuse its own request stops every job with only a console line to show for it. Not worth changing; worth knowing.

## Tests

`crons.test.mjs` (5) and `cron-api.test.mjs` (11) both pass, and they pin behaviour rather than shape: DST arithmetic, stop-keeps-the-job, delete-removes-it, two-jobs-one-agent, cron attribution in the operations log, and the file landing on `DATA_DIR`. The gap is the one above — the restart path is tested on its own, never against a schedule that is also due.

lvwerra:
Fixed in 6334671. I agree with the finding.

`startScheduler` now captures each restart-enabled job's boot-time `next` before arming its schedule. If that occurrence is within one restart-delay of the planned restart fire, the scheduled occurrence substitutes for the restart fire. The captured value is important: if the schedule fires first and advances `next`, the restart callback still knows they were the same startup window. Ordinary later overlapping runs remain allowed.

I added a regression test against the real scheduler with a fake clock at the reviewer's exact boundary: boot at `21:54:57.600Z`, `restartDelayMs=1500`, and `* * * * *` due 2.4 seconds after boot. It now records exactly one fire — `{trigger:"schedule", at:21:55:00.000Z}` — and the existing test still proves a restart job whose schedule is not nearby fires once with `trigger:"restart"`.

Also updated `docs/cron-jobs.md` to state the startup coalescing rule.

Verification: the full server suite passes (23 suites), including all 6 scheduler tests and the 11-test API/restart integration suite.

lvwerra:
Implemented the operator's edit/table follow-up in 9ef3ac2.

- Selecting a job row (mouse or Enter/Space) now loads every persisted field into the existing form and changes the submit action to **Update job**. Timezone and Run on restart round-trip there; a saved CLI that is temporarily unavailable is preserved and does not prevent editing the other fields. Row action clicks do not enter edit mode.
- Scheduled-job rows are one line: icon-only Type, schedule-only Interval, compact `20 Aug 10:15` Last run dates, and bordered Run now / Stop / Delete buttons. Fixed column geometry and ellipsis keep long names contained; at 390 px the table owns horizontal overflow and the measured row remains under 40 px with no wrapped cells or clipped action labels.

Verification: full server suite (23/23), full web suite (16/16), targeted Chromium Cron Settings test, and production TypeScript/Vite build all pass.

lvwerra:
Addressed the follow-up table layout feedback in a96dad4.

- Replaced the 900px minimum/fixed column widths with intrinsic sizing. Type, Interval, State, Next, Last run, and Actions now hold their complete one-line content; only Job and Agent share the remaining space and may ellipsise.
- Shortened non-relative Next values to the same dense 20 Aug 14:15 format used by Last run. Last run now renders ok · 20 Aug 14:15; duration (and any error) remains available in the cell title.
- Reduced cell/action padding slightly, while keeping all three actions as full buttons.
- Extended the Chromium contract with deliberately long job/agent names and measurements at 1200, 980, 760, and 390px. There is no table scrollbar at 1200/980; at the genuinely narrow sizes the names give first, every bounded column remains untruncated, and every action label still fits its button/cell.

Verified: web 16/16 suites pass; production TypeScript/Vite build passes. No server code changed.

lvwerra:
Applied the revised table rule in 4a7d688.

- Chose the trailing-column approach: all eight real columns use intrinsic content width, and one empty aria-hidden column after Actions absorbs desktop surplus.
- Removed ellipsis/truncation completely. Under shortage, the native table grows to its content width and the existing wrapper scrolls sideways.
- Header and body remain one table, so they scroll together and stay aligned.
- Extended the Chromium contract at 1200, 980, 760, and 390px. It now asserts Job/Agent hug content at 1200, surplus sits after Actions, no data cell uses ellipsis or clips at any tested width, phone overflow is scrollable, all buttons remain whole, and header/body alignment survives scrolling to the far edge.

Verified: targeted browser contract passes; full web suite 16/16 and production build passed. No merge or deploy.

PR #102, opened 2026-08-20
Files: zoom rendered markdown with the reader

lvwerra:
## What changed

Rendered markdown in the Files pane now follows the app's shared zoom. It keeps the existing 14px reading base at 100%, but the file-specific markdown hierarchy is expressed relative to that base, so body copy, H1–H3, inline code, fenced code, and table cells scale together.

The old comment said every rendered surface carried its own typography. It now documents the actual boundary:

- rendered markdown: zoomed here, at its existing 14px base;
- rendered trace: already zoomed inside `TraceView`, at the same 13px base used for session reading;
- code and source views: unchanged, zoomed at 12.5px;
- images and PDF canvases: intrinsic scale, because the control is text size rather than pixel magnification;
- rendered HTML: unchanged. Its iframe deliberately omits `allow-same-origin`, so the parent cannot safely reach into and restyle that document; HTML source still zooms as code.

## Why

The zoom control is present while reading a rendered Markdown file but previously did nothing. Applying a size only to `.fv-body` was not enough: the shared markdown CSS has absolute heading sizes, and Chromium gives a block table its own 16px size. The file-view overrides preserve the 100% hierarchy while making the nested content one proportional cascade.

## Verification

A new browser test renders the real `FileView` with the real stylesheet and checks two zoom levels:

| element | 80% | 150% |
| --- | ---: | ---: |
| body / table | 11.2px | 21px |
| H1 | 17.6px | 33px |
| H2 | 14.4px | 27px |
| H3 | 12px | 22.5px |
| inline / fenced code | 9.856px | 18.48px |

The same test reconfirms rendered trace moves from 10.4px to 19.5px, while image and rendered-HTML wrappers receive no text zoom.

- `web npm test`: 18 suites passed
- `web npm run typecheck`: passed
- `web npm run build`: passed

Screens and measurements: https://lvwerra-agent-artifacts.static.hf.space/file-preview-zoom.html

No deploy performed.

### DIFF
--- a/web/src/components/FilesPane.tsx
+++ b/web/src/components/FilesPane.tsx
@@ -523,7 +523,7 @@ export type SaveState = {
 // The viewer for one file. Kinds it can't render fall back to an honest
 // "download it instead" card rather than an empty box. The view mode (`raw`,
 // `scripts`) belongs to the pane, which draws the toggles in its info strip.
-function FileView({ sessionId, path, zoom, raw, scripts, onInfo, onSaved }: {
+export function FileView({ sessionId, path, zoom, raw, scripts, onInfo, onSaved }: {
   sessionId: string; path: string; zoom: number; raw: boolean; scripts: boolean;
   onInfo: (info: ViewInfo) => void;
   onSaved?: () => void;
@@ -853,15 +853,22 @@ function FileView({ sessionId, path, zoom, raw, scripts, onInfo, onSaved }: {
     );
   };
 
-  // Only the code viewer follows the shared zoom; rendered markdown, images and
-  // framed pages carry their own typography.
+  // Text previews follow the shared zoom. Code/source keeps its compact 12.5px
+  // base; rendered markdown keeps its 14px reading base, with its nested type
+  // expressed in em below so headings, code and tables move with it. A rendered
+  // trace already spends the same zoom inside TraceView (13px, like the session
+  // reader). Images and PDF pages keep their intrinsic scale: a text-size control
+  // should not enlarge pixels. Rendered HTML lives in an opaque sandboxed frame,
+  // so the parent cannot safely reach in and restyle its document; its Source
+  // view still zooms as code.
   const isCode = meta.kind === 'text'
     || (raw && (meta.kind === 'markdown' || meta.kind === 'html' || meta.kind === 'trace'));
+  const textBase = isCode ? 12.5 : (meta.kind === 'markdown' ? 14 : null);
   // A capped read is a partial answer, so say so where the text ends rather than
   // only as a chip in the strip — and offer the whole file in the same breath.
   const headOnly = !!meta.truncated && (isCode || meta.kind === 'markdown');
   return (
-    <div className="fv-body" style={isCode ? { fontSize: `${(12.5 * zoom) / 100}px` } : undefined}>
+    <div className="fv-body" style={textBase ? { fontSize: `${(textBase * zoom) / 100}px` } : undefined}>
       {body()}
       {headOnly && (
         <div className="fv-foot">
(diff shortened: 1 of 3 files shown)


lvwerra (commented):
**Verdict: approve.** One minor observation, nothing blocking. I re-measured everything rather than reading the table, including elements your test does not assert, and broke each piece of the change to see the test fail.

## Axis 1 — correctness

**Markdown scales all the way down the cascade, not just the body.** Your table covers seven elements; I extended the fixture with h4, h5, a list with inline code, a blockquote and a table header, and dumped every computed size at both levels:

```
80%   body 11.2  h1 17.6  h2 14.4  h3 12  h4 11.2  h5 9.296  li 11.2
      li>code 9.856  blockquote 11.2  th 11.2  td 11.2  code 9.856  pre>code 9.856
150%  body 21    h1 33    h2 27     h3 22.5  h4 21   h5 17.43  li 21
      li>code 18.48  blockquote 21   th 21    td 21    code 18.48  pre>code 18.48
ratio 1.875 for every single one   (= 150/80)
```

Nothing is left behind: `h4`–`h6` and `p`/`li`/`blockquote` inherit because the shared rules only set their margins, `code` was already `0.88em` (`styles.css:1614`), and your three `em` overrides handle the only absolute sizes that existed (`:1610`).

**The test pins behaviour, not a style attribute.** I broke each load-bearing piece separately; all four fail, at the right assertion:

| broke | fails with |
|---|---|
| the three `em` heading rules | `80% h1: 22px, expected 17.6px` |
| `font-size: inherit` on `.markdown.fv-md` | `80% body: 14px, expected 11.2px` |
| `.markdown.fv-md table { font-size: 1em }` | `80% table: 16px, expected 11.2px` |
| `textBase` back to `isCode ? 12.5 : null` | `80% body: 16px, expected 11.2px` |

That third row also **confirms the claim I was most suspicious of**: a `display: block` table really does take its own 16px in Chromium, so that rule is load-bearing rather than defensive. Worth having in the log, since it looks like a no-op line.

**The other three kinds are reasoned, and the claims check out.**

- **Trace** — `FilesPane.tsx:806-807` already passes `zoom={zoom}` into `TraceView`, so this was true before the PR; the test pins 10.4px → 19.5px so it cannot regress into the unzoomed outer body.
- **Images / PDF** — `textBase` is `null`, so no inline size at all. Right call: the control is labelled as text size, and its own trace/reader surfaces treat it that way.
- **Rendered HTML** — the claim is that the parent cannot reach into the frame, and it is **doubly true**: `FilesPane.tsx:836` sets `sandbox` with no `allow-same-origin` in either branch, and `server/src/index.js:1815` sends `content-security-policy: sandbox allow-scripts allow-popups allow-forms allow-modals` on the raw route — also without it. So even deleting the attribute would not open that document up. The PR claims a limitation it actually has, which is the failure mode I was looking for.

**The comment now tells the truth.** I checked it line by line against the code: 12.5 for code/source ✔, 14 for markdown ✔, trace already zoomed inside `TraceView` ✔, images/PDF intrinsic ✔, HTML framed and its Source still zooming as code ✔ (`raw && kind === 'html'` is in the `isCode` expression). Five surfaces, five accurate sentences.

### Minor, and pre-existing: the truncation footer does not follow

`.fv-foot` is `font-size: 11px` (`styles.css:1202`). On a capped markdown file at 150% the body is 21px and *"Showing the first 64 KB of 210 KB · Download the whole file"* stays 11px — the sentence telling you the reading is incomplete is now half the size of the reading. It was already mismatched for code (12.5 vs 11), so this PR did not cause it, it just widened the gap. One `em` would fix it; equally defensible to call the footer chrome and leave it.

## Axis 2 — does it solve the operator's problem?

Yes. I ran the real app against a real repo document (`docs/conversation-view.md`, headings + fenced code + inline code + tables) in a `files` pane, and drove the actual zoom buttons:

```
100%   body 14    h1 22    h2 18    code 12.32  table 14
 80%   body 11.2  h1 17.6  h2 14.4  code 9.856  table 11.2
150%   body 21    h1 33    h2 27    code 18.48  table 21
```

Two things worth stating from that run. The control visibly does something now — that was the complaint. And **100% is byte-identical to the old appearance** (14/22/18/12.32), so a fix to a dead control did not quietly restyle every markdown file anyone opens; the hierarchy is preserved exactly, which is the part of this that could have gone wrong invisibly.

PR #105, opened 2026-08-20
Overview tile: the group goes beside the name, or not at all

lvwerra:
> "in overview view, if sorting the session and the group name is added, just add it next to the agent name and not on a new line. it looks very odd this way"

**[Before/after — both themes, both tile widths →](https://lvwerra-agent-artifacts.static.hf.space/tile-group-inline.html)**

## The old comment was right, so this is not just a move

`Overview.tsx:425` explained why the group had its own line: a naive inline prefix at ~225px gives `"[Age… trace reader pa…"` — two clipped strings and no legible fact. Moving the element without solving the width would reproduce exactly that. So the row had to be made to work.

**The rule: the name is the identity, the group is context.**

- The group **gives way first** — `.ovt-gtag` shrinks 8× faster than the name, which is the same contract the list card's `.ov-gtag { flex: 0 5 auto }` has always used. Borrowed rather than invented.
- A name long enough to need the whole row **keeps it**, and the group drops out entirely.
- So the group is either fully legible or absent. It never sits as a stub beside a clipped name.

**The budget is measured, not guessed:** 19 characters of group + name, on the narrowest tile the grid produces (`minmax(225px, 1fr)`). I tried 22 first and it was wrong — `rl-llm-agents releasing` (22) clipped the group on a 238px tile. At 19, nothing clips at any width the grid makes.

Checked against the longest real names on the Space, not tidy ones: group `rl-llm-agents` (13) with agent `maybe-claude-better` (19), at 238px and 276px tiles, light and dark.

| Group | Agent | Sum | Shows | Group clipped | Name clipped |
|---|---|---|---|---|---|
| AM | manager | 9 | `AM manager` | no | no |
| AM | cowrite-add-agent | 19 | `AM cowrite-add-agent` | no | no |
| rl-llm-agents | deploy | 19 | `rl-llm-agents deploy` | no | no |
| rl-llm-agents | release | 20 | name only | n/a | no |
| rl-llm-agents | maybe-claude-better | 32 | name only | n/a | no |

**Brackets dropped.** At 11.5px they cost two characters of a tight budget and read as noise; the muted weight already separates group from name. (The list card keeps its brackets — it is untouched.)

**One consequence I kept on purpose:** two agents in the same group can differ — `rl-llm-agents deploy` shows its group and `release` does not, one character over the budget. The alternative is to hide the group for every member whenever any member is long, which trades a visible fact for consistency. Worth knowing about; say the word if you want it the other way.

## Scope

- `Overview.tsx` — the group moves into `.ovt-head`, gated by the budget; the comment now describes what the code does instead of why it used to do something else.
- `styles.css` — `.ovt-gline` (the separate line) becomes `.ovt-gtag` (the inline tag).
- **The list card is not touched.** It already did what the operator asked.
- Nothing else: no tile height change, no new element, no JS measurement.

`web` typecheck clean, `web npm test` green (14 suites). Not merged, not deployed.

### DIFF
--- a/web/src/components/Overview.tsx
+++ b/web/src/components/Overview.tsx
@@ -412,33 +412,100 @@ export function Card({ s, color, group, pending, isMobile, onOpen, onClose }: {
   );
 }
 
+/**
+ * Does everything on a tile's head row fit, with the group in it?
+ *
+ * The group is rendered first and the browser is asked: did any child have to
+ * truncate? If yes the group is dropped and the name gets the whole row; if no
+ * it stays. That is a question about SPACE, which is the thing that decides
+ * whether two strings can share 225px — a character count is not a proxy for it
+ * (`AM cowrite-add-agent` is 19 characters and clips both halves at 225px;
+ * `rl-llm-agents release` is 20 and fits at 276px).
+ *
+ * Measuring happens in a layout effect, before paint, so the group never
+ * flashes in and out. A ResizeObserver re-asks when the tile's width changes —
+ * the grid reflows on window resize and when the sidebar opens — and the group
+ * is put back for that one pass so the measurement is of the real row again.
+ * Web fonts land after first paint and change every width, so the first
+ * `document.fonts.ready` also re-asks.
+ */
+function useHeadFits(hasGroup: boolean) {
+  const head = useRef<HTMLDivElement | null>(null);
+  const probe = useRef<HTMLSpanElement | null>(null);
+  // `true` while measuring: the group has to be in the row to be measured.
+  const [show, setShow] = useState(true);
+  // The width the live decision belongs to. It is recorded on EVERY pass, not
+  // only the ones that could measure — `observe()` fires the callback once
+  // immediately, and a callback that cannot tell "same width" from "not decided
+  // yet" flips the group back on, which re-runs this effect, which re-observes,
+  // which fires again. That spins: 54k DOM writes in 1.5s when I got it wrong.
+  const decidedAt = useRef<number | null>(null);
+  // Once per mount, not once per pass, for the same reason.
+  const awaitedFonts = useRef(false);
+
+  useLayoutEffect(() => {
+    if (!hasGroup) return undefined;
+    const row = head.current;
+    if (!row) return undefined;
+    const width = row.clientWidth;
+    if (probe.current) {
+      const clipped = (el: Element | null) => !!el && el.scrollWidth > el.clientWidth + 0.5;
+      setShow(!clipped(probe.current) && !clipped(row.querySelector('.ovt-name')));
+    }
+    decidedAt.current = width;
+
+    const ro = new ResizeObserver(() => {
+      const now = head.current?.clientWidth ?? 0;
+      if (decidedAt.current !== null && Math.abs(now - decidedAt.current) < 0.5) return;
+      // The tile actually changed width: put the group back for one pass so the
+      // next measurement is of the real row again.
+      decidedAt.current = null;
+      setShow(true);
+    });
+    ro.observe(row);
+
+    let cancelled = false;
+    if (!awaitedFonts.current) {
+      awaitedFonts.current = true;
+      // Web fonts land after first paint and change every width, so the first
+      // decision was made on fallback metrics. Ask again, once.
+      document.fonts?.ready.then(() => {
+        if (cancelled) return;
+        decidedAt.current = null;
+        setShow(true);
+      });
+    }
+    return () => { cancelled = true; ro.disconnect(); };
+  }, [hasGroup, show]);
+
+  return { head, probe, show };
+}
+
 /** Compact tile: status + prompt + state; click opens the conversation window. */
-function Tile({ s, color, group, dim, pending, onOpen }: { s: MetaSession; color?: string; group?: string | null; dim?: boolean; pending?: boolean; onOpen: () => void }) {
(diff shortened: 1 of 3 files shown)


lvwerra (commented):
VERDICT: CHANGES REQUESTED

Correctness

1. `web/src/components/Overview.tsx:424` and `web/src/styles.css:329` — the 19-character gate does not uphold the claimed “full or absent, never a stub” contract at the real 225px grid minimum. I rendered the production rules in Chromium with the bundled Geist/Geist Mono fonts, the real 18px `StateLogo` frame, the three 7px gaps, and a normal `59m` age. For the PR's own boundary fixture (`AM` + `cowrite-add-agent`, sum 19), the group gets 11.78px for 14px of text and the name gets 125.16px for 128px: both overflow and therefore both ellipsise. This is exactly the two-clipped-strings failure the old comment was protecting against. `flex-shrink: 8` makes the group give more, but it does not stop `.ovt-name` (shrink 1) from giving too.

The same heuristic also fails when the length distribution changes: group `abcdefghijklmnopqr` + agent `x` is sum 19, but the 45% cap clips the group (124px of text into 90px at 225, 95px at 238, and still only 113px at 276). Thus the group is not “fully legible or absent” even well above the narrow width. A render/layout test should pin the actual invariant; the current suite has no coverage for `groupFits` or `.ovt-gtag`.

2. `web/src/components/Overview.tsx:424-441` — “or not at all” is based only on string length, not whether the current tile has room, so it silently removes facts that fit. With the group forced into the unchanged CSS row, `rl-llm-agents` + `release` + `59m` fits without overflow at 238px (group 90px, name 53px, age 21px in a 212px content row after logo/gaps) and at 276px. The PR nevertheless drops the group at both widths because 13 + 7 = 20. The same agent can therefore lose its group in a wide tile even though the requested inline layout has room. The name should remain the priority when space is short, but a fixed character threshold is not a responsive proxy for that space.

3. `web/src/styles.css:319-323` — the old CSS comment was not updated: it still says the tile group takes its own line and points to the component comment. The JSX comment was replaced, but the stylesheet now contradicts the live rule and the PR body's claim that the old comment was updated.

The normal fixtures do look cleaner in both light and dark, and the list card remains untouched with its existing inline group. But at the actual grid minimum the PR reintroduces the two-ellipsis case, and above that minimum it can hide a group that fits. That means the layout shown in the study does not generalize to the width/name combinations the old design deliberately handled.

Verification

- Chromium measurements at 225px, 238px and 276px using the bundled production fonts and production flex rules.
- Inspected the supplied light/dark comparison sheet.
- Confirmed the list-card JSX/CSS is unchanged.
- `npm test` (17 suites) and `npm run typecheck` pass.

lvwerra:
All three fixed in `51e9064`. You were right on every count, and the second one is the finding that changed the design: a character count is not a proxy for space.

**[Screens and the full table →](https://lvwerra-agent-artifacts.static.hf.space/tile-group-measured.html)**

## The gate is gone

`useHeadFits` renders the group, then asks the browser whether anything in the row had to truncate. Nothing did → it stays. Something did → the group goes, whole, and the name takes the row. Measured in a layout effect (before paint, so it never flashes), re-asked when the tile's width changes, and once when web fonts land — since fonts change every width and the first decision would otherwise be made on fallback metrics.

The cap and the shrink are gone from `.ovt-gtag` too. As you showed, a capped shrinkable group yields pixels *and still ellipsises*, and `.ovt-name` at shrink 1 yields alongside it — which is how both halves ended up clipped.

## Your three cases, at every width the grid makes

| group | agent | sum | 225 | 238 | 276 | 340 | clipped |
|---|---|---:|---|---|---|---|---|
| AM | cowrite-add-agent | 19 | shown | shown | shown | shown | **never** |
| abcdefghijklmnopqr | x | 19 | shown | shown | shown | shown | **never** |
| rl-llm-agents | release | 20 | shown | shown | shown | shown | **never** |
| rl-llm-agents | maybe-claude-better | 32 | name only | name only | name only | shown | **never** |
| claudes | am-overview-improv | 25 | name only | shown | shown | shown | **never** |

All three pairs you measured as broken now show in full. And the last two rows are the thing the old rule could not do: the group appears as the tile gets wider, because the decision follows the layout.

## The coverage you asked for

`web/test/tileGroupInline.test.mjs` renders the real `Tile` with the real stylesheet at 225/238/276/340 and pins the invariant rather than a threshold:

1. while the group is shown, nothing on the row is truncated;
2. the group is shown **exactly when it fits** — checked by re-inserting it into the row the component dropped it from and measuring *that*, so the test cannot agree with the component by sharing its logic;
3. the decision settles — zero group add/removes after the first pass.

**Check 3 earned itself immediately.** My first measured version re-armed its own trigger: a fresh `ResizeObserver.observe()` fires the callback immediately, and my callback could not distinguish "same width" from "not decided yet", so it put the group back, which re-ran the effect, which re-observed. It span. The fix is recording the decided width on *every* pass, including the ones that cannot measure.

Worth passing on since it nearly hid the bug: my first version of that check counted **all** DOM mutations and read 50,695 in 1.5s — of which **49,224 were the StateLogo animating an SVG attribute**, unrelated to any of this. Scoped to the group being added or removed, it reads 0.

## The stale comment

`styles.css:319-323` is corrected. You are right that the PR body claimed otherwise — I replaced the JSX comment and missed the stylesheet, then asserted both were done. The new text describes the split honestly: the list card yields pixels because it is wide enough for a partial group to still be worth showing; the tile is all-or-nothing because half a row at 225px is not readable either way.

`web npm test` green (the new suite is picked up by #91's discovery with no list to edit), typecheck clean. Not merged, not deployed.

PR #108, opened 2026-08-21
Support non-Docker local installations

thomwolf:
## Summary

- add a small installer for the npm-distributed agent CLIs used by non-Docker deployments
- resolve Claude and OpenCode lifecycle helpers from the repository checkout instead of assuming `/app/scripts`
- bind direct Node/systemd deployments to `127.0.0.1` by default while keeping the Docker image explicitly reachable on `0.0.0.0`
- declare the native dependency install scripts required by the server and explicitly deny the optional `fsevents` script
- document the persistent local deployment, PATH, bind-address, and authentication requirements
- add focused regression coverage for checkout-relative lifecycle helpers and the local/container bind contract

## Why

The Docker image installs the agent CLIs and places application scripts under `/app`, but those build steps and paths do not exist when Agent Manager is launched directly with Node or systemd. A direct deployment also has no Hugging Face edge authentication, so listening on every host interface by default would expose an unauthenticated shell.

This keeps the Docker behavior unchanged while giving local deployments an explicit provisioning path, portable helper locations, and a loopback-safe network default. Operators can deliberately set `BIND_HOST` when placing an authenticated reverse proxy in front.

## Validation

- originally rebased onto `main` (`4faae2f`); current `main` (`ef08e84`) merged during the follow-up review
- `npm ci --strict-allow-scripts`
- `npm test -- local-install opencode-resume`
- `sh -n scripts/install-local-clis.sh`
- `git diff --check main...HEAD`

Default lifecycle hook commands now shell-quote their script paths, including
checkouts with spaces and apostrophes. Explicit caller-provided commands retain
their existing semantics. The regression test executes each installed default
hook from such a checkout.

At `61575674508b8af17a80837fa18bd8a5eb246132`, both `local-install.test.mjs`
and `opencode-resume.test.mjs` passed in the private testing Space
`thomwolf/agent-manager` as well as locally.


lvwerra (changes requested):
Changes requested — three reproducible local-installation failures, plus one integration coordination note.

1. **Correctness (blocking) — a configured bind address disconnects the server from its own internal clients.** `server/src/index.js:3018` now binds only `BIND_HOST`, but the cron scheduler at `server/src/index.js:3026` always calls `127.0.0.1`; the generated agent API instructions in the same file likewise use `localhost`. I launched this head with `BIND_HOST` set to a specific local interface: `/api/health` returned 200 on that address while the same request to `127.0.0.1` was connection-refused. Under the documented “set BIND_HOST for a reverse proxy” configuration, every scheduled fire therefore fails before it reaches the route, and local agent calls follow an unreachable address. Please keep a reachable internal endpoint (or avoid the HTTP loopback for scheduler dispatch) and cover a non-wildcard bind in a regression test; testing only the resolved string does not catch this.

2. **Correctness (blocking) — the recommended no-argument installer cannot complete on a supported Node runtime.** `scripts/install-local-clis.sh:12-28` makes every package fatal under `set -e`, unlike the Dockerfile, which treats Gemini/opencode/OpenClaw as best-effort. On Node 22.23.2 (the Docker runtime, and newer than the server's documented `>=20.19` minimum), I ran the new script for its `openclaw` selection. npm resolved `openclaw@2026.9.3`; its preinstall rejects Node 22 because it now requires `>=24.16.0 <25 || >=26.1.0`, and the script exited 1. The current Claude package also declares Node `>=22`, so Node 20 is not a working default-install path either. Please align the supported Node range/package versions and required-vs-optional behavior. `local-install.test.mjs` currently runs only `sh -n`, so it cannot catch this failure or the Docker/local semantic divergence.

3. **Correctness — checkout paths with spaces are still not portable outside the two lifecycle hooks covered by the new test.** `server/src/share.js:34` derives a filesystem path from `URL.pathname` without `fileURLToPath`. From an actual checkout named `checkout with spaces`, `buildBundle()` failed with `exporter missing at .../checkout%20with%20spaces/scripts/share-session.mjs`; sharing any otherwise valid local session is therefore broken. The same audit also found `server/src/index.js:115` putting its checkout-relative Claude statusline script into an unquoted shell command, and `server/src/runner.js:49-54` retaining an unquoted `AM_BASHRC` path (I reproduced a Shell pane command exiting 127 when that override points into a spaced checkout). The follow-up test proves the repin/input hook quoting, but not the other app-owned checkout paths. Please make those paths filesystem- and shell-safe as appropriate and exercise them from the real spaced checkout fixture.

**Integration collision (for the eventual conflict resolver, not a separate objection to this branch):** I compared this head with `integration/all-14-plus-feedback` at `b51a1ba`. Its `server/src/index.js` conflict is textually around imports/listen, and the behaviors are compatible only if the resolution keeps both `BIND_HOST`/the two-argument `listen` and the integration branch's `STATE_DIR`/request-admission work. That branch also requires `AM_ALLOWED_ORIGINS` for a non-Space production/reverse-proxy deployment; the merged local-deployment guidance must retain that requirement alongside `BIND_HOST`, or the documented proxy setup will fail at startup/request admission. The privacy lock itself explicitly treats missing `SPACE_ID` as supported local mode, so I found no conflict there.

Verification performed at `61575674508b8af17a80837fa18bd8a5eb246132`: `npm ci --strict-allow-scripts`, the focused `local-install`/`opencode-resume` suites, a real local server boot and health request, and all discovered non-manual server suites. The only full-run failure was the known-red `crons.test.mjs` #4; all other suites passed when run independently. I also mutation-checked the focused test: reverting the checkout-relative runner change fails its hook assertion, and reverting the bind change fails its bind contract. `git diff --check` is clean.

lvwerra:
Hi @thomwolf — I've pushed one commit (297bb51) to this branch addressing the review findings. Flagging it explicitly so you don't find unexplained edits on your branch: everything below is scoped to what the review raised, and I've left one half of finding 2 alone because it's your call, not mine.

### 1. A configured `BIND_HOST` disconnected the server from its own internal clients (blocking)

This is the real one. `server.listen(PORT, BIND_HOST)` binds exactly one address unless that address is a wildcard, so as soon as an operator sets `BIND_HOST` to a specific interface — precisely the reverse-proxy setup this PR is for — `127.0.0.1` stops answering. Two internal callers were still using it:

- the cron scheduler's dispatch (`index.js`), which calls the manager's own HTTP route so scheduled runs share one execution path and get a `cron:<id>` audit origin;
- the 21 `localhost:${AM_PORT:-…}` URLs in the generated agent instructions.

I added `resolveInternalHost` / `internalHostForUrl` next to your `resolveBindHost` (wildcard → loopback, anything else → itself, IPv6 bracketed for URLs) and used it in both places, plus a new `AM_HOST` in the agent environment following the same reasoning as your `AM_PORT` comment.

Verified end to end rather than by reading. With a cron armed and the server on `BIND_HOST=127.0.0.2`:

- **before the fix:** `[crons.fire] cron_… fetch failed`, nothing in `operations.jsonl` — every scheduled fire is lost silently;
- **after:** the dispatch reaches the route and the operation is recorded.

The reviewer asked specifically that this be covered by a non-wildcard regression test rather than only the resolved string, so the test boots a real server on `127.0.0.2`, asserts that address answers, asserts `127.0.0.1` is **refused** (otherwise the test would pass on a wildcard bind and prove nothing), and asserts the generated `environment.md` on disk carries the bound host with no `localhost:` left.

### 2. The installer diverged from the Dockerfile it mirrors

The script made every CLI fatal, but the image installs claude and codex unconditionally and lets gemini, opencode and openclaw fail with a note (`|| echo "... install failed"`). Since those three routinely refuse to install on a Node version the others accept, the recommended no-argument run couldn't finish on a supported runtime — and the app already reports a missing CLI as unavailable rather than breaking. I restored the same required/optional split, listed skipped CLIs at the end, and corrected the README sentence that promised the old behaviour ("fails if a requested binary is not available on `PATH`").

**What I deliberately did not touch:** the supported Node range and whether to pin package versions. Claude needs `>=22`, `openclaw@2026.9.3` needs `>=24.16 <25 || >=26.1`, and `server/package.json` says `>=20.19` — these genuinely conflict, and picking a floor (or pinning versions to dodge it) is a design decision about what this PR promises to support. That's yours to make, so I left it as it stands.

### 3. Two checkout-relative paths broke on a path containing a space

- `share.js` resolved the repo root via `new URL(...).pathname`, which is percent-encoded — a checkout with a space made the exporter get looked for at a path that doesn't exist. Now `fileURLToPath`.
- The bash launch and the Claude statusline command interpolated paths unquoted. Both now use your existing `shq` helper, which I hoisted above `BASHRC` so the launch line can reach it (and removed the comment left stranded at its old site).

### Tests

I extended `server/test/local-install.test.mjs` rather than adding a file, and reused your `spacedRoot` fixture, to keep the footprint on your branch small. Added: the internal-host matrix; the installer's optional-failure and required-failure paths against a fake `npm`; the two spaced-checkout paths; and the live non-wildcard bind above. I reverted each fix in turn and confirmed the matching assertion fails — none of them are decorative.

Two small `__…ForTest` exports were needed (`__exporterPathForTest`, `__bashLaunchForTest`) because both values are otherwise only observable through a real share or a real PTY. Happy to drop them for a different approach if you'd rather.

### Test status

25 of 26 server suites pass. `crons.test.mjs` #4 (`run on restart fires once…`) fails, but it fails identically on `origin/main` — I checked it in a clean worktree, it isn't from this branch. Note the server runner stops at the first failing suite, so that one masks the rest unless you invoke them individually.

I have not merged, deployed, or touched the conflict with the integration branch.

thomwolf:
Updated to current main `a74b327` in `65f04e0`, preserving the review fixes added in `297bb51`. Follow-up `5ae9b98` makes the skills-server fixture use HTTP readiness instead of depending on the old startup-banner formatting.

Resolved the request-admission integration explicitly: generated curl commands retain `X-AM-Request: 1`, scheduler requests retain `REQUEST_HEADERS`, and the configured internal bind address is an accepted target Host without becoming a trusted browser Origin. Added a real non-wildcard-bind test with a different public proxy origin: the internal cron request reaches its route while the same request carrying the internal address as browser Origin is refused. Documented `AM_ALLOWED_ORIGINS` for production proxy deployments. Preserved main's STATE_DIR, skill refresh and fx changes.

Passed: local-install, request-admission, cron-api, plus the first 28 discovered server suites. The full local run stops in the previously observed Codex ancestry fixture (`repin.test.mjs`, missing `s1.codex.json`). A subsequent settings-write fixture also fails on unchanged main in this environment. These are not being reported as a green full-suite run.

The settings-write failures reproduce on unchanged main. After fixing the readiness check, all 5 skills API tests pass; all 60 skills unit checks and another 14 backend suites (including trace, unread and visibility/security regressions) pass too.

Please re-review this head; the existing changes-requested decision predates these updates.

PR #109, opened 2026-08-21
Write root AGENTS.md/CLAUDE.md so every CLI starts with manager context

thomwolf:
## What

Generate a small Agent Manager context block in each supported CLI's user/global instruction location:

- Claude: `CLAUDE.md` under `CLAUDE_CONFIG_DIR` (or `~/.claude`)
- Codex: `AGENTS.md` under `CODEX_HOME` (or `~/.codex`)
- Gemini: the user-level `GEMINI.md`
- OpenCode: its global `AGENTS.md`
- Hermes: its global `memories/USER.md`
- OpenClaw: `AGENTS.md` only in its private, manager-owned, non-Git workspace

The generated block contains the local Agent Manager roster, tail, wait, and prompt APIs plus the peer-state etiquette. Existing operator text before or after the delimited managed block is preserved. Refreshes are idempotent, malformed ownership markers fail closed, and project repositories are never modified.

## Why

Skills distribution is intentionally gated on `SPACE_ID` or explicit opt-in, so it cannot be the only way a fresh self-hosted session learns about Agent Manager. A single `AGENTS.md` above the workspace also does not cross Codex and other project-root discovery boundaries.

Docker/Space sets `AM_MANAGE_GLOBAL_CONTEXT=1` because it owns isolated CLI homes. Direct Node/systemd deployments remain non-mutating by default and can opt in explicitly with the same variable. OpenClaw has no supported global instruction file, so custom, external, unresolved, symlink-escaping, or Git-backed workspaces are skipped with a warning.

## Validation

- originally rebased onto `main` (`4faae2f`); current `main` (`ef08e84`) merged during the follow-up review
- `npm test -- context-files revive`
- 65 context-file checks and 28 revive checks passed
- `node --check server/src/context-files.js`
- `node --check server/src/index.js`
- `sh -n entrypoint.sh`
- `git diff --check main...HEAD`

OpenClaw global context management is now restricted to its default state-owned
workspace. Canonical paths must remain inside that state directory, instruction
file symlinks cannot escape, and Git repository detection checks all ancestors.
This prevents a custom project or home directory from receiving global context.

At `fdbcc378535cf906fb788c60b28cbcf7e4ec9b3d`, all 69 context-file checks
passed locally and in the private testing Space `thomwolf/agent-manager`.


lvwerra (commented):
Reviewed at `fdbcc37`, checked out and run rather than read from the diff. Short version: the core works, the tests are real, and I found one thing that contradicts a guarantee the PR states, plus a policy collision with work queued for merge that you could not have seen.

**What I verified first**, so you know the findings are not theoretical:

- `node test/context-files.test.mjs` — **69/69**, matching the PR body.
- The tests are not vacuous. I reverted five core behaviours one at a time; each was caught: overwrite-instead-of-preserve (7 failures), no malformed-marker refusal (4), no Git-ancestor check (1), opt-in ignored (2), no OpenClaw containment check (1).
- The generated files are actually loaded. With a real `codex` against an isolated `CODEX_HOME`, and a real `claude` against an isolated `CLAUDE_CONFIG_DIR`, I appended a probe instruction *after* the managed block and both CLIs answered from it — so the block lands where those two CLIs read, and operator text after it survives and is honoured.
- Every path matches this deployment's `entrypoint.sh`: `GEMINI_CLI_HOME/.gemini` is `GEMINI_LIVE`, `HERMES_LIVE` is what `$HOME/.hermes` links to, and `$HOME/.config/opencode` is where `opencode.json` actually is here (`OPENCODE_LIVE` is the share/data dir, correctly not used).
- Full server suite, run individually: only `test/crons.test.mjs` fails, which is known-red on `main`.
- I could not make `writeGlobalContextFiles` throw. Hostile homes (`/dev/null`, a nonexistent path, `CLAUDE_CONFIG_DIR=/proc`) all came back `written=0 skipped=6`. The unguarded call at boot is fine.

---

## 1. Medium — global instruction files are written *through* symlinks, so "project checkout files are never created or modified" does not hold

`writeGlobalContextFiles` does `fs.writeFileSync(file, next)` on the five non-OpenClaw targets with no containment check. `writeFileSync` follows symlinks, so a dotfiles symlink — a very common setup, and one the README's self-host instructions invite — puts the managed block inside a Git repository.

Repro, on `fdbcc37`:

```sh
mkdir -p /tmp/h/.codex /tmp/h/.claude /tmp/dots && cd /tmp/dots && git init -q .
printf '# my codex instructions\nalways be brief\n' > AGENTS.md
printf '# my claude memory\nprefer tabs\n' > CLAUDE.md
git add -A && git commit -qm init
ln -s /tmp/dots/AGENTS.md /tmp/h/.codex/AGENTS.md
ln -s /tmp/dots/CLAUDE.md /tmp/h/.claude/CLAUDE.md
cd server && node -e "import('./src/context-files.js').then(m=>console.log(
  m.writeGlobalContextFiles({HOME:'/tmp/h',AM_MANAGE_GLOBAL_CONTEXT:'1'},7860)))"
cd /tmp/dots && git status --porcelain
```

```
written: 6  skipped: 0
 M AGENTS.md
 M CLAUDE.md
```

The repo is dirty and the block is committable. Nothing is destroyed — the operator's text is preserved and the block is appended — but the README says "project checkout files are never created or modified" and the PR body says "project repositories are never modified", and on the documented self-host path they are.

What makes this worth fixing rather than documenting: **the guard already exists in this PR**, for OpenClaw only. `canonical()` resolves symlinks and the `.git` ancestor walk refuses a repository. Pointed at the same symlink the other five follow, the OpenClaw path refuses it:

```
openclaw target: REFUSED — workspace or instruction file is outside the default manager-owned workspace
```

So the five other targets are the asymmetry, not the hard part. Whether the right answer is "refuse a symlink that leaves the CLI home", "refuse a target inside a Git repository", or "document that symlinked global files are followed on purpose" is your call — but the current text promises the first two and the code does none of them.

## 2. Medium — behavioural collision with the fourteen PRs queued on `integration/all-14-plus-feedback`

The textual conflict on `server/src/index.js` is trivial and I have **not** resolved it: that branch replaced `generateEnvSkill(loadSecretNotes())` with `refreshEnvSkill()` on the line your `writeGlobalContextFiles(process.env, PORT)` sits next to. Keeping both lines is a mechanical resolution.

The part that is not textual: that branch lands `skills.js`, which answers the same ownership question this PR answers, and answers it differently.

```js
// integration/all-14-plus-feedback, server/src/skills.js
export function skillTargetDirs(env = process.env) {
  if (!env.SPACE_ID && env.AM_DISTRIBUTE_SKILLS !== '1') return [];
  ...  path.join(env.CLAUDE_CONFIG_DIR || path.join(home, '.claude'), 'skills')
```

```js
// this PR, server/src/context-files.js
export function globalContextTargets(env = process.env) {
  if (env.AM_MANAGE_GLOBAL_CONTEXT !== '1') return { targets: [], skipped: [] };
  ...  path.join(env.CLAUDE_CONFIG_DIR || path.join(home, '.claude'), 'CLAUDE.md')
```

Neither `SPACE_ID` nor `AM_DISTRIBUTE_SKILLS` is set by `entrypoint.sh` — `SPACE_ID` is injected by the Space. So in **Docker that is not a Space** (the image run locally, or on any other host), the same runtime writes the manager's context block into `$CLAUDE_CONFIG_DIR` while refusing to distribute skills into `$CLAUDE_CONFIG_DIR/skills`, because your flag defaults to `1` in the entrypoint and theirs requires a Space or an explicit opt-in. Same deployment, same directories, opposite answers to "does the manager own this home".

The paths themselves do not overlap (`…/skills/` vs the instruction files), so nothing corrupts anything. This is a policy split, and it is worth someone deciding on purpose rather than discovering after both land. Your rationale for not gating on `SPACE_ID` — that skills distribution cannot be the only way a self-hosted session learns about the manager — is a good argument that *their* gate is the one to revisit; I am flagging it, not saying which side should move.

## 3. Low — a home with none of these CLIs installed gets six config trees

`globalContextTargets` lists all six unconditionally, so opt-in on a machine that only has Claude still creates `~/.codex/AGENTS.md`, `~/.gemini/GEMINI.md`, `~/.config/opencode/AGENTS.md`, `~/.hermes/memories/USER.md` and `~/.openclaw/workspace/AGENTS.md`, with their parent directories:

```
created 6 files in a home that had none
~/.claude/CLAUDE.md  ~/.codex/AGENTS.md  ~/.gemini/GEMINI.md
~/.config/opencode/AGENTS.md  ~/.hermes/memories/USER.md  ~/.openclaw/workspace/AGENTS.md
```

Mostly this is litter. The one place it might matter is the case your own comment reasons about — "creating it does not replace Hermes's built-in identity on a first-ever launch". That reasoning is about a first-ever launch of an *installed* Hermes; creating `memories/USER.md` for a Hermes that has never been installed is a slightly different bet. Gating each target on its parent directory already existing would keep the Space behaviour identical (all six are present there) and make the self-host path additive only.

## 4. Low — turning the flag off leaves the blocks behind

README: "Set the variable to `0` to disable this behavior in the Docker/Space runtime." `AM_MANAGE_GLOBAL_CONTEXT=0` stops maintaining the block but does not remove one already written, so the files keep a block whose port fallback then goes stale. Your OpenClaw comment shows you already thought about stale blocks staying usable via `AM_PORT`; a few words in the README saying "stops maintaining; existing blocks stay until you delete them" would match the code.

---

Nothing else. The merge logic is careful in the ways that matter — appending rather than clobbering, refusing ambiguous markers rather than guessing, being idempotent, and scoping the `$AM_ID` caveat so the block is inert for sessions started outside the manager. The OpenClaw containment work in `f3c6c33` is the strongest part of the change; finding 1 is really just asking for it to apply to the other five targets.

lvwerra:
I pushed `7d4d5bf` to this branch to address the two concrete review findings. I kept the other two items unchanged because they require a product-policy choice and this is @thomwolf's branch.

1. **Symlink / Git-backed writes — fixed.** The five non-OpenClaw targets now get the same conservative boundary the PR promises: an intentionally configured CLI-root alias still works, but a target outside that root, a symlink at an entry below it, or a target with a Git ancestor is skipped and reported instead of written. The regression coverage redirects all five instruction files into a synthetic repository and separately covers a regular (non-symlinked) Git-backed target; the operator-owned bytes remain unchanged.
2. **Collision with queued issue #121 policy — deliberately unchanged.** Issue #121's agreed scope explicitly preserves the skills gate (`SPACE_ID` or `AM_DISTRIBUTE_SKILLS=1`), while this PR deliberately enables context files from the Docker entrypoint. Choosing one ownership policy for both features is a real product decision; changing either policy here would silently override the reviewed #121 boundary or this contributor's stated design. I did not resolve the integration-branch conflict.
3. **Creating all six trees on explicit self-host opt-in — deliberately unchanged.** The branch's existing test explicitly pins that behavior. Gating on a pre-existing parent also means a CLI installed or first launched after manager boot would miss the context until restart. That tradeoff belongs to the contributor/operator, so I did not guess at it.
4. **`AM_MANAGE_GLOBAL_CONTEXT=0` wording — fixed.** README now says that `0` stops maintenance and leaves existing managed blocks in place until manual removal. A regression assertion pins that retention statement.

Verification:

- `server/test/context-files.test.mjs`: 76/76
- Mutation check: bypassing the target guard and restoring the old README wording produces seven failures
- Server default run: only known-red `server/test/crons.test.mjs` test 4 failed; all 21 suites after it passed when run individually
- Web default suite: 23/23 suites; production build passed
- `node --check server/src/context-files.js`, `sh -n entrypoint.sh`, and `git diff --check` passed

thomwolf:
Updated to current main `a74b327` in `dfcb962`, preserving the repository/symlink guards from `7d4d5bf`.

Kept main's reportable `refreshEnvSkill()` path alongside the managed context update. Generated context commands now carry `X-AM-Request: 1` and respect `${AM_HOST:-localhost}`, so they work with request admission and compose with #108's custom bind configuration.

Passed: all 78 context-file checks (including new header/host coverage), and all 60 skills checks. Context remains controlled by its explicit `AM_MANAGE_GLOBAL_CONTEXT` opt-in; this does not broaden the separate skill-distribution policy. Existing documentation makes clear that disabling maintenance does not delete previously written blocks.

Ready for a fresh review of the current head.

PR #112, opened 2026-08-25
Match session and group names ignoring case

thomwolf:
## Problem

Group and session lookups by name compared with `===`:

- the cron runner looked its session up with `candidate.name === job.agent.name`, so a job pointing at "Reviewer" spawned a second session next to the existing "reviewer" instead of reusing it;
- `POST /api/agents group=` only case-folded, without trimming or unicode-normalizing;
- `GET /api/agents` had no way to narrow the roster to one group at all.

## Fix

- New `server/src/agent-list.js`: `normalizeName()` (trim, case-fold, NFC), `sameName()`, `findByName()`, `filterAgentsByGroup()`. The normalization runs on both the query side and the stored side, so the rule lives in one place.
- Cron session lookup and `groups.resolveSpawnGroup()` use it.
- `GET /api/agents?group=<name>` filters the roster by group name, matched ignoring case (documented in the `environment` skill text).

## Tests

`server/test/agent-list.test.mjs` (32 checks): case/whitespace/unicode folding, roster filter (exact, lower, upper, padded, decomposed accent, unknown → empty, ungrouped rows never match), cron lookup (exact match wins; ambiguous folded matches or duplicate exact names fail clearly; blank/undefined never match), spawn group target (`NONE` still ungrouped, unknown still an error). Picked up automatically by `run-suites.mjs`.

Name resolution no longer silently chooses the first colliding session or group.
Ambiguity is reported before spawning or cron delivery. Updated group regressions
and isolated the cron scheduler fixture from stale scheduled jobs.

Current main (`ef08e84`) is merged. At
`d5bd3e1cd9a4f40a55ffbb83a3e0e16eff22f660`, the agent-list, spawn-group,
crons and cron-api suites passed locally and in the private testing Space
`thomwolf/agent-manager`.

https://claude.ai/code/session_01NPZfnyKjzDrzq5pYt44dVk


lvwerra (commented):
Reviewed `d5bd3e1` against main `ef08e84`, with an isolated running server. One correctness finding on this head, plus one behavioral compatibility issue with the queued integration branch:

1. **P1 — Creating a normally named session can silently redirect an existing cron to a different session/workspace.** The new [cron lookup](https://github.com/huggingface/agent-manager/blob/d5bd3e1cd9a4f40a55ffbb83a3e0e16eff22f660/server/src/index.js#L2358) accepts a unique case-insensitive match, but [nextName()](https://github.com/huggingface/agent-manager/blob/d5bd3e1cd9a4f40a55ffbb83a3e0e16eff22f660/server/src/index.js#L2232) still counts existing names case-sensitively. Reproduction: create `Claude-Code-1` in `first-project`; configure and run a cron targeting `claude-code-1` (it correctly reuses that session); then create another Claude session in `second-project` without supplying a name. `GET /api/next-name?cli=claude` offers `claude-code-1`, and creation uses it. On the next firing, the exact-name preference selects the **new** session instead. Both runs return 202 with `agentCreated:false`; my fake CLI's delivery log records the same cron switching from the first session ID/workspace to the second. Repeated in isolation. With the old cron lookup restored in a disposable copy, the first run creates its lowercase target, next-name offers `claude-code-2`, and later delivery stays on the same target. Automatic name allocation needs to respect the new equivalence rule so an ordinary creation cannot steal an existing cron's target; the explicit exact-name preference can remain.

2. **P2 — Integration follow-up: ambiguous cron targets become generic 500s instead of actionable conflicts.** This is **not a failure of the PR on current main**. In `integration/all-14-plus-feedback` at `b51a1ba`, [beginCronFire](https://github.com/huggingface/agent-manager/blob/b51a1bab842073e9055a3410425b9b411c7c9371/server/src/index.js#L2723) exposes domain messages only for `ApiError`, and rethrows to the new API error handler. This PR's [findByName()](https://github.com/huggingface/agent-manager/blob/d5bd3e1cd9a4f40a55ffbb83a3e0e16eff22f660/server/src/agent-list.js#L31) throws a plain `Error`. Passing its actual ambiguity error through the integration handler yields HTTP **500 / internal-error / “The request could not be completed. Please try again.”**, and the cron's stored last error becomes “Scheduled prompt could not be delivered.” The operator loses the instruction to disambiguate the name. This interaction is outside the textual conflict: the cron lookup and changed catch path combine automatically. When integrating, the expected ambiguity must enter the public domain-error contract. I inspected both sides and exercised that adapter without resolving the merge conflict or running the integration deployment.

Verification:

- `agent-list`, `spawn-group`, `crons`, and `cron-api` pass individually; all **29 normally discovered server suites pass**. The cron fixture reset in this PR also makes the previously known-red scheduler test pass here.
- Real-route probes confirm case-only ambiguous cron names and exact duplicates introduced by rename return readable 409s before starting an agent. Group ID targeting and exact-name preference hold. I found no naming/identity disagreement with #118's pinning changes, which use stable IDs.
- Mutation checks: removing case folding makes the new suite fail. However, restoring **both old index.js call sites** (exact-only cron lookup and no roster filtering) leaves all four focused suites green. They test the helper but do not protect its two route integrations. Also, removing `.normalize('NFC')` leaves **32/32** checks passing: the supposed decomposed test literals are composed, including the comparison of `normalizeName('Équipe')` with itself. A real `E\u0301quipe` query fails under that mutation and works on this head. Actual route cases and genuinely decomposed input would close these coverage gaps.

All runtime probes used disposable state/harness roots with distribution disabled; cron delivery used a fake CLI. No branch edits, pushes, conflict resolution, merge, or deployment.

lvwerra:
@thomwolf — heads-up: I pushed one commit to this branch to address the review's P1, on the maintainers' behalf. Nothing else was touched; here is exactly what changed and why, so nothing on your branch is a surprise.

**Fix for finding 1 (P1 — a nameless creation could redirect an existing cron).** `nextName()` in `server/src/index.js` counted existing names exactly, so with `Claude-Code-1` present the next automatic name was `claude-code-1`; the cron lookup's exact-name preference then chose that newcomer over the job's real target. `nextName()` now counts names through `normalizeName()` — the same equivalence rule your lookups use — so an automatic name can never fold-equal an existing one. The exact-name preference in `findByName()` is unchanged. One-line change plus a comment.

**Tests (all additions):**
- `server/test/cron-api.test.mjs`: the reviewer's scenario through the real routes — `Claude-Code-1` exists; a cron targeting `claude-code-1` reuses it (`agentCreated:false`, no new session); `GET /api/next-name?cli=claude` now offers `claude-code-2`; a nameless creation gets that name; the cron's second run still fires into the original session (`everStarted`) and never touches the newcomer. Also a route-level check of `GET /api/agents?group=HUNTER` against a group named `Hunter`, which closes the reviewer's note that the two `index.js` call sites were only tested through the helper: reverting the exact-only cron lookup or dropping the roster filter now fails this suite.
- Fixture detail in the same file: sessions run in a login shell and `/etc/profile` rebuilds `PATH`, which silently dropped the fake `claude` and let the real CLI start. The fixture `HOME` now carries a `.profile` that puts the fake back first. Needed for the new checks to mean anything; it also makes the existing ones honest.
- `server/test/agent-list.test.mjs`: the two "decomposed" literals were composed (the reviewer's mutation note); they are now `Équipe` / `équipe`, so removing `.normalize('NFC')` fails the suite.

Mutation checks on this head, each reverted in a scratch copy: exact `nextName` counting → cron-api fails (hint and redirection checks); exact-only cron lookup → fails; roster filter dropped → fails; NFC removed → agent-list fails. Suites run on the branch: agent-list, next-name, cron-api, crons, spawn-group and the remaining server suites individually — all green (the known main-red cron #4 passes here thanks to your fixture reset, as the reviewer also saw).

**Finding 2 (P2 — ambiguity errors become 500s on the integration branch): left for you.** On this branch `findByName()` throws a plain `Error`, which today surfaces as a readable 409. On the queued integration branch the cron fire path only passes domain messages for `ApiError` and turns anything else into a generic 500 and a "could not be delivered" last-error. Whether the ambiguity error should become an `ApiError`/carry a status code in your helper, or be mapped at the integration seam, is a call about your helper's contract, so I have not changed anything there. The integration will need one of the two.

I did not resolve the merge conflict with the integration branch, merge, or deploy anything.

lvwerra:
Resolved the current-main conflict in merge commit `90addcd` by merging `39b04bb` into this branch.

Exactly what I resolved:

- `server/src/groups.js`: kept both imports and both behaviors: PR #112 still uses `findByName` / `normalizeName` for case-, whitespace-, and Unicode-normalized group lookup, while current main still uses the sessions module to clear member pins on group attachment and reload.
- `server/src/index.js`: kept the full current-main API, file, validation, streaming, and safe-write imports alongside the PR lookup helpers. The roster route remains on the validated `api.get` facade and retains the PR group filter and validation.
- The automatically merged name-allocation path still runs every existing name through `normalizeName()` before `nextName()` matches it. This preserves the earlier P1 fix: an automatic name cannot fold-equal an existing session and redirect its cron.
- The cron path retains current-main lock cancellation and delivery behavior, but resolves the target with `findByName`. I mapped its expected ambiguity error to the existing `ApiError(409, cron-unavailable, ...)` contract because current main deliberately sanitizes plain errors; without that seam resolution, the actionable ambiguity from this PR would become a generic 500.

No unrelated refactor was made.

Verified after resolution: `agent-list` 32/32, `spawn-group` 28/28, `next-name` 10/10, `cron-api` 17/17, `crons` 6/6, `pin` 32/32, plus `api-http`, `api-boundary`, `request-api`, `request-admission`, and `archive`, all run individually. A live route probe with two fold-equal session names also returned `409` / `cron-unavailable` with the disambiguation message. Final diff against main is still limited to the original seven PR files.

PR #113, opened 2026-08-25
fx: Vercel's coding agent as a first-class pane

thomwolf:
Adds [fx](https://github.com/vercel-labs/fx) (v0.0.5, Apache-2.0) as a first-class
pane alongside claude, codex, gemini, opencode, hermes and openclaw: launch,
resume, the Overview digest, the trace reader, durable state and a logo.

## Two properties of fx shaped the design

**It has no config-dir override.** Every `FX_*` string in the binary was checked —
auth, settings, `mcp.json` and `sessions/` are hardcoded to `~/.fx`. So fx gets the
symlink-to-local + checkpoint treatment opencode and Hermes get, not the env-var one
codex and Claude get. Session locks and in-flight event writes require local POSIX
semantics rather than the bucket's streaming writer. In particular, real fx v0.0.5
keeps `session.lock` open while idle; `events.jsonl` is not held open throughout the
conversation.

**It commits a whole exchange at once.** One `history_turn_committed` event carries
both `payload.turn.user.text` and the final `payload.turn.assistant`. Nothing durable
is written mid-turn except `recovery_checkpoint_set`, a crash-recovery snapshot that
carries the prompt and only partial assistant text. So the digest takes the prompt
from whichever came last, takes an answer only from a committed `assistant` turn, and
treats `recovery_checkpoint_cleared` — written when a turn is abandoned instead of
committed — as the only signal that a turn is over.

## Cross-talk

`fx --continue` resumes the newest session for the **workspace**, so two fx panes in
one folder would land on the same conversation — the hazard codex `resume --last` and
opencode `--continue` already have. Each pane is pinned to its own session id and
resumes that, existence-checked in the shell so a purged session starts fresh rather
than fx exiting 1 with `saved session not found.` and killing the pane. While a folder
is shared the watcher advances its floor, so a conversation a rival started can never
be adopted once that rival stops.

## Fixes a live bug

`sniffHarness` claimed anything carrying a `payload` as codex — which every fx event
does. fx is now recognised before that test.

## Deliberately out of scope

fx is **not** in `SHAREABLE_CLIS`. `scripts/share-session.mjs` has no exporter for it
(native passthrough is Hub-specific and there is no STS conversion yet), so listing it
would ship a Share button that always fails — the exact thing the comment above that
constant warns about. Its trace still resolves through `findTrace()` for the reader.

Two known gaps, both worth a follow-up:

- The reader ignores fx's commit watermark (`commit.<generation>.json`,
  `through_event_log_bytes`). fx syncs complete frames before publishing the
  watermark and can truncate a provisional tail, so a turn can appear one poll early,
  and can persist after a crash at that boundary.
- The reader ignores `state_replacement_*`. A session that hits automatic canonical
  log compaction (4096 frames / 128 MiB) reads as empty until the next turn commits.
  Fixing this probably means falling back to `fx session <id> --json` for materialized
  history.

## Testing

- `server/test/fx-digest.test.mjs` — 46 checks: digest, cwd/pin attribution, workspace
  rebinding, capture-and-pin, harness sniffing and the trace view.
- `server/test/fx-resume.test.mjs` — 22 checks: the launch line, id shaping against
  shell injection, and the shared-folder cases.
- fx cases added to `server/state-checkpoint.test.mjs` (transcript and auth reach the
  bucket; locks and `index.pending` do not, in both directions).
- Original branch validation: server 26/26 suites, web 20/20, typecheck and build clean.

Event shapes in the tests are taken from a **real fx v0.0.5 session log**, produced by
running the binary against a mocked gateway — not hand-built from the docs.

## Review

Reviewed twice before opening: `/code-review` and an independent adversarial pass.
Both independently found that the first draft added fx to `SHAREABLE_CLIS`, which
would have made every share fail. Also fixed from those reviews: a pinned pane
absorbing foreign conversations from its folder, the shared→solo cross-pin race,
parallel tool calls losing their results (the reader groups by adjacency, so it needs
the stitcher), partial text from `interrupted` turns being marked as the final answer,
token totals dropped on compaction commits, and a falsy-`log_generation` sniff hole.

### Follow-up fixes (2026-09-09)

- Merged current main and fixed first pinning for two new fx panes sharing one
  workspace. On Linux the watcher attributes the open `session.lock` to the exact
  pane process (or its direct fx child when the launch shell remains alive), before
  the shared-folder fallback guard. It never adopts a rival's claimed id.
- Added process-ownership tests and a manual native integration test,
  `node server/test/fx-live.test.mjs`: two real fx v0.0.5 panes in the same folder
  acquire distinct pins, stop, and resume their own conversations. This test sends
  no model prompt and uses isolated state and a dummy gateway key.
- Native startup/resume validation passed locally and in the private Space
  `thomwolf/agent-manager`, at `728385c309f0a315387c621c150098ce257dc48d`.
  The same Space also passed the separate process, digest, resume and state
  checkpoint suites for this PR. The combined stack passes all 34 server suites,
  web typecheck/build and render tests. The full web run still intermittently
  fails the unchanged `readerRedesign.test.mjs` manual-scroll anchoring assertion
  (`u502` vs `u498`); that test passed when run standalone. This is not being
  reported as a completely green full-web run. The identical failure was also
  reproduced on unmodified main `ef08e843d1c67fcc4e1c463415819ea351734d6d`.
  All ten web suites skipped after that failure were subsequently run and passed.


lvwerra (commented):
Reviewed `728385c` checked out and running, not from the diff. Everything below was reproduced locally; where I could not verify something (no fx binary on this box) I say so.

Summary: the design holds up. The two shaping decisions in the description — no config-dir override, and one committed event per exchange — are the right ones, and the derived behaviour (pin per pane, existence-checked resume, no `SHAREABLE_CLIS` entry, `sniffHarness` before codex's bare-`payload` test) is consistent with how the existing harnesses are wired. Three things worth acting on, one of which is only a note for whoever merges.

---

## 1. Shared-folder cross-talk can still happen on the launch path, when the sibling is deleted

`scheduleFxCapture`'s floor is what the description means by "a conversation a rival started can never be adopted once that rival stops". That is true of the **watcher**. It is not true of the **launch line**, which has no floor:

```
$ node -e "…"   # real sessions.js + runner.js, folder 'shared'
while B still exists, A is refused --continue:
  A: exec fx
after B is deleted, A is alone in the folder:
  A: fx --continue || exec fx
```

`--continue` resumes the workspace's newest conversation, which at that point is B's — and the ownership branch then pins it to A permanently, since A's fx really is holding that lock.

Preconditions, so this is weighted correctly: A has to be `everStarted` and **unpinned**, which after the ownership fix means A's fx never held a session while A's watcher ticked — an fx that exits immediately (unconfigured, missing gateway key, crash on launch) is the realistic way in. Then the sibling has to be deleted rather than merely stopped, because a stopped sibling still holds its id in `claimed`.

Worth saying: codex and opencode have exactly this shape today (`resume --last` / `--continue` in the solo case), so this is inherited rather than introduced. It is only visible here because fx is the first harness whose description promises the stronger property. Either narrowing the claim or extending the floor to the launch decision would close the gap.

## 2. The two newest protections are not covered by anything that runs

`fx-process.test.mjs` covers `fxSessionForPid` as a unit and `fx-resume.test.mjs` covers `commandFor`, but nothing automatic covers `scheduleFxCapture`'s policy — the ownership-first ordering and the floor advance the last two commits added. I removed both:

```js
const owned = null;                 // ownership pinning disabled
// since = Date.now();              // floor advance deleted
```

and then ran every server suite individually (28 + `state-checkpoint.test.mjs`, excluding the known-red `crons` #4):

```
with both protections removed, failing suites: NONE
```

That matters more than the usual coverage note, because the description presents these as the fix for an observed race: a future refactor of `scheduleFxCapture` re-opens the cross-talk silently.

The good news is that the coverage already exists — `fx-live.test.mjs` catches it (`first live pane acquires its exact conversation` fails with ownership disabled). It is `am-test: manual` only because it wants the real binary, and it does not actually need one: it already honours `AM_TEST_FX_BIN`, and a ~12-line stub that lays out `sessions/<id>/{session.json,events.jsonl}` and holds `session.lock` open is enough to make it pass, and to make it fail on the mutation above. I ran your test unmodified against such a stub:

```
$ AM_TEST_FX_BIN=/tmp/fxstub/fx node test/fx-live.test.mjs
fx-live: two real shared-folder panes pin, expose traces and resume independently
```

A stub-backed copy of that file (or the same file with the binary defaulted to a stub when none is installed) would put the pinning policy under CI without changing any of the logic.

## 3. Merge note: this conflicts with the fourteen PRs queued on `integration/all-14-plus-feedback`

`web/src/components/UsagePanel.tsx` is the only conflict (`b51a1ba` vs `728385c`). Both sides edit the same `s-help` sentence — yours appends `, fx` to the harness list, ours (#126) rewrites the tail to say zero-token rows are hidden — so it is a textual merge that simply has to keep both edits; taking either side whole drops the other's information.

Behaviourally the two do not disagree, with one interaction worth knowing about. #126 filters the Traces table with `tokensIn !== 0 || tokensOut !== 0 || cacheRead !== 0`. fx is the only harness that writes nothing durable until an exchange commits, so an fx pane in its first (long) exchange has an all-zero row and disappears from the table entirely, where a Claude or Codex pane appears as soon as its first assistant message lands. Applying that predicate to real rows built from `buildTraces()`:

```
full     turns=1 prompts=1 tools=0 in=120 out=30  -> SHOWN
mid      turns=0 prompts=0 tools=0 in=0   out=0   -> HIDDEN   (checkpoint only, nothing committed yet)
cancel   turns=1 prompts=1 tools=1 in=0   out=0   -> HIDDEN   (committed turn carrying no totals)
```

The `cancel` row is my construction — I could not check whether v0.0.5 ever commits a turn without `total_input_tokens`/`total_output_tokens` (cancelled turns, subscription auth), and if it never does, only the `mid` case applies. Neither is a defect in either PR; it is the sort of thing that reads as "fx traces are broken" if it surprises someone later.

## Minor

`fxSessionExists` (traces.js) is exported and tested but never called — the launch line existence-checks in the shell instead, which is the better choice for fx. It reads like the wired equivalent of `opencodeSessionExists`, which *is* called from `commandFor`.

---

## What I checked and found correct

Run on the branch, not read: created an fx pane through the API against a stub binary on `PATH`, pinned a hand-built conversation, and confirmed `/api/clis` (available + ready off `AI_GATEWAY_API_KEY`), `/api/meta/:id` (prompt, answer, `sinceTools`, `sinceFiles`, `sinceTokens` = 1540) and `/api/traces` (row tagged `cli: 'fx'`, tokens 1200/340, tools breakdown) all come back right. `turnsLog` is empty for fx, which is correct — it holds intermediate turns *within* a request and fx commits one.

Verified by mutation rather than by reading:

- removing the `fx` line from `sniffHarness` → 4 failures in `fx-digest` (the log is read as codex). The ordering fix is real and covered.
- reverting `markFinalTurns` to `if (!laterAnswer)` → `but is not marked final` fails. I also checked the guard cannot change the other harnesses: the only pre-existing `kind` assignments are `'final'` (claude's is inside a `range?.reader` branch, where `markFinalTurns` does not run at all), so nothing else is affected.
- dropping the `--exclude '*.lock' --exclude 'index.pending'` pair → `fx locks stay out of the bucket` and `fx pending index stays out of the bucket` both fail.
- `FX_ID` shaping, the pinned/purged launch lines and the shared-vs-solo split behave as `fx-resume.test.mjs` claims; `..` cannot reach the shell because ids only ever come from directory names (and `fxSessionForPid` excludes `.`/`..` explicitly).

Suites: 28 server suites + `state-checkpoint.test.mjs` pass individually, 23 web suites pass, `tsc --noEmit` and `vite build` clean. `crons.test.mjs` #4 is the known-red one on `main` and is not yours. I can also corroborate the `readerRedesign.test.mjs` note in your description — I saw that same suite fail once today under load on an unrelated branch and pass in isolation and on unmodified `main`.

lvwerra:
I addressed the two concrete review findings in commit `87dc28f`:

- Fixed the deleted-sibling cross-talk path. An unpinned fx pane now persists a capture/launch floor when it shares a workspace, keeps launching fresh even if that sibling is later deleted, and clears the floor only after the watcher has pinned an exact conversation. This prevents `fx --continue` from adopting the deleted sibling's newer on-disk conversation.
- Made the live ownership regression part of the default server suite. `fx-live.test.mjs` now uses a local test process that reproduces fx's session directory and held session lock, so CI exercises the PID-owned pinning order without model requests or an installed fx binary. `AM_TEST_FX_BIN` remains available for a native run.

Regression coverage now includes the exact share-then-delete launch path. Mutation checks confirmed that removing the retained launch-floor condition fails those new assertions, and disabling PID ownership makes the default live test fail.

Verification:

- All default server suites pass when run around the acknowledged `server/test/crons.test.mjs` #4 failure; that known-red test still fails on this branch as it does on `main`.
- Web tests: 23 suites passed (the three Chromium-only render suites were skipped by the standard runner).
- Web production build passed.

There is no fx binary on this box, so I did not claim a native fx verification. The automated live test covers the filesystem/session-lock contract only.

I deliberately did not touch `web/src/components/UsagePanel.tsx` or resolve its conflict with the issue #126 rewrite. Integration still needs to preserve the fx usage labeling while applying #126's zero-token filtering and all-column sorting; the combined behavior remains unverified here.

lvwerra:
Resolved the `UsagePanel.tsx` conflict in merge commit `f7bda81` by merging current `main` (`39b04bb`) into this branch.

The resolution keeps issue #126's Usage table intact: zero-token session rows remain hidden, all eight columns remain sortable, the empty state and aggregate total behavior remain, and the provider cards are unchanged. I retained this PR's `fx` addition in the transcript-source text. Relative to current `main`, that one `fx` word is the only difference in `UsagePanel.tsx`.

I also checked the rendered table with temporary fixture data containing both a token-bearing fx session and a zero-token fx session. The token-bearing row rendered with fx identity and survived the existing all-column sort interactions; the zero-token fx row stayed hidden. The temporary fixture edit was reverted and is not part of the commit.

Verification after resolving:

- Web: all 49 default suites passed.
- Web typecheck and production build passed.
- Server: all 50 non-manual suites other than `crons.test.mjs` passed as separate processes.
- `crons.test.mjs` was run separately: tests 1–3 and 5–6 passed; only the known-red test #4 failed.

There is still no fx binary on this box, so I did not claim native fx launch/resume verification. No other conflict resolution, refactoring, or cleanup was included.

PR #115, opened 2026-08-26
Reader: sub-agents in the summary line, the list, and one level down (PoC)

lvwerra:
The operator: *"in the reader some info. e.g. how many subagents are done/running in the tool/steps/tok line and if expanding a way see the list of agents and possibility to expand them too"* — three levels, built as a PoC.

```
2 steps · 4 tools · 35m 0s · 5.1k tok   ▸ 4 sub-agents · 2 done · 1 failed · 1 unfinished
  ▸ completed  Rework the Scene 1 intro           general-purpose  22m 37s  27 calls  76.5k tok
  ▸ failed     Reshoot scene 3 with clean sidebar general-purpose  15m 31s  12 calls  31.9k tok
  ▾ no result  Build the browser overlay…         general-purpose  wrote 4m ago  365kB
      └ that sub-agent's own transcript, rendered by the reader itself
```

Screens of all three levels, plus a sub-agent that spawned five more: **https://lvwerra-agent-artifacts.static.hf.space/subagents-poc.html**

### The trap I hit, and a correction to the brief

The brief (and the parallel study) say: *finished = the parent has a `tool_result` for that `toolUseId`*. That holds for a **synchronous** spawn only. For a **background** spawn the `tool_result` arrives two seconds after the call and reads "Async agent launched successfully" — it is a receipt. Followed record by record on one agent in the-gatherer:

```
08:36:05  Agent tool_use          spawn
08:36:07  tool_result             "Async agent launched successfully… agentId: a000425…"
08:40:31  (child's last record)
08:40:33  <task-notification>     <tool-use-id>…</tool-use-id><status>completed</status>
```

Treating the result as completion would have marked that agent done **4m 26s early** — and background is the normal case, not the exception: 22/22 of release-video's spawns and 38/38 of the-gatherer's carry a launch receipt. If the study's 30-of-32 "resolve to a `tool_use` with a matching `tool_result`" was read as completion times, those times are spawn times.

So the rule here is: the **notification** is the outcome for a background spawn; the result is the outcome only when it is the agent's own report. Observed statuses: `completed`, `failed`, `killed`. A test pins it, and the mutation that gets it wrong (any result is an outcome) fails four checks.

**The notification pays for itself twice:** it also carries `<usage><subagent_tokens>…<tool_uses>…<duration_ms>…</usage>`, which is the only place a sub-agent's own spend is written down. That is where the rows' duration, call count and token total come from — reported as the sub-agent's own totals, not as cost, since nearly all of what these read was cache.

### Liveness, and the third state

`mtime` is never a verdict — measured silence inside a *live* sub-agent reached 601s. The row shows `wrote 4m ago` as a fact and stops there. Three states, because a pane killed mid-task leaves sub-agents that never finish and never write again: an outcome-less sub-agent reads `running` only while the pane's own process is alive, and `no result` / `unfinished` otherwise.

### Cost

Nothing reads a parent transcript. The roster is the directory beside it — 187 bytes per sub-agent naming the task, the spawning `toolUseId`, the parent agent and the depth — because the parents here are 292 MB / 101 MB / 10 MB. The statuses come from the window the reader already holds.

```
GET /api/agents/:id/subagents            the roster, from the directory
GET /api/agents/:id/subagents/:agentId   one sub-agent's transcript, as a trace
```

**The PoC's honest limit:** a completion recorded outside the loaded window is not seen, so an older sub-agent reads `unfinished` rather than `done`. The bounded-tail fix belongs on the roster endpoint — that is the next step, not a hidden flaw.

### Level three is the reader, one level down

A sub-agent's transcript is an ordinary trace, so it goes through `splitExchanges` and comes back as exchanges `ExchangeView` renders: task prompt in the prompt band, its own `18 steps · 16 tools · 10m 13s · 171k tok` line, its own fold, its report as markdown. Depth then comes free — rl-llm-wiki ran 3 sub-agents, one of which ran 5 more, and the nested strip reads `5 sub-agents · 5 done` with `↳2` on each row. The directory is flat, so the same endpoint answers for a child of a child.

Also: the count is **this turn's** spawns, and a session usually holds more (including agents that other agents started, which the parent's tool calls never mention). That is stated under the list — `5 more in this session, from other turns` — rather than folded into a number that would disagree with the rows above it.

### Two things the parser had to give up

- **A dropped record.** In a *sub-agent's* transcript (CLI 2.1.209) the completion notification carries `isMeta: true`; in a *parent's* it does not. Dropping it meant a sub-agent that spawned sub-agents could never show one as finished. Kept now, and only when `origin.kind === 'task-notification'`.
- **A prompt that was not one.** Those records open with `[SYSTEM NOTIFICATION - NOT USER INPUT]` instead of a tag, so once kept they read as operator input and opened a fresh exchange with harness noise in the prompt band. The marker joins the two the reader already knew about, on both sides of the wire.

### Codex — left out, deliberately, and not as a zero

Codex's sub-agents are real (a rollout with `parent_thread_id`, `sub_agent_activity` events) but differently shaped, there is exactly one such thread on this machine to test against, and no terminal event has been observed — so the "finished" half would have nothing to stand on. I did not fill the slot with peer-session data either: a `collaboration.spawn_agent` peer is a top-level session in the sidebar, and calling it a sub-agent would misreport it. The strip renders only when a turn actually spawned something, so a Codex pane shows **nothing**, not "0 sub-agents".

### State

Typecheck clean. 21 web suites green, including the new `test/subagents.test.mjs`; every new assertion mutation-checked (any-result-is-an-outcome → 4 fail, exchange-scoped search → 5, two-states-only → 1, Bash-counts-as-agent → 2). Two existing suites stub `../../api` and gained the two new exports.

`server/test/crons.test.mjs` #4 fails, and `server/mobile.test.mjs` — both on clean `4faae2f` as well; I verified by stashing. Not mine.

PoC quality as asked: no polling, no session-wide count, no cost panel. Not merged, not deployed.

### DIFF
--- a/server/src/index.js
+++ b/server/src/index.js
@@ -29,7 +29,7 @@ import {
 // frontend's framing is unchanged.
 const TERM_CTRL = '\x00\x00AM:';
 import { buildUsage } from './usage.js';
-import { buildTraces, traceDigests, digestFor, traceLocation, readTrace, readTraceBundle, readTraceByPath, traceHarnessOf } from './traces.js';
+import { buildTraces, traceDigests, digestFor, traceLocation, readTrace, readTraceBundle, readTraceByPath, traceHarnessOf, subagentRoster, readSubagentTrace } from './traces.js';
 import { initPush, publicKey, deviceCount, addSubscription, removeSubscription, sendToAll } from './push.js';
 import { startVisibilityWatch, isPublic, visibility } from './visibility.js';
 import { kindOfName, kindOfFile, mimeOf, readTextHead, TEXT_MAX } from './preview.js';
@@ -646,6 +646,40 @@ app.get('/api/agents/:id/tail', (req, res) => {
   const text = capturePane(s.id, lines);
   if (text === null) return res.json({ id: s.id, state: 'stopped', text: '', note: 'not running' });
   res.json({ id: s.id, state: deriveState(s, agentInfo().get(s.id)), text });
+});
+
+// The sub-agents this session spawned, and one sub-agent's own transcript.
+//
+// Both read the `subagents/` directory next to the session's transcript and
+// never the transcript itself: the parent is up to 292 MB on this machine and
+// the roster is a directory listing. Whether a sub-agent has FINISHED is not
+// answered here — that fact lives in the parent's own records, which the reader
+// already has for the window it is showing, and inventing an answer from file
+// mtime would be wrong several times an hour (measured: p99 silence 112s inside
+// a live sub-agent, max 601s).
+app.get('/api/agents/:id/subagents', async (req, res) => {
+  const s = store.get(req.params.id);
+  if (!s) return res.status(404).json({ error: 'not found' });
+  try {
+    res.json({ id: s.id, ...(await subagentRoster(s)) });
+  } catch (e) {
+    console.error('[subagents]', e && e.message);
+    res.status(500).json({ error: (e && e.message) || 'roster failed' });
+  }
+});
+
+app.get('/api/agents/:id/subagents/:agentId', async (req, res) => {
+  const s = store.get(req.params.id);
+  if (!s) return res.status(404).json({ error: 'not found' });
+  try {
+    res.json(await readSubagentTrace(s, req.params.agentId, traceOpts(req.query)));
+  } catch (e) {
+    if (['no-trace', 'unsupported-harness', 'trace-not-user-conversation'].includes(e && e.code)) {
+      return res.status(404).json({ error: e.message, code: e.code });
+    }
+    console.error('[subagent-trace]', e && e.message);
+    res.status(500).json({ error: (e && e.message) || 'sub-agent trace read failed' });
+  }
 });
 
 // Block until the target reaches one of `state` — so a coordinating agent makes
--- a/server/src/traces.js
+++ b/server/src/traces.js
@@ -1045,7 +1045,14 @@ async function normalizeClaude(file, out, range) {
   for await (const j of jsonLines(file, range)) {
     // These embed whole file contents; share.js drops them and so do we.
     if (j.type === 'file-history-snapshot' || j.type === 'file-history-delta') continue;
-    if (j.isMeta || j.sourceToolUseID) continue;
+    // `isMeta` is the harness talking to itself, and dropped — with one
+    // exception, added for the reader's sub-agent strip: a `task-notification`
+    // is the ONLY record that says a sub-agent finished, and what it handed
+    // back. In a sub-agent's own transcript (CLI 2.1.209) those records carry
+    // isMeta, while in a parent's they do not, so dropping them meant a
+    // sub-agent that spawned sub-agents could never show any of them as done.
+    // It still renders as a `system` turn, which the reader does not draw.
+    if ((j.isMeta || j.sourceToolUseID) && j.origin?.kind !== 'task-notification') continue;
     if (j.type === 'queue-operation') {
       const text = queueKey(j.content);
       if (j.operation === 'enqueue') {
@@ -1107,7 +1114,12 @@ async function normalizeClaude(file, out, range) {
         // Injected environment/reminder blobs are not prompts — show them, but
         // as system so the conversation reads correctly (same rule the digest
         // uses to avoid counting them).
-        if (j.type === 'user' && (t.startsWith('<') || t.startsWith('[Request interrupted'))) {
+        // `[SYSTEM NOTIFICATION - NOT USER INPUT]` is how CLI 2.1.209 opens a
+        // task-notification inside a SUB-AGENT's transcript, where the parent's
+        // copy opens with the tag itself. It says so on the tin; without it in
+        // this list the record reads as something the operator typed, and the
+        // reader would open a new exchange with harness noise in the prompt band.
+        if (j.type === 'user' && (t.startsWith('<') || t.startsWith('[Request interrupted') || t.startsWith('[SYSTEM NOTIFICATION'))) {
           out.push({ role: 'system', ts, blocks: [textBlock('text', t)] });
           continue;
         }
@@ -1949,6 +1961,89 @@ export async function readTrace(session, opts = {}) {
     sessionId: hit.sessionId || session.sessionUuid || null,
     size: st.size,
   }, opts);
+}
+
+/**
+ * The sub-agents a Claude session spawned, from the directory beside its own
+ * transcript. (PoC — see docs/… nothing; this is the reader's sub-agent strip.)
+ *
+ *   <project>/<session-uuid>.jsonl          the parent, up to 292 MB here
+ *   <project>/<session-uuid>/subagents/
+ *       agent-<agentId>.jsonl               the child's transcript
+ *       agent-<agentId>.meta.json           187 bytes, and the whole point
+ *
+ * The roster is a directory listing plus a few kilobytes: the sidecar names the
+ * task (`description`), the exact spawning call (`toolUseId`), the parent agent
+ * and the depth. Nothing here opens the parent transcript, which is the only
+ * reason this can be asked for on a poll — reading the parent per poll is not
+ * survivable at 292 MB.
+ *
+ * Two timestamps come free from stat(), and they are not the same thing:
+ *   spawnedAt   — the sidecar's mtime. Written once, when the agent is created.
+ *   lastWroteAt — the transcript's mtime. Moves while it works, and then stops
+ *                 whether it finished or died. It is reported, never judged:
+ *                 measured gaps between consecutive records inside a LIVE
+ *                 sub-agent run to 601s (p99 112s), so "silent means dead" is
+ *                 wrong several times an hour. Whether one is finished is
+ *                 decided in the client from the parent's own records.
+ */
+export async function subagentRoster(session) {
+  const loc = await traceLocation(session);
+  if (!loc || session.cli !== 'claude' || loc.format !== 'jsonl') {
+    return { supported: false, reason: session.cli === 'claude' ? 'no transcript yet' : `${session.cli} keeps sub-agents elsewhere`, dir: null, agents: [] };
+  }
+  const dir = path.join(loc.path.replace(/\.jsonl$/, ''), 'subagents');
+  let names = [];
+  try { names = await fsp.readdir(dir); } catch { return { supported: true, dir, agents: [] }; }
+
+  const agents = [];
+  for (const name of names) {
+    const m = /^agent-([A-Za-z0-9_-]+)\.meta\.json$/.exec(name);
+    if (!m) continue;
+    const agentId = m[1];
+    const metaPath = path.join(dir, name);
+    const filePath = path.join(dir, `agent-${agentId}.jsonl`);
+    let meta = {};
+    try { meta = JSON.parse(await fsp.readFile(metaPath, 'utf8')) || {}; } catch { /* a half-written sidecar is still an agent */ }
+    const [metaSt, fileSt] = await Promise.all([statRetry(metaPath), statRetry(filePath)]);
+    agents.push({
+      agentId,
+      agentType: meta.agentType || null,
+      description: meta.description || null,
+      toolUseId: meta.toolUseId || null,
+      parentAgentId: meta.parentAgentId || null,
+      depth: Number.isFinite(meta.spawnDepth) ? meta.spawnDepth : null,
+      spawnedAt: metaSt ? Math.round(metaSt.mtimeMs) : null,
+      lastWroteAt: fileSt ? Math.round(fileSt.mtimeMs) : null,
+      bytes: fileSt ? fileSt.size : 0,
+      hasTranscript: !!fileSt,
+    });
+  }
+  agents.sort((a, b) => (a.spawnedAt || 0) - (b.spawnedAt || 0));
+  return { supported: true, dir, agents };
+}
+
+/**
+ * One sub-agent's transcript, read as an ordinary trace.
+ *
+ * It IS an ordinary trace — same records, same normalizer, so the reader can
+ * render a sub-agent with the component it already uses for a session. The file
+ * is small (11 MB across all 38 of the largest session here) where the parent
+ * is not, which is the other half of why this feature is affordable.
+ *
+ * `agentId` arrives from the browser and lands in a path, so it is matched
+ * against the id shape rather than trusted.
+ */
+export async function readSubagentTrace(session, agentId, opts = {}) {
+  if (!/^[A-Za-z0-9_-]{1,64}$/.test(String(agentId || ''))) {
+    const e = new Error('not a sub-agent id'); e.code = 'no-trace'; throw e;
+  }
+  const loc = await traceLocation(session);
+  if (!loc || session.cli !== 'claude' || loc.format !== 'jsonl') {
+    const e = new Error('this session keeps no sub-agent transcripts'); e.code = 'unsupported-harness'; throw e;
+  }
+  const file = path.join(loc.path.replace(/\.jsonl$/, ''), 'subagents', `agent-${agentId}.jsonl`);
+  return readTraceByPath(file, opts);
 }
 
 /**
--- a/server/src/traces.js
+++ b/server/src/traces.js
@@ -1183,7 +1183,7 @@ async function normalizeClaude(file, out, range) {
 // attach to it — so one row reads "text + 16 tool calls (exec_command,
 // apply_patch, write_stdin)" the way the Hub's own viewer shows it, instead of
 // 17 separate rows.
-async function normalizeCodex(file, out, range) {
+async function normalizeCodex(file, out, range, allowSubagent = false) {
   const stitch = makeStitcher();
   const seenThinking = new Set();
   let cur = null;      // the assistant turn being built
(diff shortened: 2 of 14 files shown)


lvwerra:
Both rounds of feedback are in, pushed as `7a49ab3`. The collapsed view is now the primary one, and the second disclosure is gone.

**Built and shot from the running app** (section 0 of the mock — the rest of that page is where the design was worked out): **https://lvwerra-agent-artifacts.static.hf.space/subagent-rows-mock.html**

```
❯ the sheet is close. rebuild the five scenes we flagged…
  11 steps · 2 tools · 8 sub-agents · 1h 13m · 9.9k tok
  All eight are running. I'll hold the contact sheet until they report.
  ⠹ working · Bash cd /data/workspaces/release-video/film && for i in 1 2 3…
    ⠋ sub-agent  Reshoot scene 3 with clean sidebar          1h 21m  wrote 10m ago
    ⠙ sub-agent  Build intricate grid grounds                1h 21m  wrote 10m ago
    – sub-agent  Build drift-plus-noise grounds  no write in 15m  wrote 44m ago
```

### The four things that decide whether it works

**1 · Running only.** A finished sub-agent leaves the strip; the count in the summary line is what persists. A strip that keeps them grows for the length of the turn and stops being a glance.

**2 · The cap — and a correction to what I reported last round.** I said the busiest real exchange spawned four. That is release-video only. Counting the-gatherer as well: **168 `Agent` calls across 219 exchanges, and its busiest four turns spawn 40, 36, 22 and 12.** Twenty-two at once is real; you were right and I was measuring one session. The strip shows six and then one line — *"…and 6 more running — open the work to see them all"* — and only that line grows. The step list is uncapped, because that is the place to go looking.

**3 · No layout shift.** A row that has just finished holds its place for six seconds with its ✓ and its duration, then fades. And the linger only applies to rows *this client has seen running*: keyed on "first seen finished" alone, every sub-agent that completed hours ago flashed into the strip for six seconds on page load — the exact jitter the linger exists to prevent. I hit that in the first build and the fix is in the commit.

**4 · The killed-pane case.** Three inputs, one function:

| state | rule |
| --- | --- |
| `done` / `failed` / `killed` | the parent recorded an outcome |
| `no result` | no outcome **and the pane is not running** — nothing under a dead pane says "running" |
| `no write in 15m` | no outcome, pane alive, transcript quiet past the threshold — mark goes inert |
| `running` | otherwise |

The threshold is **15 minutes** because the measured worst silence *inside a run that finished normally* is 601s (median 2.7s, p99 112s). My first draft was exactly ten minutes and the new test caught it by one second. It is still a guess, so the row reports how long it has been quiet instead of declaring it dead, and keeps its transcript one click away. This is the case that matters: nine of release-video's 22 sub-agents have no recorded outcome at all.

### The earlier feedback, also in

- **One expandable.** The count is plain text in the summary line; `SubAgentsButton`/`SubAgentsList` are gone.
- **`Agent` is its own step kind**, so a spawn is a row where it happened, among the surrounding tools. It had to stop being a `tools` row: consecutive same-tool calls collapse into `Read ×4`, and spawns arrive consecutively — three inside ninety seconds in one real turn — so grouping rendered four sub-agents as one `Agent ×3` row with no state and nothing to open.
- **One row component, two places.** Opening a row in the strip and opening it in the step list give the same thing: the task (the call's own prompt, clamped) and its trace, drawn by this component one level down.
- **Sub-agents are not double-counted** as tools in the summary line.
- **A spawn with no transcript says so** rather than offering a triangle that 404s — the-gatherer has 168 spawns and 38 sidecars, because the older CLI kept no per-sub-agent file.

### Cost

The roster is fetched only when a turn has a sub-agent whose end is unrecorded, and polled every 15s only while the pane is alive — `lastWroteAt` is the one value that goes stale on its own. It is a directory listing plus 187 bytes per sub-agent; the 292 MB parent beside it is never read for this.

### State

Typecheck clean, 21 web suites green. `test/subagents.test.mjs` gained seven checks over the status rule and the grouping; mutation-checked — group `Agent` like a tool (2 fail), count spawns as tools too (1), let mtime decide without the pane state (1), a 60s stale rule (1).

`server/test/crons.test.mjs` #4 and `server/mobile.test.mjs` fail on clean main too (verified by stashing). Not mine.

Screens are from a live pane whose transcript is a fixture in the real record shapes, so all six states appear in one turn; the sidecars, child transcripts, descriptions and notification numbers are real, and the pane really is running.

lvwerra:
Both changes are in, pushed as `e6c14af`. The first one **deleted** more than it added.

Mock, primary panel refreshed: **https://lvwerra-agent-artifacts.static.hf.space/subagent-rows-mock.html**

### 1 · The strip is the unit, not the row

A sub-agent that finishes keeps its place — ✓, duration, calls, tokens — until the **last** sibling lands; then the whole strip goes and the rows are in the collapsed work with every other step.

```
❯ the sheet is close. rebuild the five scenes we flagged…
  11 steps · 2 tools · 8 sub-agents · 1h 13m · 9.9k tok
  ⠹ working · Bash cd /data/workspaces/release-video/film && for i in 1 2 3…
    ✓ sub-agent  Recapture second-agent sequence    1h 13m  169 calls  331k tok
    ✓ sub-agent  Rapid-fire feature captures        1h 43m  218 calls  329k tok
    ⠋ sub-agent  Reshoot scene 3 with clean sidebar 1h 12m  wrote 1m 14s ago
    – sub-agent  Build drift-plus-noise grounds  no write in 15m  wrote 44m 26s ago
```

**The linger is gone, not disabled.** 40 lines of component, a ref, a timer, a keyframe animation and a `.leaving` class — all of it existed to soften rows leaving one at a time, which no longer happens. Deleted.

**The cap now does real work.** Rows accumulate, so over ten the rows that give way are the **finished** ones, oldest first: the strip is about what is still going, and a landed sub-agent is one click away in the work. Two things fall out of that: the strip's height stops growing at the cap instead of resizing on every completion, and the overflow line describes what it holds back **by state** — *"…and 12 more — 12 done — open the work to see them all"*. You were right that `…and 12 more running` would be a lie in exactly the state the cap exists for.

**The all-done transition, measured rather than argued.** Reader scrolled 400px up from the bottom; the strip's 164px of rows disappeared; `scrollTop` stayed 2578 and the reference prompt band stayed on the same pixel (`-2423`). The strip is the last thing in the last exchange, so nothing above it can move. Pinned to the bottom, the content slides down by the removed height and stays pinned — the same thing that already happens when the working line goes at the end of a turn.

### 2 · The nested trace is a trace

Deleted: the rail around it, the clamped task panel, the "show the whole task" button, the dashed state pills. What renders inside an opened row is an ordinary exchange — the task in the prompt band the reader already draws for a user prompt, because that is what it is, then steps, tools and timings exactly as at the top level. The step body's own indent is the hierarchy.

Two consequences worth naming:

- The child window grew from 384 KB to **2 MB** so the task is actually *in* it. Children here are 0.27–12 MB, most under four, so most arrive whole and the prompt band is the first thing in the box. On the few that overflow it, the tail keeps the report and the row above still names the task.
- Nested tasks are now as long as the operator wrote them — no clamp — exactly like a top-level prompt. That is the price of "highlighted like user prompts", and it seemed the right side of the trade.

### And one real bug this surfaced

The roster that supplies `wrote 4m ago` and the stalled state **never reached the rows**. The effect that fetched it depended on the pane's liveness, and its cleanup ran while the request was in flight — measured in the browser: `roster 29, done? true`, every single time. So rows kept saying "running" with no last-write time, and `no write in 15m` could never fire. The roster now lives in `lib/subagentRoster.ts`: one module-level cache per session with a shared in-flight promise, seeded into state on mount, so a remount starts from what is already known and the reader fetches once instead of once per exchange. The screenshots in the mock are the first ones where those facts actually appear.

### State

Typecheck clean, 21 web suites green. `test/subagents.test.mjs` gained nine checks over the lifetime and the cap; mutation-checked — drop live rows first instead of finished (3 fail), call every hidden row running (3), hide finished rows from the strip (6), keep the strip up when all are done (1).

`server/test/crons.test.mjs` #4 still fails on clean main. Not mine.

lvwerra:
Codex is in scope now, and the reasoning that excluded it was wrong in both halves. Pushed as `a5bcf18`.

The fixture you made worked: `codex-subagent-fixture` produced a parent rollout and three children at 13:55 today (`Curie`, `Herschel`, `Boole`), so I built against that rather than am-dev-3's description.

Mock, with a new Codex panel (0b): **https://lvwerra-agent-artifacts.static.hf.space/subagent-rows-mock.html**

### The terminal event exists — the study looked for the wrong record

`sub_agent_activity` is emitted **zero** times in a run with three sub-agents on CLI 0.149.1. What it emits instead, in the parent's own rollout, is one `agent_message` per child:

```
author: "/root/pty_summary"   recipient: "/root"
Message Type: FINAL_ANSWER
Task name: /root
Sender: /root/pty_summary
Payload: A PTY (pseudo-terminal) is a software interface that…
```

Parent-recorded, names which child, timestamped, and carries what it handed back — the analogue of Claude's `<task-notification>`. **So all three states rest on the same footing as Claude's, and nothing has to be shown as unknowable.** That is the answer to your question about which of the three cases I found: a real terminal signal, not "listable but unassertable".

Two records that look like signals and are not:

| record | what it actually is |
| --- | --- |
| `spawn_agent` output — `{"task_name":"/root/pty_summary"}` | an **acknowledgement**, arriving instantly. Reading it as a result marked every codex sub-agent done the moment it started — Claude's launch-receipt trap, one shape further along. It bit me before the test did. |
| `wait_agent` output — `{"message":"Wait completed.","timed_out":false}` | **names no agent.** It says something finished without saying what, so it cannot carry a per-agent state. |

And one Codex-specific hazard: codex documents this very message format **in its own developer prompt** — `Message Type: MESSAGE | FINAL_ANSWER … Sender: <author>` — in the same thread the real posts arrive in. A loose match reads the explanation as an event; the type has to *be* final, not appear in a list of options. Tested with a doc block that names a real task, which a loose parser completes.

### The join is the task name

There is no shared id anywhere in the chain: `spawn_agent` carries `task_name`, the child's rollout header carries `agent_path`, the completion post carries `Sender:`. All three are the same string, so the name is the link — exact, not timestamp correlation. The rows show the nickname codex gives each child (a better handle than a path), and only when it *is* a name — Claude's `general-purpose` would be noise on every row.

Also worth recording: `collaboration.spawn_agent` is **not** Agent Manager's peer-spawning tool, which is what the earlier scope note assumed. The children are native sub-agent threads (`thread_source: "subagent"`, `parent_thread_id`), and codex said so itself in that run.

### Plumbing

- **roster**: a codex branch — every rollout whose header names this pane as `parent_thread_id`. First lines rather than a directory listing, memoized per file+mtime since a header never changes.
- **child trace**: a thread id is resolved *through* the roster, so a browser cannot ask for a rollout this pane did not spawn.
- **normalizer**: `agent_message` becomes a system turn (the same way a Claude notification arrives) with encrypted parts dropped.
- The "a sub-agent rollout is not a session view" guard is now opt-in rather than absolute: same file, two callers, one of which asked for it by an id it resolved from this pane's own roster.
- **client**: `spawn_agent` joins `Agent`/`Task`. The row, the strip, the three states, the cap, the nested trace — unchanged.

One honest difference in what you see: on the Codex path the nested prompt band shows the **forked conversation**, not a task description, because the message sent to a sub-agent is encrypted. The row's name (`pty_summary`, `Curie`) is what identifies the task. Nothing renders as nothing.

### State

Typecheck clean, 21 web suites green. Seven new codex checks over the real record shapes; mutation-checked — read the ack as a result (4 fail), match `FINAL_ANSWER` loosely (1), drop `spawn_agent` from the spawn tools (7).

`server/test/crons.test.mjs` #4 still fails on clean main. `cron-api.test.mjs` failed once in a full-suite run and passes 11/11 in isolation — a flake, not a regression.

lvwerra:
Both items done, pushed as `ea94980`, and **deployed to am-dev-3** (`4f00a9e`, RUNNING) so the operator can check it now.

Mock refreshed: **https://lvwerra-agent-artifacts.static.hf.space/subagent-rows-mock.html**

## 2 · The nested trace — it was a real bug, and it was Codex-only

**Cause: the child's rollout genuinely contains the parent's conversation.** Codex spawns with `fork_turns: "all"`, so the sub-agent's rollout opens with the whole parent thread copied into it and only then the task it was given:

```
 0  session_meta (the child)          its own header
 1  session_meta (the parent)      ┐
 …  developer/user messages        │  the forked conversation
14  inter_agent_communication…     ┘
15  agent_message  NEW_TASK → /root/pty_summary    ← the child's own start
16… its reasoning, its tools, its answer
```

Read whole — which is what the reader did — an expanded sub-agent replays the parent's history before showing a line of its own work. So it is the file saying that, not the renderer, and the fix is in the parser: drop everything up to the `NEW_TASK` hand-off addressed to that thread. **Only when that boundary is actually found** — a hand-off outside the window means the child is shown whole rather than shown empty. On the fixture run, 30 turns of inherited history became 3 turns of its own work.

Not a display special-case: a child with no fork (Claude's, which begin with their task) is untouched.

**Neither candidate you flagged was the cause, and I checked both.** The id→rollout resolution does not fall back — an unknown id throws, now pinned by a test. And the nested renderer is handed the child's own turns; it was rendering exactly what the endpoint returned.

**Verified on am-dev-3 after deploying**, on the operator's own runs:

| pane | child | what it renders now |
| --- | --- | --- |
| `codex-1` | `hugging_face_news` (Kant) | 3 turns — its own brief, "Latest significant Hugging Face news as of 2026-08-31…" |
| `codex-1` | `amd_news` (Popper) | 3 turns — "AMD news brief — as of 2026-08-31…" |
| `claude-code-2` | `Research Hugging Face news` | 4 turns, its own report — **the Claude path never had the bug** |

And through the UI on am-dev-3: opening `hugging_face_news_2` (Dirac) shows `3 steps · 10 tools · 1m 18s · 7.6k tok` and that child's own report.

## 1 · Rails from the working line down to the rows

`Rails` moved out of `FilesPane.tsx` into `components/Rails.tsx`; both use it and the `.rails`/`.rail` CSS in `styles.css` is untouched. It was genuinely reusable — `prefix: boolean[]` plus `isLast`, nothing file-specific in it.

Two details:

- **The elbow closes on whatever is actually last.** With the cap in play that is the overflow line ("…and 2 more — 2 done"), not the last sub-agent row; a `├` at the bottom pointing at nothing reads as a row that failed to render. `railIsLast(index, shown, hasOverflow)` decides it and is tested three ways.
- **The rail cell is `--mark-w` wide** — the same cell the working line's spinner hangs in — so the tree descends from that spinner rather than from the text column. The file browser's em padding left the elbow touching the mark.

Rails are drawn **in the strip only**. In the step list a sub-agent is one step among many, and a rail there would draw a hierarchy that is not in the data. At depth 2 there is nothing to fight: the rails belong to the head row, so an opened row's nested trace keeps its own indentation underneath.

## State

New server suite `test/subagent-traces.test.mjs`: the fork is dropped, the child's own work survives, a rollout with no hand-off is shown whole rather than empty, and a sub-agent rollout is still refused as a session view. Mutation-checked — no trim (2 fail), trim always (1), drop the session-view guard (1). Plus three checks on the elbow rule.

Typecheck clean, 21 web suites green, every server suite green run individually except `crons.test.mjs` #4, which fails on clean main.

lvwerra:
**"for claude it seems a bit recursive, still?" — reproduced, and it was three bugs in the fork path.**

I had reported the Claude side clean. That was drawn from `general-purpose` children, which inherit nothing. The operator's run is all forks, and forks do.

### What was actually happening

Fetched from am-dev-3 **before** changing anything — `claude-code-2-baec86`, roster of nine children, every one `agentType: "fork"`. Child `a69e9dcf…` ("Research Meta news") began:

```
turn 0  assistant  tool_use Agent · tool_result "Fork started — processing in background"
turn 1  system     <fork-boilerplate> You are a worker fork. The transcript above is
                   the parent's history — inherited reference, not your situation.
```

`subagent_type: "fork"` inherits the parent's transcript, and the **last inherited record is the `Agent` call that created this fork**. So the child contains its own spawning call: it is credited with itself (`1 sub-agent` in its own summary line — the operator's tell), and opening that row opens the same file, one level further down, forever. It is codex's `fork_turns: "all"` exactly, one harness over — the same bug I fixed there last round and then failed to look for here.

Two more from the same data:

- **`Fork started — processing in background` is a launch receipt, not a report.** It was read as an outcome, so every fork was marked done at spawn. `LAUNCH_RECEIPT` now matches both acknowledgements.
- **`fork` in the row is real data**, not a leak from the codex work: it is the `subagent_type` the parent asked for, shown the same way codex's names (Curie, Herschel) are.

### The fixes

- **Parser** (`normalizeClaude`): a fork's inherited prelude is dropped at the `<fork-boilerplate>` boundary the harness writes itself — and only when that boundary is in the window, so a fork whose boundary is outside it is shown whole rather than shown empty. Opt-in via the same `allowSubagent` flag the codex path uses; reading the file as a session is unchanged.
- **Guard, regardless of the data** — `canOpenAgent(agentId, ancestors)` refuses a row whose agent is already open above it, and refuses anything past `MAX_NEST = 3` (deepest real nesting measured here is 2). Such a row still renders **once**, saying `already open above` or `nested too deep to open here`, and does not open.
- One state word per row: a row that cannot be opened says why, instead of stacking `no result` next to `already open above`.

### Counts

Re-checked, no inflation: the operator's turn reads `4 steps · 3 sub-agents` with three rows; the open child reads `1 step · 4 tools · 38s · 7.5k tok` — no sub-agent of its own. Every spawn is one row and is counted once; nothing is dropped from the list to make a number look right.

### Tests

- `server/test/subagent-traces.test.mjs`: a fork fixture — the inherited turns gone, **the spawning call gone with them**, its own answer kept, and the untrimmed read left alone.
- `web/test/subagents.test.mjs`: `canOpenAgent` against a self-referential chain and against the cap, and the fork receipt not counting as a result.
- Mutation-checked: removing the trim fails 2 checks, removing the cycle rule 1, removing the cap 1, narrowing `LAUNCH_RECEIPT` back 2.
- The guard also shot end-to-end against a **deliberately self-referential fixture** (`sa-fixture.py cycle`, a child transcript naming its own spawning call): one row, `already open above`, nesting depth 0.

Full suites green except `server/test/crons.test.mjs` #4, which is red on clean main and not mine.

**Live on am-dev-3** (`5ba0d39`), verified against the operator's own run. Mock refreshed with a new section 0c — the reproduction, the fix, and the guard: https://lvwerra-agent-artifacts.static.hf.space/subagent-rows-mock.html

PR #117, opened 2026-09-02
Remote pane: the state above the input, receipts under their message

lvwerra:
The operator: *"the status (waiting/working/offline etc) should be between input prompt and messages field (like the working widget in the terminal traces, also with the same braille). also the read/pending signs should be at the bottom of the message field it concerns, not on the right."*

Mock, with screenshots of the built thing: **https://lvwerra-agent-artifacts.static.hf.space/remote-pane-state.html**

## What changed

**The state moves above the composer.** A row of its own between `.rp-body` and `.rp-composer` — the reader's working widget, brought to this pane. `working` runs the shared `ov-spin` braille spinner; `listening` and `not connected` draw that motion's completed path, state carried by colour. The app's own words for the three, unchanged: `REMOTE_STATE_LABEL` says `working` / `listening` / `not connected`, because a remote agent has no process here to be "stopped".

**Receipts move under their message.** `.rp-user` is now a caret plus a column: the words, then the receipt on a line of its own, left-aligned with the text. A receipt that trailed the last word landed in a different place in every message and stranded itself under a wrapped one.

## The braille vocabulary had no renderer, and that is why this touches `.status`

The spinner-and-completed-path vocabulary was written as `.status.working` / `.waiting` / `.idle` / `.stopped`. Nothing in the app drew it:

- bare `.status` is a plain colour dot carrying inline provider/CLI colours (`UsagePanel.tsx:92`, `SettingsView.tsx:362`), and its own comment warns that anything drawn on it paints over those;
- `stateLogo.test.mjs:100-105` holds Overview, Sidebar, TerminalPane, RemotePane and Locked to drawing *identity* state on a `StateLogo` frame instead, asserting none of them renders a state-classed `.status`.

So the class the operator's braille mark lives in was one the panes are forbidden to use. Rather than duplicate the spinner and the rectangle under a pane-private name — the "second thing that also means busy" the block exists to prevent — the vocabulary now has a name of its own: **`.state-mark` plus a state class**. Same rules, same measured `--mark-*` contract, no bare-dot baggage, and one live consumer. `stateLogo.test.mjs` is untouched and still green: this is not an identity mark.

Renamed in `styles.css` and in the two suites that name the selectors (`statusMark.test.mjs`, `statusMark.render.test.mjs`); their content is unchanged, and both still lint and measure the same contract. No renderer existed, so nothing's appearance changed anywhere else.

## Two judgement calls

**The state is said once.** It led the bottom context row before, and I removed it there rather than leaving a copy: the same word twice, forty pixels apart, reads as two opinions about one connection. The bottom row keeps what it is for — harness, working directory, host, `last seen`, the key legend. `last seen` in particular stays put, so one place narrates time. Its separators are now interleaved instead of prefixed to each item, because with the state gone the first item varies and a prefixed dot would leave a leading `·`.

**The header keeps its `StateLogo`.** Not a duplicate: it is the CLI tile whose frame carries the state, the mark every pane and every sidebar row draws, and dropping it here would make remote panes the odd one out. Two different shapes at opposite ends of the pane — *which agent this is* at the top, *what it is doing right now* above the input. Terminal panes already pair a header mark with a working line in the body. (The note that `RemotePane.tsx:208` paints a `.status` dot is not so — it is `<StateLogo cli="remote" state={state} …>`.)

## The watch-outs, measured

| Risk | Result |
|---|---|
| a receipt landing moves a bottom-pinned log | both receipts are one element at one line height. In the running app, **727px into a 1351px log, scrollTop and scrollHeight are identical** before and after. Giving them different heights shifts the log 2px — that is how the check was proved |
| a long message strands or misaligns its receipt | the receipt starts on the text's left edge at 620px and at 260px, wrapped to 3+ lines, in both states |
| the mark ignores the pane's zoom | `--mark-size` is the contract's one absolute value (12.5px, measured against Geist Mono's braille ink) and a lint holds it to a single declaration. So the row sizes the *mark* as `0.96em` of the pane instead — 12.5px at 100%, **18.72px at 150%**, growing by exactly the row's ratio |

## Tests

Three, each mutation-tested (break the source, watch the check fail):

- `web/test/remotePaneState.test.mjs` — source invariants, in the default run: the row is written between the log and the composer, the bottom row does not say the state again, the header keeps the StateLogo, the mark is `.state-mark` and the pane declares no spinner of its own, the receipt is inside the message body after the text, neither receipt variant resizes itself, and the `deliveredThrough` wording is intact. Five mutations, all caught.
- `web/test/remotePaneState.render.test.mjs` — Chromium geometry: block heights equal pending vs delivered at 620px, 260px and 200%; receipt alignment under a wrapped message; the row's position; `ov-spin` on a contract braille glyph vs the still path; the mark doubling with the pane. Three mutations, all caught.
- `server/remote-pane.test.mjs` — the real server and the real component, 27 checks, states produced the way the product produces them (an uncollected message, the agent answering, Disconnect). Includes the pinned-log measurement above. Manual, in `npm run test:ui`, with a free port, `PUBLIC_DIR` pinned to the build under test — inheriting the Space's `PUBLIC_DIR=/app/public` silently measures the deployed bundle, which cost an hour here.

`web: 21 suites passed`, `npm run test:render` green, typecheck and build clean. Server suites all pass except `test/crons.test.mjs` #4, which is red on `e492f8a` before this branch and is being handled elsewhere.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "start": "node src/index.js",
     "dev": "node --watch src/index.js",
-    "test:ui": "node terminal-ui.test.mjs && node screenshot-input.test.mjs && node reader-info.test.mjs",
+    "test:ui": "node terminal-ui.test.mjs && node screenshot-input.test.mjs && node reader-info.test.mjs && node remote-pane.test.mjs",
     "test:screenshots": "node screenshot-input.test.mjs",
     "test:mobile": "node mobile.test.mjs",
     "test": "node test/queued-prompts.test.mjs && node ../scripts/run-suites.mjs"
--- a/server/remote-pane.test.mjs
+++ b/server/remote-pane.test.mjs
@@ -1,0 +1,272 @@
+#!/usr/bin/env node
+/**
+ * The remote pane's state row and its receipts, against the real server and the
+ * real component:
+ *   - the state sits between the log and the composer, in all three states, and
+ *     the bottom context row no longer says it a second time;
+ *   - the mark is the app's `.state-mark`: `working` animates the braille
+ *     spinner, `listening` and `not connected` draw its completed path;
+ *   - a message's receipt sits under the message, on its text's left edge, and
+ *     says `pending` until a poll collects it and `delivered` after;
+ *   - that transition does not move the log — the two receipts are one height,
+ *     so nothing below a message shifts when its receipt lands;
+ *   - the mark grows with the pane's zoom control, like the rest of the pane.
+ *
+ * The states are produced the way the product produces them, not by poking
+ * classes on: `working` is an uncollected operator message, `listening` is the
+ * agent having answered it, `not connected` is Disconnect.
+ *
+ * Set REMOTE_PANE_PUBLIC_DIR to a prebuilt web/dist to skip the build,
+ * REMOTE_PANE_PORT to pin the port, and REMOTE_PANE_SHOTS to a directory to
+ * write PNGs of each state.
+ * am-test: manual — Chromium, a full web build and a free port; `npm run test:ui`.
+ */
+import fs from 'node:fs';
+import net from 'node:net';
+import os from 'node:os';
+import path from 'node:path';
+import { spawn, spawnSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+import { chromium } from 'playwright';
+import { chromiumLaunchOptions } from '../scripts/test-chromium.mjs';
+
+const HERE = path.dirname(fileURLToPath(import.meta.url));
+const ROOT = path.dirname(HERE);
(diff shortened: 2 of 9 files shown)


lvwerra:
Closing as a duplicate of #116, which is the branch carrying the operator's latest feedback (running line inside the log, all three states in that position).

My fault, not the author's: I dispatched this task twice. A prompt to one agent timed out, I assumed it hadn't landed and re-sent it to another — it had landed. Both branches did the job properly.

Worth salvaging from here if #116 ever needs it: the `.state-mark` rename in this branch identified a real gap — the braille vocabulary had no renderer of its own and was keyed to `.status.working`. #116 solved the same problem differently, by sharing `.cx-running` with a context override, so nothing is lost today.

PR #119, opened 2026-09-04
Rebuild reader with retained transcripts, recovery and bounded rendering

lvwerra:
## Summary

Rebuild the reader around three explicit owners: bounded transcript transport,
a retained/recoverable reader store, and a measured conversation view. Reader
mode is now an independent surface, not an overlay on an attached terminal.

| Reported symptom | Structural change |
| --- | --- |
| Existing sessions flash “working” when opened | Reading never attaches/claims/resizes a PTY; task activity comes from the transcript |
| Old windows get stuck loading | Request deadlines, cancellation, retry/backoff and foreground recovery; no group first-paint gate |
| Terminal bleeds in from the bottom | Terminal and reader are mutually exclusive; only the reader scroller moves |
| No first prompt without switching to terminal | Composer remains available while loading, empty or failed, including `no-trace` |
| Reader becomes slow with history | Retained stores, bounded rendered rows, cached search text and one conversation-wide outcome index |

Also includes versioned continuous windows, stable record/native-message
identities, cross-page tool/final/queue reconciliation, source replacement
notices, SQLite WAL-aware caching and mutable-tail replacement, and paged/live
child readers with a logical fork boundary. Legacy window clients remain
supported. No new runtime dependency or database migration.

The UI keeps the existing conversation typography and restrained controls:
loaded-history scope, Earlier, Latest, Refresh and actionable error states.

## Validation

- Web production build, all 23 default suites, and `test:render` pass.
- All 27 non-cron default server suites pass, including new protocol regressions.
- Full-app reader-info, attachment-input and terminal UI integration suites pass.
- Chromium regression: first send; hung-leader isolation; deadline/retry;
  no terminal connection on reader activation; 500 exchanges/search with fewer
  than 60 mounted rows; mobile/zoom containment; warm return; prepend/append
  preserving the same row and pixel offset.
- Protocol fixtures: >8 MiB continuous catch-up, cross-window joins, source
  reset, child boundary, WAL-only updates and concurrent parse deduplication.

Server `npm test` stops at an **existing cron restart-test failure**, reproduced
identically on unchanged main `705e32dc`. Remaining suites were run separately.
No production sessions were prompted/restarted; no deployment or merge performed.

## Review notes

- [Architecture and tradeoffs](docs/reader-architecture.md)
- [Detailed regression/validation table](docs/reader-validation.md)
- Switching back to terminal reconnects canonical backend history; local terminal
  selection/scroll position may reset. Reader search is explicitly loaded-history
  search, and virtualized content is not all present for native browser Find.
- SQLite query-level paging/worker offload remains a follow-up. Activity adapters
  beyond Claude/Codex and actual suspended-iOS validation are also called out.
- **Overlaps with #80** (WAL revisions, polling and mutable messages). Those
  requirements are implemented within this new contract; coordinate the two PRs
  rather than merging both implementations unchanged. #80 was not modified.


lvwerra:
Integrated the independent `reader-traces` review of `88c44f5` in `72ef78a` and `78a8a46`.

Fixed metadata-only resets, dropped backward paging, stale working on warm return, inactive-cache eviction, and unnecessary metadata serialization. Reduced automatic whole-file summary revalidation from 30 seconds to five minutes; explicit refresh remains available. Child readers now open at the top of a bounded 2 MiB context window, offer Latest explicitly, and disclose unloaded earlier history.

Integration also uncovered and fixed queued-prompt identity churn, masked lifecycle summaries, lost search anchors, tall-answer restoration drift, and the frontend's blanket rejection of HTML/XML prompts.

[Finding-by-finding decisions, tests and remaining limits](https://github.com/huggingface/agent-manager/blob/feat/reader-redesign/docs/reader-review-followup.md). Full-parser optimization and head-first discovery through large fork preludes remain explicit follow-ups.

Validation: production build; all 23 web suites; four render suites; six trace-related server suites; full-app reader-info integration; diff check. All passed. The previously documented baseline cron failure and Vite chunk warning are unchanged. No merge or deployment performed.

PR #120, opened 2026-09-04
Preview reader and terminal file links inside sessions

lvwerra:
Agents frequently mention workspace files, but reader links are treated as browser paths and terminals only detect web URLs. File references now open a read-only preview inside the originating session pane, preserving its mounted reader/terminal, scroll position and draft. Back/Escape returns through followed links; Close returns to the session. External web links still open new browser tabs.

- Reader Markdown links and inline-code paths resolve against their originating session. Absolute paths, `file:///` links, and `:line:column` / `#L42` references are supported.
- Terminal output gains clickable file paths, including soft wraps and wide-character positioning. Explicit OSC 8 file hyperlinks use the same in-pane preview; HTTP(S) links open new tabs with `noopener noreferrer`.
- **Open in file viewer** creates a retained, passive Files pane in the app (or reuses one already showing that file). It never repoints an unrelated editor. The file selection survives navigation/reloads in the same browser, using existing Files memory; configured extra roots remain read-only.
- The preview reuses the existing Files viewer and styling for Markdown, code, images, HTML and PDFs. It offers source/preview, wrapping, copy path/link and download. Previewing alone never creates a session or attaches another terminal.
- Resolved URLs identify a named root and file independently of the originating session. Relative links and images in Markdown resolve beside the document.
- A read-only API validates lexical and real paths on every request, caps text previews, and preserves the existing HTML sandbox. Optional `AM_FILE_LINK_ROOTS` exposes additional named directories; workspace access is available by default.

Relative paths use the session's configured folder; shell `cd` commands are not inferred. Remote/imported conversations without a local source show an unavailable-file message. Paths outside configured roots must be enabled explicitly. Filenames with spaces or delimiter characters in terminals need an OSC 8 link. These behaviors and configuration are documented in `docs/file-links.md`.

Validation:

- `cd server && node test/file-links.test.mjs` — resolution, additional roots, symlinks, special characters, missing files, bounded reads, download headers and read-only access.
- `cd web && npm test` — all 25 suites pass, including actual reader/xterm clicks opening in-pane previews, preserved source state, canonical URLs, Markdown assets, line jumps and mobile previews.
- `cd web && node test/sessionFilePreview.test.mjs` — retained viewer creation/reuse/reload, failure recovery, focus restoration, and full App navigation to a new Files pane on page 2 of a fixed 1×1 group.
- `cd web && node ../scripts/run-suites.mjs --manual promptBand.render remotePaneState.render` — focused rendering regressions.
- `cd web && npm run build` — TypeScript and production build.

Live preview: [DEV 3](https://lvwerra-am-dev-3.hf.space) (private), source `ae5854a`, deployment `a71565b`.

Live checks passed for in-pane code and PDF previews, unchanged source/URL, mobile layout, and retained viewer navigation/reload. The live test used existing demo files and simulated only passive-pane creation in its browser, leaving no extra sessions or writes behind. Local tests exercise the full creation request and fixed-grid navigation.

The release-video Space (`lvwerra/agent-manager-dev`) is left unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ on local POSIX storage and is restored/checkpointed by
 database backups rather than copying live WAL files. See
 [`docs/agent-state-checkpoints.md`](docs/agent-state-checkpoints.md).
 Scheduled prompts are documented in [`docs/cron-jobs.md`](docs/cron-jobs.md).
+Reader and terminal file links open previews in new tabs; supported paths and
+additional file locations are documented in [`docs/file-links.md`](docs/file-links.md).
 
 ## Architecture
 
--- a/docs/file-links.md
+++ b/docs/file-links.md
@@ -1,0 +1,76 @@
+# File links
+
+File references in the reader and terminal open a read-only preview in a **new
+browser tab**. Web links also open new tabs. The originating session, scroll
+position and composer stay in place; the preview does not start or attach to an
+agent. Markdown, source code, images, HTML and PDFs use the existing Files viewer.
+Unknown binary files offer a download. Transcript JSONL files open as source.
+
+## Supported references
+
+- Markdown links: `[readme](./README.md)`, absolute filesystem paths, and
+  `[source](src/app.ts:42:3)` or `[source](src/app.ts#L42)`.
+- Inline code containing a path, such as `` `docs/notes.md` ``. Fenced code
+  blocks are left alone.
+- Plain paths in terminal output, including soft-wrapped paths and optional
+  line/column suffixes. Explicit OSC 8 `file:///…` hyperlinks also work, including
+  paths with spaces. HTTP(S) OSC 8 links and detected web URLs open new tabs.
+- Relative links and images inside a Markdown preview resolve beside that file.
+
+Relative references from a local session resolve against its configured working
+folder, not the folder of whichever agent is currently selected. `workspace/`
+explicitly means the workspace root. `./workspace/` names an actual subdirectory
+of the session folder. A trace of a local session uses that source session;
+imported traces and remote conversations report that their files are unavailable
+here. Shell `cd` commands and historical per-tool working directories are not
+inferred. Use an absolute path when an agent works outside its configured folder.
+
+After resolving a reference, the preview URL becomes, for example:
+
+```text
+/#file=my-project%2FREADME.md&root=workspace&line=42
+```
+
+“Copy link” copies this URL. It identifies a named file location, independent of
+the originating session. Session rename or deletion does not invalidate it, but
+moving/deleting the file does. It opens the current file contents, not a snapshot.
+The link uses the same private Agent Manager access as the rest of the app.
+
+## Additional file locations
+
+The workspace directory is available as `workspace`. Operators can make other
+directories available for **read-only previews** by setting a JSON map before
+starting the server:
+
+```sh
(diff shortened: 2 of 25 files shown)


lvwerra (commented):
[P1] Bind session-originated links to the originating working folder, not the shared workspace root (`server/src/file-links.js:66-78`).

The only boundary enforced for a session link is the global `WORKSPACES_DIR`. I reproduced this with the server fixture by adding `other-session/secret.txt` next to the existing `team` tree while session `b` had `path: "team/docs"`: both `resolve({ session: "b", file: "../../other-session/secret.txt" })` and the absolute path to that file succeeded. Since preview, raw, and download all use this resolver, a link in agent output or in an untrusted Markdown file can open a sibling session's files. The real-path guard correctly blocks a symlink out of the global root, but it does not stop `..`, an absolute path, or a symlink from leaving one session folder for another folder under that root. The current test even asserts that one `..` traversal succeeds and has no sibling-session case, so its “scope” coverage does not exercise this boundary.

[P2] Refocus the transient preview when following a nested file so Escape still works (`web/src/components/PaneFilePreview.tsx:21-44`).

The focus effect depends only on the boolean `opened`. Following `readme.md -> next.md` changes the trail from one item to two while `opened` remains true, and the keyed `FileLinkPage` removes the focused link. Focus falls back to `document.body`, so the next Escape does not bubble through `.pane-file-overlay` and nothing happens. I reproduced this in `sessionFilePreview.test.mjs` by replacing the Back-button click after `Next` with `page.keyboard.press("Escape")`; it timed out waiting for the original Readme preview. This contradicts the PR and docs claim that Back/Escape traverse followed links.

[P2] Do not force Markdown fragment links into new tabs (`web/src/components/FileLinkContent.tsx:33-36`).

`resolveMarkdown` deliberately leaves `#fragment` links as in-document anchors, but `linkFileContent` then adds `target="_blank"` to every anchor, including references that were not decorated as file links. With a preview containing `[Jump](#section)`, clicking Jump opens a second Agent Manager tab at `#section` instead of scrolling to the heading in the current preview. I reproduced the page count growing from one to two in the real browser fixture. This is a regression from the existing Files preview behavior and is not covered by the new tests.

Verification notes:

- The focused server test, both new browser tests, all 25 web suites, and the production build passed on `ae5854a`. The full server run stopped only at the known-red cron test #4.
- Mutation checks were effective: disabling the real-path guard failed `file-links.test.mjs`, and disabling in-pane click interception failed `sessionFilePreview.test.mjs`.
- A hostile Markdown case containing `script`, `onerror`, and a `javascript:` link remained inert after both sanitization and file-link decoration.
- This head currently has merge conflicts with queued PR #141 in `server/src/index.js` and PR #146 in `web/src/components/FilesPane.tsx`; the latter resolution must preserve #146's keyboard model inside the new `FilesBrowser`/linked-view wrapper.

lvwerra:
Recording the operator's decision on the P1 above, so it is not re-raised later.

**Won't fix — this is the intended behaviour, not a gap.** Agent Manager is a single-operator deployment: every session in the workspace belongs to the same person, and their agents are expected to be able to link any file in it. A link crossing from one session to another is not a boundary being breached, because there is no boundary between them by design.

The review was right to flag it — the code enforces only `WORKSPACES_DIR` and nothing narrower, which is worth knowing — but the narrower boundary it proposed is not wanted here.

Left as-is deliberately: `server/src/file-links.js:66-78`.

PR #135, opened 2026-09-09
Usage traces: hide zero-token sessions and sort all columns

### LINKED ISSUE #126: Usage table: hide zero-token sessions and sort by every column
# Usage table: hide zero-token sessions and sort by every column

## Requested change

The operator requested: “in the usage table: omit empty sessions (no tokens) and allow to sort on any column.” Implement these two changes together in one small PR, scoped to the per-session **Traces** table in Settings → Usage. Provider usage/quota cards and the underlying accounting are unchanged.

## Current behavior

- [UsagePanel.tsx](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/UsagePanel.tsx) renders every `t.sessions` row and eight static column headings, followed by the server-provided total.
- [TraceStats / Traces](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/api.ts#L368) expose numeric `tokensIn`, `tokensOut` and `cacheRead`, as well as the values needed for every displayed column.
- [buildTraces](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/server/src/traces.js#L776) includes sessions with zeroed statistics, orders by most recent activity, and returns totals that also include traces not attributed to a current session. Cache creation is already included in `tokensIn`; cached reads are separate.

No overlapping issue/open PR was found in the current lists. Recheck main and relevant open work before implementation.

### Hide sessions with no recorded tokens

- Omit session rows whose `tokensIn`, `tokensOut` and `cacheRead` are all zero. Do not infer token usage from turns, prompts, tools, file count, session state or activity timestamps.
- Keep input-only, output-only and cache-read-only sessions. Do not filter on formatted values such as `0`, `1.0K`, or a rounded display total.
- Apply this as a display filter; do not delete sessions/traces or change the API response/accounting for other consumers. No “show empty sessions” preference is needed.
- After a successful response with no qualifying sessions, show a concise “No sessions with recorded tokens” empty state. An unresolved/failed request must not be converted into that success-empty state.
- Keep the server-provided aggregate row at the bottom, outside the sorted session rows. Do not recalculate it from the visible rows: it includes deleted/unattributed traces and may contain other activity from filtered zero-token sessions. Retain the explanation of the total's scope and briefly explain that zero-token session rows are hidden.

### Sort every existing column

Each heading must offer keyboard- and pointer-accessible ascending/descending sorting, show the active direction, and expose the active sort state to assistive technology. Keep the existing compact table styling; no new table library or general visual redesign is needed.

| Column | Sort on | First activation of a different column |
| --- | --- | --- |
| agent | Displayed session name, case-insensitive natural ordering | Ascending |
| turns | Raw `turns` number | Descending |
| prompts | Raw `prompts` number | Descending |
| tools | Raw `toolCalls` count, not the tooltip breakdown | Descending |
| web | Raw `web` count | Descending |
| tok in | Raw `tokensIn` value, matching the displayed column | Descending |
| tok out | Raw `tokensOut` value | Descending |
| last active | Raw `lastTs`, not the formatted “ago” text | Descending / newest first |

- Start with **last active, descending**, preserving the current default. Repeated activation of the active heading toggles direction.
- Use deterministic ties, for example name then session ID. Zero/unknown activity timestamps belong after known timestamps in either direction; do not present an unknown date as the oldest known activity.
- Sort a derived copy rather than mutating API/state arrays. Sorting must not refetch/reparse traces or change provider cards and totals. Keep the chosen sort through ordinary renders/data updates while the panel is mounted; no new persisted preference is required.
- Preserve path/tool/cache tooltips, agent logos, token formatting, responsive horizontal scrolling and focus visibility. A focused header must remain focused after its sort updates the rows.

## Required tests and acceptance

Add a small pure filter/sort helper where useful and browser coverage of the actual UsagePanel. Include:

- Mixed fixtures: zero-token sessions with nonzero prompts/tools, input-only, output-only and cache-read-only sessions; empty/all-zero responses; a filtered session gaining tokens in a later response.
- All eight columns in both directions; default newest-first; first-activation direction; numeric values that expose lexical/rounded sorting errors (`9`, `20`, `999`, `1000`, `1040`); mixed-case/numeric names; equal values/duplicate names; missing activity timestamps.
- No input-array mutation and deterministic results. The total remains unchanged and last, including when there are no visible sessions; provider cards remain unchanged.
- Real header clicks and Enter/Space activation, visible direction and accessible sort state, retained keyboard focus, preserved sort on re-render, and no extra network requests caused by sorting. Verify narrow-window/zoom behavior.
- Run focused tests, the default web suite and production web build/typecheck. Report exact results and unrelated baseline failures separately; place new tests in normal discovery. No server changes are expected.

Use fixture/mock APIs only; do not read or alter private live usage logs, prompt real sessions, or deploy/restart production for this feature. Open one focused PR referencing this issue with a screenshot of the sorted table and empty state. No new polling, parser/accounting changes, provider-card filtering, additional columns, or broader Usage-page redesign.

lvwerra:
Closes #126

## What changed

- Hides only session rows where raw `tokensIn`, `tokensOut`, and `cacheRead` are all zero. Input-only, output-only, and cache-read-only rows remain visible.
- Keeps the server-provided aggregate unchanged and last, including in the new successful-empty state.
- Makes all eight existing headers native keyboard/pointer controls. Agent names use case-insensitive natural ordering; numeric columns use their raw fields; unknown activity timestamps stay after known timestamps in either direction.
- Preserves newest-first as the default, toggles the active column, exposes `aria-sort` and a visible direction arrow, and retains the selected sort and keyboard focus through rerenders.
- Adds a pure filter/sort helper plus browser coverage of the actual Usage panel.

## Deliberately unchanged

This stays within the agreed per-session Traces-table boundary. It does not change provider usage/quota cards, trace parsing or accounting, API responses, polling, persisted settings, columns, or the broader Usage-page design.

## Evidence

Sorted on raw token input (descending):

Usage Traces table sorted by token input

Successful response with no token-bearing sessions (the server total remains):

Usage Traces empty state with aggregate total

## Verification

- `npm test -- usageTraces` — 1/1 suite passed
- `npm test` — 24/24 discovered web suites passed (the runner reported its three pre-existing manual render suites as skipped)
- `npm run typecheck` — passed
- `npm run build` — passed (existing bundle-size advisory only)
- Mutation check: the focused suite failed when zero-token filtering was replaced with unconditional inclusion, and failed when Tools was changed from raw `toolCalls` to `prompts`; both mutations were reverted before the passing runs.

### DIFF
--- a/web/src/components/UsagePanel.tsx
+++ b/web/src/components/UsagePanel.tsx
@@ -1,6 +1,13 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import * as api from '../api';
 import type { ProviderUsage, QuotaWindow, Traces, TraceStats } from '../api';
+import {
+  DEFAULT_TRACE_SORT,
+  firstTraceSortDirection,
+  visibleTraceSessions,
+  type TraceSort,
+  type TraceSortKey,
+} from '../lib/usageTraces';
 import Logo from './Logo';
 
 const PROVS = [
@@ -10,6 +17,17 @@ const PROVS = [
   { id: 'hermes', label: 'Hermes', color: '#a78bfa' },
   { id: 'openclaw', label: 'OpenClaw', color: '#c83636' },
   { id: 'gemini', label: 'Gemini CLI', color: '#4796e3' },
+];
+
+const TRACE_COLUMNS: { key: TraceSortKey; label: string }[] = [
+  { key: 'agent', label: 'agent' },
+  { key: 'turns', label: 'turns' },
+  { key: 'prompts', label: 'prompts' },
+  { key: 'tools', label: 'tools' },
+  { key: 'web', label: 'web' },
+  { key: 'tokensIn', label: 'tok in' },
+  { key: 'tokensOut', label: 'tok out' },
+  { key: 'lastTs', label: 'last active' },
 ];
 
 const fmtTok = (n = 0) =>
@@ -64,12 +82,39 @@ function TraceRow({ label, cli, path, st, strong }: { label: string; cli?: strin
   );
 }
 
+function TraceSortHeader({ column, sort, onSort }: {
+  column: (typeof TRACE_COLUMNS)[number];
+  sort: TraceSort;
+  onSort: (key: TraceSortKey) => void;
+}) {
+  const active = sort.key === column.key;
+  const direction = active ? sort.direction : firstTraceSortDirection(column.key);
+  return (
+    <th aria-sort={active ? (sort.direction === 'asc' ? 'ascending' : 'descending') : undefined}>
+      <button
+        type="button"
+        className={`tr-sort${active ? ' active' : ''}`}
+        aria-label={`Sort by ${column.label}${active ? `, currently ${direction === 'asc' ? 'ascending' : 'descending'}` : ''}`}
+        onClick={() => onSort(column.key)}
+      >
+        <span>{column.label}</span>
+        <span className="tr-sort-arrow" aria-hidden="true">{active ? (sort.direction === 'asc' ? '↑' : '↓') : ''}</span>
+      </button>
+    </th>
+  );
+}
+
 export default function UsagePanel() {
   // Each provider loads independently (undefined = loading, null = failed), so
   // the page frame paints immediately and cards fill in as answers arrive —
   // one hung provider no longer blanks the whole page.
   const [prov, setProv] = useState<Record<string, ProviderUsage | null | undefined>>({});
   const [t, setT] = useState<Traces | null>(null);
+  const [sort, setSort] = useState<TraceSort>(DEFAULT_TRACE_SORT);
+  const sessions = useMemo(() => t ? visibleTraceSessions(t.sessions, sort) : [], [t, sort]);
+  const activateSort = (key: TraceSortKey) => setSort((current) => current.key === key
+    ? { key, direction: current.direction === 'asc' ? 'desc' : 'asc' }
+    : { key, direction: firstTraceSortDirection(key) });
   useEffect(() => {
     let alive = true;
     for (const p of PROVS) {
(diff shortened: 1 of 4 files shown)


lvwerra (commented):
Review of #135 against #126. Checked out `112d4d1` (based on `ef08e84`, the exact main commit the issue cites, so none of the issue's line pointers had gone stale) and ran the branch rather than reading the diff alone.

**No findings.** The change is correct and stays inside the agreed Traces-table boundary. Details of what was checked, so the empty review is not taken on trust:

**Sorting, in the real app** (PR build served by the server on a throwaway data dir, Settings → Usage, traces endpoint intercepted with a fixture of nine sessions: a zero-token row that is also the newest, `9/20/999/1000/1040` across every numeric column, `alpha`/`Alpha`/`agent 2`/`agent 10`, two rows tied on `turns` and on `lastTs`, an unknown-timestamp row, and input-only / output-only / cache-read-only rows).
- All eight columns sort correctly in both directions; direction toggles on re-activation by click, Enter, and Space, and a different column starts in its specified direction (agent ascending, everything else descending).
- Ties are deterministic (name then id, and independent of input order). Unknown `lastTs` stays last in both directions. Natural, case-insensitive name order holds (`agent 2` before `agent 10`).
- Hiding and sorting compose: the zero-token row is hidden even though it would top the default newest-first order; the input/output/cache-only rows all stay visible. The filter works on raw fields, so `1000` and `1040` (both displayed as `1.0K`) sort correctly.
- `aria-sort` and the arrow track the active column; the clicked/keyed header keeps focus after the rows reorder. The traces request fired exactly once across all sorts. Provider cards and the total row are unchanged by sorting. At 360px the page has no horizontal overflow and the table scrolls inside its container.
- The real (unmocked) endpoint on a fresh data dir returns no sessions with a non-zero total; the table shows “No sessions with recorded tokens” with the server total still last.

**Verification claims**: `npm run typecheck` passes, `web/test/usageTraces.test.mjs` passes, `npm test` reports 24 suites passed with the three `*.render.test.mjs` suites skipped as usual, and `npm run build` passes with only the pre-existing chunk-size advisory. The new test is in normal discovery.

**Does the test actually fail without the fix?** I ran twelve mutations against the focused suite: unconditional inclusion, ignoring `cacheRead`, ignoring direction, dropping the tie-break, dropping the unknown-timestamp handling, lexical numeric comparison, case-sensitive names, agent-first-descending, no direction toggle in the panel, no empty row, no `aria-sort`, and reversing the input before sorting. Eleven were caught. The only survivor is the input reversal, which is expected: the comparator is a total order, so the output cannot depend on input order. That is the determinism the issue asks for, not a coverage gap.

**Scope**: only `UsagePanel.tsx`, the new `lib/usageTraces.ts` helper, the `.traces-table` CSS block, and the new test are touched. No server, API, parsing, polling, persisted-preference, provider-card, or column changes. Server rows always carry numeric fields (`emptyStats()` in `traces.js`), so the helper's `!== 0` and subtraction comparisons are safe on real data.

Issue #131, opened 2026-09-09
Make privacy verification and lock transitions reliable


lvwerra:
## Goal

Make the existing privacy safety lock a coherent, tested lifecycle: verify the deployment's privacy, apply the same effective lock state to new requests and existing client connections, and show an accurate status that automatically clears after successful verification.

This is one focused PR, approved by the operator after discussion. It is a safety net for accidental public exposure, not a Reader performance fix or a replacement for keeping the Space and storage private. Preserve the single-operator deployment model, supported local mode, and the existing explicitly unverified bucket warning policy.

This public issue describes desired safeguards and compatibility. Follow [SECURITY.md](https://github.com/huggingface/agent-manager/blob/main/SECURITY.md) for private coordination of vulnerability evidence; do not paste private review reproductions, operator data, credentials or live endpoint details here or into a public PR.

## Requirements

| Area | Required behavior |
| --- | --- |
| One authoritative policy | Use a small, explicit server-owned visibility model shared by HTTP admission, connection lifecycle and status responses. Represent confirmed private, confirmed public, unknown/verification failure, and deliberately unverified bucket coverage separately. Publish an effective locked flag and accurate reason; unknown is not a claim that a resource is public. Avoid independent safety decisions in individual UI panes. |
| Verification evidence | Define and document exactly which response/status/schema establishes a privacy verdict for the Space, bucket discovery and each discovered bucket. Verify current official HF API semantics rather than assuming every unauthorized/not-found response proves privacy. Network failures, timeouts, rate limiting, server errors and malformed/unexpected responses are not successful verification. A valid response about one resource must not refresh the verification age of another. |
| Startup and outages | A Space startup without valid privacy evidence stays locked while bounded checks continue; health and the safe status/setup page remain available. Once the required privacy checks have succeeded, allow only a short, explicit grace period for transient verification outages. Measure it from the last successful applicable verification, not from the latest attempt, retry or page visit. Select and document one short constant in the PR; two normal polling intervals is an implementation starting point, not an operator-selected duration. No new settings UI is required. Expiry must take effect even if a check is still pending. Do not carry an unvalidated grace entitlement across process restarts. |
[...]

Issue #127, opened 2026-09-08
Reader/terminal switching: reduce terminal delay and preserve scroll position


lvwerra:
# Reader/terminal switching: reduce terminal startup delay and preserve reading position

## Operator report and PR scope

The operator reports that switching from Reader to Terminal takes **1–2 seconds** to show the terminal, and switching back leaves the Reader **away from the bottom**, despite having been at the bottom before switching.

Address these two mode-switch problems together in one focused PR. The delay and incorrect position are user-reported; source inspection below identifies relevant lifecycle paths but does not establish the root cause or constitute a fresh reproduction. First reproduce with isolated fixtures and measure the switch timeline.

## Current implementation and related work

Checked main at `ef08e843d1c67fcc4e1c463415819ea351734d6d` on 2026-09-08:

- [TerminalPane lifecycle](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/TerminalPane.tsx#L468): its terminal effect depends on `session.id` and `reading`. Entering Reader closes the WebSocket, disposes xterm and removes its surface; entering Terminal constructs/opens xterm and connects again. The boot cover probes upper-screen content after output bursts, with a 150 ms probe and a 20-second safety cap. A local preview also exists; inspect its refresh/lifetime rather than assuming it covers every warm switch.
- [Terminal restore](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/server/src/index.js#L2943): the backend sends a canonical grid/restore frame and retained ANSI output. Detaching a viewer is not stopping its session. Preserve output ordering and controller/geometry semantics.
[...]

Issue #123, opened 2026-09-08
Save settings immediately and make file-editor saves reliable


lvwerra:
# Save settings immediately and make file-editor saves reliable

## Agreed behavior and PR boundary

The operator approved reliable file/settings saving with one explicit change: **settings save immediately**. Remove the existing autosave debounce; do not wait for blur, a period without typing, or closing the settings panel.

- Settings configuration and secret-description changes start saving as soon as a valid edit occurs. If a write is already active, retain the latest pending changes and send them as soon as it settles, without another delay timer.
- Files remain explicit-save. Typing in the file editor must not start an autosave.
- “Saved” means the current edited version was successfully persisted, not merely that some older request succeeded.
- Older success/failure responses must never erase newer edits, clear their existing draft, or incorrectly mark them saved.
- Failures and conflicts remain visible with an appropriate Retry action. Navigation must not silently discard pending changes.
- No new backups, trash, file version history or restore feature. Preserve the existing unsaved-file draft behavior correctly; this is not a request to remove it.

Review coverage: B06, B11, C05, U05, and the settings-endpoint portion of B02. Keep the PR scoped to files, configuration and secret descriptions, their save contracts, and focused UI/test changes. Do not migrate every application store or build a new global state framework.

## Starting points and coordination

Checked `main` at `ef08e843d1c67fcc4e1c463415819ea351734d6d` on 2026-09-08. Recheck the branch and existing work before implementation.

- [SettingsView](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/SettingsView.tsx#L156): configuration and secret-description loading, the two 900 ms save effects, saved-state indicators, and dependent status refreshes.
- [FilesPane/FileView](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/FilesPane.tsx#L626): `flush`, `pending`, `inflight`, `onEdit`, `SaveState`, close/navigation guards and remembered drafts.
- [Settings routes](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/server/src/index.js#L1165): `loadAmConfig`, `loadSecretNotes`, `/api/config`, `/api/secrets`, generated environment-skill updates, and their persistence/error contract.
[...]

Issue #122, opened 2026-09-08
Files: safe replacements and dozens of reader attachments


lvwerra:
# Files: safe replacements and dozens of reader attachments

## Goal and agreed scope

Support normal batches of dozens of reader attachments while making workspace uploads and folder operations safe and deliberate. This is one PR covering the upload flow from selection to storage and prompt delivery, plus the existing file-operation boundary checks.

The operator agreed to:

- Remove the five-file reader attachment limit in both frontend and backend; support dozens of files, with 50-file acceptance coverage. **50 is a test case, not a new per-prompt limit.**
- Adjust the 20-uploads-per-minute restriction so an ordinary supported batch works, while retaining file-size/storage safeguards and bounded upload work.
- Keep progress, cancellation, retry and a compact attachment list usable for large batches.
- Keep same-name reader attachments as separate stored files.
- Require explicit confirmation before replacing an existing workspace file, with **Cancel as the default**.
- Preserve existing workspace files on failed/cancelled upload; enforce intended filesystem boundaries; prevent removal/relocation of folders that configured agent workspaces or shared skills depend on.

Review coverage: S03, B04, B05, plus the operator's request for larger reader batches. The IDs are bookkeeping; implementation requirements are self-contained below.

**Out of scope:** file-editor save-state/autosave redesign, trash/restore/version history, transactional relocation of agent workspaces, a general upload service rewrite, new remote-agent attachment support, durable exactly-once prompt delivery, and another reader redesign. Preserve current supported harness behavior rather than adding integrations here.

## Starting points and existing work

Checked `main` at `ef08e843d1c67fcc4e1c463415819ea351734d6d` on 2026-09-08. Recheck before implementation.

- [File routes](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/server/src/index.js#L1754): `resolveSafe`, file create/write/upload, `dependedOnDir`, rename/move/delete. Upload is under `/api/files/:id/upload`.
- [FilesPane](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/FilesPane.tsx#L934) and `web/src/api.ts::uploadFile`: selection, target folder, collision/error handling.
[...]

PR #136, opened 2026-09-09
Centralize browser request admission for HTTP and terminals

### LINKED ISSUE #130: Centralize browser request protection for HTTP and terminal connections
# Centralize browser request protection for HTTP and terminal connections

## Goal, product boundary and disclosure

Implement one consistent request-admission policy for browser-triggered state changes and terminal WebSocket connections. The normal Agent Manager UI should continue working transparently, both standalone and embedded on Hugging Face. Unrelated pages and untrusted file previews must not be able to perform privileged manager actions through the operator's browser.

This is a **public hardening/compatibility specification**, not a vulnerability report. Do not add private review reproductions, payloads, real deployment identifiers, credentials or exploit narratives to this issue or a public PR. Follow [SECURITY.md](https://github.com/huggingface/agent-manager/blob/main/SECURITY.md) and use GitHub's private security-reporting channel for any sensitive findings and remediation coordination.

Keep this to one PR. Preserve the existing private, single-operator deployment model: **no new user accounts, login screen, multi-user permissions or blanket API lockdown**. Browser-request protection does not authenticate a native client or make a publicly reachable manager safe. Preserve the existing edge/access restrictions and visibility lock; changes to visibility verification or revoking already-established connections when privacy changes belong in a separate batch.

## Integration starting points

Planning inspected main on 2026-09-09. Recheck current main and open work before implementation. This section identifies integration surfaces, not affected-version or exploit evidence.

- `server/src/index.js`: app middleware order, raw/text/JSON routes, HTTP server and WebSocket upgrade/attachment handling, internal cron requests, generated environment instructions and remote connection prompts.
- `web/src/api.ts`: shared fetch wrapper **and** XMLHttpRequest attachment uploads; `web/src/components/TerminalPane.tsx`: browser socket setup and reconnect behavior. Inventory other direct request paths rather than assuming all callers use the wrapper.
- `server/src/operations.js`: attribution is a separate concern. Preserve who-requested-what semantics without treating a display identity as authorization or recording rejected request contents.
- `web/vite.config.ts`: explicit local development UI/API/socket proxy arrangements. Coordinate [#108](https://github.com/huggingface/agent-manager/pull/108) for local installation support.
- `server/src/remote.js`, generated remote prompts, `docs/remote-agents.md`, `docs/cron-jobs.md`, bundled helpers/hooks and tests: maintain non-browser compatibility. Some protocol reads maintain contact or delivery state; classify by effects, not solely by HTTP verb.

### 1. Explicit origin configuration shared by HTTP and WebSockets

- Introduce a small tested policy module rather than independent route-specific string checks. Define permitted browser page origins using the full **scheme, hostname and effective port**. Normalize valid origins consistently, reject malformed/ambiguous inputs and require exact membership; no suffix, substring, wildcard-subdomain or all-localhost-ports allowance.
- Derive normal Space origins from trusted deployment configuration, with deliberate exact-origin configuration for supported custom hosts/proxies. Production must not silently enable development origins. Local backend and Vite UI ports must be explicitly supported in local mode, including documented alternate ports and IPv6 loopback if supported.
- Never populate the allowlist from an unchecked request `Host`, `Forwarded`, `X-Forwarded-*`, parent-page URL or query parameter. Document any proxy trust assumption and validate the target host against configured deployment identity. Do not globally trust all proxies as a shortcut.
- Distinguish a missing Origin from an explicit opaque/null Origin. A presented disallowed, malformed or opaque origin must not become permitted just because another header, caller ID or query parameter is present. Do not fall through to a native-client exception after a failed browser-origin check.
- Add Fetch Metadata as defense in depth where appropriate, with an explicit policy for absent or unfamiliar values. It must not be the only protection, must not treat same-site as automatically same-origin, and must not break the trusted UI's legitimate embedded request context. Verify the combination in real browser fixtures, including cross-origin ancestors, rather than guessing the headers sent by an iframe.

An Origin identifies a scheme/host/port and may be absent or opaque in browser contexts; those cases need separate treatment. [MDN Origin reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Origin).

### 2. Protect effects before they happen

- Apply admission centrally to mutating HTTP actions, including bodyless commands, JSON/text requests, streaming workspace uploads, reader attachments and aliases. Check before body capture or handlers can write files, schedule work, deliver input, start processes, change state or begin an external operation. Keep raw streaming paths streaming; the guard should not read an entire upload body.
- Require a deliberate browser anti-forgery mechanism in addition to origin validation. For this API-driven single-owner app, prefer an app-specific non-safelisted request header with restrictive preflight handling; a vetted token design is an alternative if the deployment requires it. Apply it consistently via shared client helpers, not individually maintained per-action boilerplate. Do not confuse a non-secret protocol marker with an authentication credential.
- Do not implement the protection merely by withholding CORS response headers. Reject disallowed operations on the server. OPTIONS/preflight requests must perform no work, grant only documented exact origins/methods/headers if cross-origin UI access is genuinely needed, and never reflect arbitrary credentials/origins. Do not enable broad CORS for third-party clients; command-line clients do not need it.
- Inventory read-looking routes that establish control, contact, delivery/acknowledgement state or other meaningful effects. Protect those effects or separate them from safe reads; do not exempt all GETs automatically. Keep ordinary page/static loads, read-only previews, downloads, health/setup information and normal Reader reads functional. Do not turn every harmless read into a new login/token flow.
- Reject before successful WebSocket upgrade where practical, and always before terminal attachment, startup, replay, control or resize. Browser socket handshakes use the same exact-origin policy. Do not require an HTTP header that the browser WebSocket API cannot set; if additional handshake proof is chosen, define a compatible transport and keep credentials out of URLs/logs. Disallowed connections must never emit terminal output or affect an existing viewer.
- HTTP failures use a stable, actionable error response and no domain side effect. WebSocket failures must not cause endless aggressive reconnects, restart a stopped agent, lose the user's Reader draft or prompt the user to disable protection.

OWASP documents custom-header protection for API requests as relying on restrictive browser preflights, not on the header being secret. It also calls for protecting state-changing GETs. [CSRF prevention guidance](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#employing-custom-request-headers-for-ajaxapi). WebSocket handshakes need their own origin validation. [WebSocket security guidance](https://cheatsheetseries.owasp.org/cheatsheets/WebSocket_Security_Cheat_Sheet.html#origin-header-validation).

### 3. Deliberate compatibility for native clients

- Define and test a supported no-browser-Origin path for trusted in-container agents, internal scheduler/hooks, command-line scripts and remote clients using the existing deployment authentication. **Absence of Origin alone is not permission.** Require explicit request proof appropriate to that path, such as a non-safelisted protocol header within the existing trusted deployment boundary, or separately validated credentials where applicable. Reject contradictory browser-origin/Fetch Metadata evidence instead of letting a client label override it.
- Keep authentication, request intent and audit attribution distinct. `from`, a session ID, a remote name and attribution fields remain labels; none becomes a secret or grants authority. Do not accept an arbitrary Authorization-looking string as verified authentication, or copy the operator's HF credential into browser JavaScript to solve compatibility.
- A protocol marker is acceptable as an anti-forgery gate under the documented private/edge-protected model, not as proof of a particular native caller's identity. If a design requires new app-level credentials, an external authentication service or manual per-agent provisioning, stop and obtain a separate product/security decision rather than silently expanding this PR into a new auth system.
- Update bundled HTTP/WS clients, generated instructions, remote connection prompts, internal cron requests, lifecycle hooks and relevant integration fixtures in the same PR. Verify frontend uploads that bypass fetch and retain legitimate text/plain prompt support. Changes must not require the operator to hand-edit every normal UI request.
- Document the exact command-line compatibility change and rollout path for already-running agents or copied remote polling scripts. Old clients must get an actionable failure and a concrete upgrade instruction; do not add a production-wide bypass or a catch-all legacy/no-Origin exception. Do not restart or re-prompt live agents as part of implementation/testing. Any disruptive rollout action requires separate authorization.
- Preserve normal external edge authentication and local deployment protections. A connection appearing to come from loopback is not alone enough to bypass browser checks when a proxy or local browser can reach that address.

### 4. Preserve embedding, previews and existing behavior

- Verify standalone Space UI, the UI inside a Hugging Face-style parent frame, configured proxy deployment and local Vite development. Authorize the actual app's origin, not every origin belonging to the hosting provider. Preserve intended embedding; do not add a frame-deny policy or trust the whole parent site merely to pass an iframe test.
- Keep file/HTML preview isolation intact and usable. Do not give an untrusted preview same-origin privileges, allow it to obtain request credentials, or route its commands through a trusted parent bridge. Read-only preview rendering and downloads must remain possible under the scoped policy.
- Preserve browser request cancellation, multipart/raw content handling, attachment progress, multi-tab use, terminal ownership/geometry, Reader-only operation and remote long polls. If a token design is selected, define tab/reload/rotation behavior without automatically replaying an uncertain mutation.
- Log only bounded rejection metadata needed for diagnosis, such as a reason code and route class. Do not capture rejected bodies, tokens, cookies, full URLs/query strings, file contents or terminal output. Do not redesign the broader audit capture/retention policy in this PR.
- Security tests must run with the real guard enabled. Existing fixtures that relax audit attribution must not accidentally disable request protection. Avoid a general production-disable flag introduced solely to keep old tests green.

## Required tests

Use a policy-unit matrix plus real HTTP/upgrade integration and browser tests. Include positive cases as rigorously as negative cases. Browser inability to read a response is not proof that a rejected action did not occur: assert handler invocation counts and unchanged fixture state.

| --- | --- |
| Origin policy | Exact trusted origin, equivalent default-port normalization, a different scheme/port/host, local UI versus unrelated local port, malformed/duplicate origin representations, missing origin and opaque/null origin. Only the intended branch may admit the request. |
| Configuration/proxies | Space configuration, exact custom origin, local backend/Vite configuration, missing/invalid settings, HTTPS termination and untrusted forwarded/host values. No request can broaden the allowlist; normal configured deployments remain usable. |
| HTTP methods/bodies | Representative state changes with JSON, text/plain, form-compatible encodings, bodyless commands and streamed files. Missing/invalid request proof is rejected before action/body storage; valid UI/native flows preserve payloads and run once. OPTIONS, rejected requests and aliases cannot trigger work. |
| Browser contexts | Trusted standalone UI and app iframe, unrelated browser page, sandboxed local preview, forms, no-cors requests and preflighted requests, with real browser-generated metadata. Negative fixtures cause zero privileged side effects, including when responses are unreadable. Keep all fixtures harmless and synthetic; coordinate any sensitive proof privately. |
| WebSockets | Allowed browser handshake and documented native handshake, rejected/missing/opaque origins as defined, failed upgrades and reconnects. No rejected client attaches/starts/claims/resizes a PTY or receives replay. An already-connected valid viewer is unaffected by rejected attempts. |
| Native clients and attribution | Internal cron, local agent helper, generated hook, copied remote client and deliberate API scripts; marker/proof missing versus present, invalid claimed identity, contradictory browser headers and credentials handled by a mocked trusted edge. Preserve attribution checks separately; caller labels alone never satisfy admission. |
| Effects behind reads | Inventory the relevant read-looking protocol endpoints and test that rejected requests cannot change contact/delivery/control state. Ordinary Reader reads, previews, downloads and health/setup reads remain usable with their documented access policy. |
| UI compatibility | Actual shared fetch and XMLHttpRequest paths, file upload progress/cancel, settings/file saves, reader attachments/first prompt, multiple tabs, browser return, terminal opening and remote streaming. Use fake dispatchers/PTYs/external services; do not perform real costly or destructive actions. |
| Failure and migration | Actionable rejection shape, no content/credential capture in diagnostics, no automatic mutation replay, no reconnect storm, already-open tab/script behavior, and updated helper examples. No test-only bypass reaches production defaults. |

- New suites must be normally discovered. Run focused policy/API/browser checks, relevant terminal/remote/upload/operations suites, normal web/server suites and the frontend production build/typecheck. Report exact commands, tested runtimes/browsers, and unrelated baseline failures separately; do not omit coverage just because an earlier unrelated suite stops discovery.
- Isolate all state/workspace/home roots and transports. Use ephemeral local test origins and mock the HF edge where needed. Do not probe public third-party sites, production APIs or live agents; do not launch paid jobs, send notifications, change hosting privacy or deploy. Report which embedding/proxy behavior was fixture-tested versus actually verified; do not claim a production edge test that was not performed.
- Include a concise policy table in the implementation documentation mapping browser, embedded UI, native/internal client, remote client, preview and malformed/unknown cases to admission rules. Document middleware/upgrade ordering and how adding a new endpoint/client inherits the protections.

## Handoff and exclusions

Coordinate the shared frontend/server request paths with [#122](https://github.com/huggingface/agent-manager/issues/122), [#123](https://github.com/huggingface/agent-manager/issues/123), [#127](https://github.com/huggingface/agent-manager/issues/127) and [#129](https://github.com/huggingface/agent-manager/issues/129), plus local hosting [#108](https://github.com/huggingface/agent-manager/pull/108) and generated instructions [#109](https://github.com/huggingface/agent-manager/pull/109). Their feature scopes remain unchanged.

No sharing-publication redesign, privacy-verdict/revocation overhaul, broad CSP rollout, secret-redaction project, request/fleet quotas, API schema rewrite, new login/RBAC system or general route extraction. Open one focused PR with policy rationale, caller migration, compatibility evidence and tests, following private security coordination where needed. Do not merge, deploy, disable protections or update live agent scripts without separate authorization.

lvwerra:
Closes #130

## Agreed scope implemented

- One tested admission policy for HTTP actions and terminal upgrades: exact configured scheme/host/effective-port origins, target-host validation, duplicate/malformed header handling, and Fetch Metadata checks that preserve the trusted embedded app context.
- Central HTTP admission before parsing, streaming uploads, audit capture and handlers. The non-secret `X-AM-Request: 1` marker is independent of authentication and attribution. OPTIONS has no handler effects and grants no API CORS access.
- Protected contact/delivery and external-operation reads, including HEAD and Express aliases. Ordinary Reader reads, previews, downloads and health/setup pages retain their existing access model.
- Terminal admission before upgrade/attachment/startup/replay/control; browser sockets use exact Origin, native sockets require explicit intent. Bounded frontend reconnects offer an explicit retry without clearing retained output or Reader drafts.
- Shared fetch/XHR client updates, internal cron requests, generated environment/remote instructions, documented copied-client migration and native integration fixtures. Lifecycle adapters were inventoried: they use local breadcrumbs, not HTTP, so their protocol is unchanged.
- [Policy, configuration, route inventory and migration documentation](https://github.com/huggingface/agent-manager/blob/be62f4ddbcdd60dcc6bc7749bcd41fd5261e6e85/docs/browser-request-protection.md).

This remains a single-owner application behind its existing private deployment/edge boundary. The marker is not a secret, an identity or a replacement for edge authentication.

## Compatibility and deliberate exclusions

Standalone and embedded app requests, HTTPS termination with either a preserved public Host or configured loopback upstream Host, native clients behind a mocked private edge, and the real Vite API/socket proxy were exercised with local fixtures. The allowed page origin is the app's origin, not the hosting provider or parent page. No production HF edge/cookie behavior is claimed as live-tested.

The frontend adds intent transparently, including attachment XHR. Existing copied scripts must update their headers; old clients get a stable actionable refusal. No automatic mutation replay or production legacy bypass was added. No installed live scripts were changed and no agents were restarted/re-prompted.

Deliberately unchanged: visibility verification/revocation (#131), sharing/publication, broader CSP or audit/redaction policy, quotas, authentication/accounts/RBAC, API schema/error redesign, general route extraction, upload storage/limits (#122), settings persistence (#123), and Reader mode/history product scope (#127/#129). Preview isolation and the existing audit-attribution policy remain intact.

## Verification

Node **22.23.2**, Chromium **153.0.8010.12**, OpenSSL **3.0.20**. Test processes used isolated state/home/config roots, synthetic local origins and fake dispatchers or test shells; no production services, paid agents/jobs or notifications were exercised.

| Command (working directory) | Result |
| --- | --- |
| `npm ci --no-audit --no-fund` (server and web) | Passed |
| `node --test test/request-admission.test.mjs test/request-http.test.mjs` (server) | 7 tests passed; policy/header/config matrix, HTTP ordering/body types/reads/preflight, upgrade refusal and an unaffected valid viewer |
| `node test/request-api.test.mjs` (server) | Passed; real backend admission, attribution, uploads/attachments, preview sandbox, terminal startup refusal, remote contact/delivery/stream cancellation, generated clients and audit-test-flag separation |
| `npm test` (web) | All 25 discovered suites passed, including new browser/proxy suites and the real TerminalPane retry/draft regression |
| `npm run build` (web) | Typecheck and production build passed; existing bundle-size warning |
| `npm test` (server) | Stopped at the known baseline `crons.test.mjs` test 4 after six suites passed |
| Remaining server suites, explicit command below | All 25 selected suites passed; includes one already-passing repeated suite, covering all other normal suites |
| `npm run test:ui` (server) | All three manual terminal, screenshot/upload and Reader suites passed |
| `npm run test:render` (web) | All four render suites passed |
| `npm run test:mobile` (server) | Existing ambiguous Back-to-list selector reproduced on both this branch and clean `ef08e84`; local-only diagnostic continuation described below |

Remaining server command (run despite the known cron failure):

```sh
node ../scripts/run-suites.mjs first-run hidden input-required next-name opencode-resume operations queued-prompts reader-protocol repin request revive slowfs spawn-group subagent-traces terminal-modes trace-download trace-tail trace-window usage migration resize state-checkpoint
```

Browser fixtures use actual shared fetch/XHR and browser-generated headers, including an app inside a cross-site ancestor, opaque preview, forms, no-cors/preflight requests, multiple tabs, settings/file saves, first prompt, raw uploads and upload progress/cancellation. Negative checks assert unchanged fixture state, not merely unreadable responses. The Vite fixture uses explicit alternate ports and the actual proxy configuration.

Mobile diagnostic continuation: in separate local checkouts only, disambiguated the two Back-to-list selectors to select the visible button, then reran `npm run test:mobile` on both clean main and this branch. Both passed the same 21 checks (including terminal ownership, retained history, orientation, touch and embedded keyboard geometry), then encountered the same existing hidden-sidebar interaction timeout. Those diagnostic fixture edits are not part of this PR; the full mobile suite is **not** claimed green.

Eight mutation checks were detected: removing exact-origin enforcement, required intent, protected-read classification, central HTTP wiring, pre-upgrade admission, shared fetch intent, XHR intent, and changing the terminal retry budget. They ran in a separate disposable checkout; all mutations were restored and focused transport tests passed again.

Two review rounds are required for this security-adjacent hardening change. This PR is the handoff for the assigned review; it has not been merged or deployed.

### DIFF
--- a/docs/api-errors.md
+++ b/docs/api-errors.md
@@ -1,0 +1,159 @@
+# HTTP validation and failures
+
+This is the compatibility inventory for the Express **4** request boundary.
+The routes and successful payloads are unchanged. HTTP failures retain the
+top-level `error` string used by copied shell/agent clients, and add `code`:
+
+```json
+{"error":"name: must not be empty","code":"invalid-input","details":[{"field":"name","message":"must not be empty"}]}
+```
+
+`error` is display text, not HTML; clients branch on `code`, not English.
+Expected exceptions use `ApiError`. Arbitrary exceptions (including library
+errors carrying their own `status`) do not authorize exposing their messages.
+Unexpected errors return a generic 500. Legacy local 5xx JSON replies are also
+sanitized before the audit recorder sees them. An optional `requestId` is a
+server-generated UUID, never a request-supplied identifier. No exception body,
+stack, upstream response or new secret-bearing diagnostic is logged here.
+
+## Status and code contract
+
+| Status | Codes | Safe extra data / meaning |
+| --- | --- | --- |
+| 400 | `invalid-input`, `bad-request`, `invalid-json`, `invalid-path` | `details: [{field,message}]`; no rejected values echoed |
+| 400 | `origin-required`, `unknown-origin` | Existing attribution requirement, unchanged ordering |
+| 400 | `bad-repo`, `not-a-bundle` | Existing import refusals |
+| 400 | `request-aborted` | Upload/body interrupted before a response was committed |
+| 403 | `locked` | `reason: public-space`; health/info/visibility remain accessible |
+| 403 | `forbidden`, `no-access`, `no-hf-token`, `backup-unavailable` | Expected access/capability refusal |
+| 404 | `not-found`, `api-not-found` | Missing item versus nonexistent API route |
+| 404 | `no-trace`, `unsupported-harness`, `trace-not-user-conversation` | Expected Reader availability, not a failed fetch substituted with empty content |
+| 409 | `conflict`, `not-archived`, `input-not-ready`, `invalid-workspace`, `remote-agent`, `cron-unavailable`, `backup-running`, `no-transcript` | Action cannot run in the current state; not permission to replay it |
+| 409 | `file-changed` | Existing `mtime` preserved; browser also retains `tag`/`currentTag` when supplied by file-save endpoints |
+| 409 | `redaction-blocked` | Existing `hits` rule/count map preserved |
+| 409 | `conflict` on paused remote messages | Existing `stop` and `reason` retained, plus additive `error`/`code` |
+| 413 / 415 / 429 | `payload-too-large` / `unsupported-media-type` / `rate-limited` | Body limits, media/encoding policy, existing rate limits |
+| 5xx | `internal-error` | Generic safe message; optional bounded `requestId`; a settings write may name its validated data-root-relative file in `details: [{field:"path",message}]` |
+
+`reason`, `hint`, `hits`, `mtime`, `tag`, `currentTag`, `retryAfter`, and
+`details` are optional, not universally present. A 5xx never carries an absolute
+path or exception text; the settings filename is relative to the durable data
+root (for example, `am-config.json`). The browser retains these
+allowlisted fields in `ApiError.data`, with bounded strings/collections. It
+retains `status` even for an unreadable proxy response. No automatic retry,
+new timeout policy, mutation replay or delivery guarantee is added.
+
+The audit recorder keeps one outcome. If the connection closes before the
+server finishes responding, `ok:false` and `incomplete:true` distinguish it from
+success; `status:499` is audit-only when no headers were committed. This is not
+evidence that the operation was rolled back or that a replay would be safe.
+Request-body capture and filtering are unchanged.
+
+## Route / response / mutation inventory
+
+Paths below are relative to `/api`. The handler inventory was checked against
+main `ef08e843d1c67fcc4e1c463415819ea351734d6d`, rather than relying on planning
+line numbers. `apiRoutes` covers every API registration in `index.js`, including
+the generated `mkdir`/`touch` routes. Global JSON parsing, privacy lock and
+operation attribution stay in their original order; validation runs after
+route-specific parsers and before the final action. The separate browser-origin
+and audit-filter work is not reimplemented here.
+
+| Routes | Forms and responses | Mutation / compatibility notes |
+| --- | --- | --- |
+| GET `health`, `info`, `visibility`, `clis`, `tree`, `sessions`, `next-name` | JSON | Health/public lock exceptions unchanged; next-name validates CLI |
+| GET `usage`, `operations`, `traces`, `meta`, `meta/:id` | JSON | Provider enum; operation limit 1–1000, existing timestamp cursor; optional provider/digest subsystem fallbacks remain local |
+| POST `push/subscribe`, `push/unsubscribe`, `notify` | JSON command | Subscription structure, endpoint and notification fields validated before dispatch; notification zero-device/per-device partial outcomes remain **200 `ok:false`** |
+| POST `sessions` | JSON creation, 201 | CLI required; optional name/prompt/path; blank name retains generated default, blank path retains workspace root; group must exist; optional prompt still starts asynchronous work |
+| PUT `sessions/:id` | JSON selected-field update | Display name only, no folder rename |
+| POST `sessions/:id/input`, `attachments/insert` | JSON command | Text / attachment ID types checked before resolution or agent input; existing attachment limits and insertion acknowledgement unchanged |
+| POST `sessions/:id/attachments` | Raw bytes, 201 JSON metadata | Any file media type, including JSON, bypasses JSON parser; existing quotas, image validation and cancellation retained |
+| DELETE `sessions/:id/attachments/:attachmentId`; GET `…/raw` | JSON command; binary stream | Stream errors before headers are JSON, after headers terminate; disconnect destroys file source |
+| GET `agents`, `agents/:id`, `…/tail`, `…/subagents`, `…/subagents/:agentId` | JSON | Tail line integer validation with existing 2000-line clamp (including clients requesting more), trace query validation; roster/trace distinctions retained |
+| GET `agents/:id/wait` | JSON long poll | State list, timeout 1–300 s, settle 0–60 s; existing wait/deadline behavior retained, disconnect stops polling |
+| POST `agents`, `agents/:id/prompt`, `agents/:id/stop` | Text body **or** JSON `prompt` / `text`; `from` body/query, launch CLI/name/path/group query aliases | Preserve curl's non-JSON default content type as literal prompt text, not form decoding. Sender, target, installed CLI, self-target and group resolution checks remain before action |
+| GET `remote/:name/ping`, `…/messages`, `sessions/:id/remote`; POST `remote/:name/hello`, `sessions/:id/remote/paused` | JSON | Remote peer strings/paused boolean; scalar sequence cursor; existing remote-origin fallback unchanged |
+| POST `remote/:name/messages`; GET `…/prompt`, `…/stream` | Text/JSON message command; plain-text instructions; heartbeat/NDJSON long poll | Paused poll's **200 stop result** is deliberate; paused message is 409. Stream sends connected/heartbeat/one JSON line, never JSON error appended after bytes. Wait remains 5–1800 s, no global short timeout |
+| POST `demo`, `welcome/seen`, `relaunch`, `update`; GET `update/check` | JSON commands/status | Demo boolean. `no-space`, `no-token`, `cooldown`, `busy`, `upstream-unreachable` and upstream refusal capability results remain **200 `ok:false`**; thrown request failures now return safe 500 |
+| GET/PUT `config` | JSON **replacement** | Omitted sections/fields keep existing replacement defaults; booleans are not coerced; explicit zero, false and backup `exclude:[]` preserved. No durable-store/recovery rewrite |
+| GET/PUT `secrets` | JSON notes-map replacement | Notes contain arbitrary user-defined keys with string values, including empty strings; no global field stripping |
+| GET `backup/status`; POST `backup/run` | JSON status / starts asynchronous Job | Missing prerequisites, already-running and bad IDs explicitly classified; status subsystem's optional lookup fallbacks unchanged; no Job lifecycle redesign |
+| GET `skills`, `skills/:name`; PUT/DELETE `skills/:name` | JSON listing/read; text replacement / delete | PUT must decode to a string, including valid empty text; JSON objects cannot silently overwrite a skill with empty text; distribution remains best effort |
+| GET `files/:id`, `…/preview`, `folders` | JSON | Scalar paths; existing lexical/realpath, file-kind and size checks retained |
+| GET `files/:id/raw`, `…/download`, `trace/:id/download` | Binary/download stream | Existing sandbox/content-disposition headers; explicit callback forwarding; no JSON appended after headers |
+| PUT `files/:id/write` | Text replacement, JSON result | Valid empty text allowed; existing `base` revision precondition, text-size and path checks retained; JSON objects rejected before writing |
+| POST `files/:id/mkdir`, `…/touch`, `…/rename`, `…/move`; DELETE `…/entry` | JSON commands | Single filename / destination types validated before filesystem changes; root/dependency/existing-entry checks retained |
+| POST `files/:id/upload` | Raw bytes → JSON completion | Raw JSON files also bypass body parsing. One completion/error path; abort unpipes/closes destination and releases listeners. Existing overwrite policy unchanged (separate file-protection work) |
+| POST `overview/hidden`, `move`; POST/PUT/DELETE `groups[/id]` | JSON | Hide boolean/ref; live move targets; group PUT is a selected-field patch, not replacement. `layout:null` means auto; explicit layout is 1–3 rows/columns; session ordering retains live membership checks |
+| GET/POST `crons`; PUT/DELETE `crons/:id`; POST `…/run` | JSON; creation 201, manual fire 202 | Create requires name/agent/prompt/five-field schedule/time zone. PUT patches selected fields, nested agent/schedule objects still replace as before. Strict runOnRestart/state types. 202 means accepted, not finished; later failure stays in `last` with safe text. No target identity/scheduling behavior change |
+| GET/POST `sessions/:id/share`; GET/POST `share/access` | JSON | Visibility/name/user lists validated; redaction 409 and successful share's `granted`/`grantErrors` retained. Partial grant errors get a safe message/code; best-effort access-list lookups remain unchanged |
+| POST `trace/import`; GET `trace/bundles` | JSON | Bare repo or pasted dataset URL still supported; known domain codes kept. No safe re-import/transaction changes |
+| GET `trace/:id`, `…/location`, `files/:id/trace`; PUT `trace/:id/source` | JSON trace pages/windows/summary; selected source update | Trace modes mutually exclusive; safe nonnegative integer cursors, v1/v2; existing page/window-size clamps retained (500 turns / 8 MiB). Source kind enum, ref/target checks before storing |
+| POST `sessions/:id/archive`, `…/unarchive`, `…/stop`; DELETE `sessions/:id` | JSON commands | Existing archive/delete and `ifNeverStarted=1` rules unchanged; optional attachment cleanup remains isolated |
+| `/ws` | WebSocket terminal protocol | Not JSON wrapped; framing, origin admission and terminal lifecycle unchanged |
+| Other `/api` paths; static assets and SPA paths | API 404 JSON; ordinary static/SPA responses | API errors never fall through to an HTML application page; non-API failures remain Express/static concerns |
+
+## Validator conventions
+
+- Validate at the route boundary before any domain action, not only in TS.
+  Schema helpers distinguish missing values from null/empty/false; no generic
+  truthiness or numeric-string conversion for JSON. JSON-only payloads require
+  `application/json`; bodyless commands may omit it. Malformed JSON, unsupported
+  encodings and parser limits get deliberate JSON errors. Express's strict
+  parser rejects scalar JSON documents (including bare `null`).
+- Queries are scalar strings: duplicate/bracket-structured parameters are
+  rejected. Numeric query fields use complete integer syntax, safe integer
+  bounds and route limits. Existing documented clamping of trace size budgets
+  stays in the domain reader. Paths reject null characters and retain existing
+  containment checks; do not introduce a second path resolver.
+- Unknown fields retain each endpoint's existing handling, usually ignored.
+  No input object is stripped or rewritten by validation. Notes-map extension
+  keys remain supported. Intentional empty group names on creation, empty text
+  file replacement and null group layout keep their special meaning.
+- Schedule validation is reused, with a public validation error instead of
+  arbitrary parser exception text. Store schemas/durability, scheduler identity,
+  import replacement, async work receipts and delivery retries are separate work.
+
+## Adding a handler and client
+
+```js
+// server/src/index.js — register via api, not app, and add its input case to
+// createValidator. Use ApiError only for a message known to be public.
+api.post('/api/example', async (req, res) => {
+  if (!available()) throw new ApiError(409, 'example-unavailable', 'Try after the current task finishes.');
+  res.json(await performValidatedAction(req.body));
+});
+```
+
+Promise rejections and synchronous throws enter one Express error path. For
+event emitters, explicitly forward errors and remove timers/listeners on close;
+the promise wrapper cannot catch unrelated future callbacks. Downloads pass
+their callback error to `next`; `pipeResponse` and `remoteStream` demonstrate
+the before/after-headers distinction. A late failure after a completed response
+is consumed, not sent again. Do not change status or append JSON to committed
+stream bytes.
+
+```ts
+// web/src/api.ts — the existing local fetch keeps attribution; json uses the
+// single decoder. For a text success, use decodeResponse(r, 'text').
+export const example = (body: ExampleInput) =>
+  fetch('/api/example', { method: 'POST', headers: HEADERS, body: JSON.stringify(body) }).then(json);
+// UI: catch ApiError; display .message, branch on .code, inspect .data.
+// Keep the draft and existing explicit failure actions; never replay here.
+```
+
+`decodeResponse` reads a failure body once, capped at 64 KiB, cancels oversized
+reads, and never shows a proxy's HTML/status text. Empty success is allowed;
+malformed expected-JSON success is `unreadable-response`. Legacy `{error:string}`
+still works. AbortError identity remains intact for Reader cancellation;
+`timeout`, `offline`, and `network-error` are distinct transport failures with
+no HTTP status unless headers were already received. XHR remains the upload
+transport: same decoder, progress, explicit cancellation and upload deadline.
+TraceUnavailable and RedactionBlocked remain subclasses with their existing
+constructors/fields; generic HTTP failures are not Reader empty states.
+
+Tests: `server/test/api-boundary.test.mjs` exercises the real Express stack with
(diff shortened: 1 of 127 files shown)


lvwerra (commented):
**Round 1 review: no blocking defects found.** I could not break the admission policy, and the compatibility claims hold up in a real browser. Three non-blocking observations below, ranked; none of them needs to hold the merge, but the first is a product decision worth making deliberately.

Per SECURITY.md this comment describes classes of behaviour only — no reproductions, payloads or deployment identifiers.

## What I checked, and how

Checked out `be62f4d` and ran it, rather than reading the diff alone.

**Compatibility, in a real browser against the real server** (headless Chromium, built frontend, isolated state/home roots, synthetic local origins):

- Standalone: every API request the app itself makes carries the intent marker; a protected read succeeds through the app's own client and is refused from the same document without the marker.
- **Embedded in a cross-site parent frame: identical results, and the terminal WebSocket opens.** This is the case that would have been worse to break than the gap being closed, so I tested it as a real iframe with a genuinely cross-site ancestor rather than trusting the header matrix.
- I also captured what a current browser actually sends on a WebSocket handshake before reasoning about it — see observation 2.

**Mutation checks.** I reconstructed five of the eight claimed mutations and reverted each in a scratch copy: exact-origin enforcement, the required marker, protected-read classification, the central HTTP wiring, and pre-upgrade admission. Every one is caught (2, 5, 2, 1 and 1 failing suites respectively; the central-wiring mutation is caught only by the real-backend suite, which is the right place for it). The claim that the tests fail when the fix is removed is accurate.

**Attempted bypasses that did not work:**

- *Path spelling against the protected-read matcher.* This matcher is the only path-dependent part of the guard — every non-GET/HEAD method is admitted-or-refused regardless of path — so it is the whole surface for this class. I probed a range of alternative spellings (separator, case, dot-segment, encoding, parameter and whitespace variants) against a protected read with no marker. The only non-404 answer is the SPA fallback serving `index.html`; no spelling reaches a real handler unguarded. I initially misread that fallback as a bypass and confirmed it against the handler's own response before concluding otherwise.
- *Effect behind a read.* I walked the GET inventory for handlers with side effects. The classification looks right to me: the remote contact/delivery routes are covered, and the connectivity-check route is correctly excluded because it deliberately does not stamp liveness. `next-name` and the digest routes compute from existing state and reserve nothing.
- *Test flag leaking into the guard.* The pre-existing audit-attribution relaxation does not relax admission — confirmed live, not just by reading, since "security tests must run with the real guard enabled" is an explicit requirement.

**Migration coverage** is complete as far as I can find: docs, the generated remote instructions, the environment skill, the internal scheduler request and the bundled fixtures are all updated, the last via a shared native client and a shell wrapper rather than per-call edits. No bundled hook or helper script makes HTTP to the manager, which matches the inventory claim in the PR body.

**Suites I ran:** the new policy/HTTP/API suites; `npm test` and `tsc --noEmit` in web (25 suites); and 19 server suites via the runner. All pass. `crons.test.mjs` #4 is the known baseline failure and is unrelated.

## Observations

**1 — Bounded reconnects turn an ordinary backend restart into a manual reconnect on every open pane.** After four attempts (about eleven seconds) a pane goes to the paused state and needs a click. The previous behaviour backed off to a 15s ceiling and recovered on its own. Bounding the retries is what the issue asked for, and pausing is clearly right for a refusal — but a restart or a sleeping Space is a normal event, and with several panes open that is now a click each. The client cannot tell the two apart because browsers report both as a 1006 close. It could: on exhausting the budget, make one ordinary marked request to a cheap endpoint, and treat an answer as "the server is fine, this is a refusal, stay paused" versus no answer as "keep backing off". Worth a deliberate decision rather than leaving it as a side effect of the bound.

**2 — The Fetch Metadata destination check has no WebSocket branch, though the mode check does.** I measured what a current browser sends on a same-origin WebSocket handshake: no Fetch Metadata headers at all, so nothing breaks today and the policy unit test's synthesized handshake matches reality by omitting them. But the destination rule admits exactly one value, and a browser that begins sending a websocket destination on handshakes would have every terminal connection refused, with the failure surfacing as the paused state from observation 1. It is a one-value change to make that forward-safe, and the asymmetry with the mode check looks unintentional.

**3 — The production edge assumption is the one thing neither of us can test here, and it fails closed for everything at once.** Writes and terminals now both depend on the deployment presenting a target host in the configured set and the page origin matching the configured app origin exactly. The PR is honest that no production edge behaviour was live-tested. If either differs from the fixture assumption, the symptom is not partial degradation — it is every state change and every terminal refused simultaneously. That argues for confirming it against the real deployment before or during rollout rather than after, and for making sure the refusal reason code is visible enough at that moment to tell "origin mismatch" from "host mismatch" quickly. Not a code change; a rollout note.

## Scope

I found nothing implemented from the excluded list — no CSP work, no quotas, no accounts or RBAC, no route extraction, no sharing or visibility changes. The frontend and server surfaces stay inside the stated boundary, and the deliberate exclusions in the PR body match what the issue excluded. The required documentation table, ordering description and route inventory are all present.

lvwerra:
Round-one follow-up, pushed to the same branch at `6c641c6` ([review delta](https://github.com/huggingface/agent-manager/compare/be62f4d...6c641c6)). All three observations are addressed:

1. **Product decision: retain bounded terminal retries for now.** I agree this has a real usability cost: after four automatic retries, a longer restart or sleeping Space requires a manual retry in each affected pane. That is now documented explicitly, and the misleading source comment promising automatic recovery from every transient drop is corrected. I have not added the suggested HTTP health probe: reachability of HTTP does not reliably classify failure of the separate upgrade path, and an unanswered probe does not establish that further upgrade retries will help. Keeping the deterministic bound avoids adding another transport-dependent retry policy here. This is a deliberate tradeoff, not a claim that automatic outage recovery is impossible or undesirable. Please put this choice and its per-pane cost in front of the operator as planned. Existing output/draft preservation and the explicit retry action are unchanged.

2. **Fixed the WebSocket Fetch Metadata destination asymmetry.** Upgrades now accept the WebSocket destination as well as the existing absent/empty cases. HTTP actions still reject that destination; unfamiliar metadata and all other admission requirements remain unchanged. Added a policy matrix and real HTTP/upgrade regression coverage, including rejection before parsing/attachment and continued usability of an admitted viewer. This addresses forward compatibility, not a claim that the currently tested browser sends those headers.

3. **Added an [authorized-rollout checklist](https://github.com/huggingface/agent-manager/blob/6c641c6/docs/browser-request-protection.md#authorized-rollout-checks).** The deployment owner must verify actual upstream target identity and exact app page origin, then confirm both standalone and genuinely embedded UI actions and terminals. The checklist explains where the existing bounded refusal reasons distinguish target from origin configuration, including the limits of browser WebSocket diagnostics. Stop rollout if either context fails; resolve trusted configuration without adding a bypass. Production edge behavior is still unverified here. No deployment, live probing, installed-client changes or operator messages were performed.

Verification this round (isolated state/home roots, synthetic local origins; Node 22.23.2 and Chromium 153.0.8010.12):

- New policy/HTTP tests: both additions failed before the fix; all nine checks pass with it. Three mutation checks were caught: reverting the fix (two failures), incorrectly extending it to HTTP (two), and removing destination validation (four). All mutations restored.
- Web `npm test`: all 25 suites pass, including actual embedded/proxy/browser request paths and bounded TerminalPane retry/Reader draft coverage. `npm run build` and all four `npm run test:render` suites pass.
- Server `npm test`: six discovered suites pass, then the already-known main-red cron test #4 stops the runner. Continuing the remaining suites separately covers all 30 other default suites successfully; the continuation filter repeats one earlier suite. All three `npm run test:ui` suites pass.
- The previously documented mobile baseline failure was not changed or reclassified; the manual mobile suite was not rerun in this round.

The full suites ran on `b046185`; `6c641c6` changes only documentation wording. Final focused server and browser runs using `node ../scripts/run-suites.mjs request` pass on `6c641c6` (three server suites, two browser suites). The nine policy/HTTP checks can also be run directly from server with `node --test test/request-admission.test.mjs test/request-http.test.mjs`.

Server continuation command:

```sh
node ../scripts/run-suites.mjs first-run hidden input-required next-name opencode-resume operations queued-prompts reader-protocol repin request revive slowfs spawn-group subagent-traces terminal-modes trace-download trace-tail trace-window usage migration resize state-checkpoint
```

No broader policy or excluded feature work was added. Ready for the required second independent review; not merged or deployed.

lvwerra (commented):
**Round 2 review: no blocking defects, and nothing non-blocking worth holding either.** I went after ground round one did not cover rather than repeating it: the delta it could not see, origin *representations* (round one probed path spellings), and the destructive/lossy path end to end. Everything I tried failed closed.

Per SECURITY.md this describes classes of behaviour only — no reproductions, payloads or deployment identifiers. Checked out `6c641c6` and ran it against a real server with isolated state and synthetic local origins.

## The delta round one could not review

The one substantive change since `be62f4d` is the WebSocket destination branch. I reconstructed the three mutations the author claims and one more, and every count matches:

| Mutation | Failing suites |
| --- | --- |
| Revert the fix (WS destination rejected again) | 2 |
| Extend the allowance to HTTP as well | 2 |
| Remove destination validation entirely | 4 |
| Remove mode validation *(mine, not claimed)* | 2 |

The change only widens what an *upgrade* may present, and an upgrade from a disallowed origin is still refused before the destination is ever consulted — I confirmed that directly rather than reasoning about ordering. HTTP still rejects that destination.

## Origin representations

Round one covered path spellings against the protected-read matcher and noted that is the only path-dependent surface. The other half is how an origin can be *spelled*, which the issue also requires. I ran the full matrix against the real policy with a configured production-shaped deployment:

- **Admitted, correctly:** the exact origin, an uppercased host, an explicit default port, and a zero-padded port that normalises to the same authority. Those are the same origin and admitting them is right.
- **Refused:** trailing dot, trailing slash, wrong scheme, wrong port, subdomain, suffix- and prefix-shaped lookalikes, embedded userinfo, `null`, empty, and whitespace-padded.
- **Refused as ambiguous:** duplicate Origin, duplicate Host and duplicate marker headers, detected off the raw header list rather than the joined value.
- **Target host:** a wrong host and a right host on a wrong port are both refused. The loopback upstream authority is an accepted *target* but grants nothing on its own — an evil origin against it is still refused, which is what the comment in the module promises.

## Allowlist cannot be widened by a request

I tried both directions — forwarded/original-host variants to introduce a new host, and the same headers to rescue a request whose real Host was wrong. All refused. A request with no Host header at all is refused rather than defaulting. The policy reads the raw `host` header, not Express's proxy-aware accessor, so a `trust proxy` setting could not change this either.

## The destructive path

This is what I was asked to break, and I could not.

- **Irreversible session actions from a disallowed origin** — deletion, archiving and stopping — are all refused, the target survives, and **nothing is written to the operation log**. I checked the log rather than only the response, because an unreadable response is not proof of no effect.
- **A large streaming upload from a disallowed origin** is refused with no file created anywhere under the state root. The guard sits ahead of body handling, so the bytes are never stored.
- **HEAD, trailing-slash and case variants** of every protected read are refused exactly as the GET is, while the same reads succeed for a properly marked client.
- **OPTIONS** answers 204, returns zero CORS headers, and leaves state untouched.
- **A form-shaped simple request** (no marker possible, cross-site metadata) is refused.

## Terminals, including the case the issue calls out by name

I set up a genuine viewer attached over the allowed origin, then fired a burst of refused upgrades at the same session: a cross-site origin, an opaque origin, a native handshake with no proof, a duplicate Origin, and one presenting the newly allowed WebSocket destination from a bad origin. Every one was refused before upgrade with **zero frames delivered**, and the live viewer's frame count, connection and control were unchanged throughout. A native handshake carrying the marker still connects. The terminal ownership and geometry suites pass alongside.

## Frontend request-path inventory

I re-derived it rather than trusting the claim. Every API call funnels through the shadowed wrapper in `api.ts`; the attachment XHR sets the same header explicitly; the marker is applied to safe methods too, which is what makes the protected reads work. I also checked the **service worker**, since it is a request path entirely outside that module — it registers push and notification-click handlers only, has no fetch handler and no caching, so there is nothing there to miss. That matches the PR's inventory claim.

## The retry change

The paused-state retry calls the reconnect path *without* the terminal reset, so retained output survives; only the exited-state control passes the reset flag. The failure counter resets on a real grid frame rather than on socket open, so a socket that is admitted and immediately dropped does not refill the budget. The bound can only reduce reconnect attempts relative to main, so it cannot newly restart a stopped agent. Round one's usability observation stands as a product decision the author has documented and escalated; I have nothing to add to it.

## What I did not independently redo

Round one built a real cross-site iframe and confirmed the embedded UI and its terminal both work. I did not rebuild that fixture; I ran the checked-in web suite, which includes the embedded/proxy/browser request paths, and it passes 25/25 with the production build and typecheck clean. The production edge assumption remains untestable here, as both the PR and round one say; the added rollout checklist is the right place for it.

## Scope

Nothing from the excluded list appears. The `remote.js` change is generated-prompt text plus migration guidance, including the instruction not to auto-replay an uncertain write. Documentation carries the policy table, ordering description and route inventory.

## Commands run

`node --test test/request-admission.test.mjs test/request-http.test.mjs` (9 pass) · `node test/request-api.test.mjs` (pass) · `node ../scripts/run-suites.mjs request operations terminal-modes resize revive` (7 suites) · `npm test` and `npm run build` in web (25 suites, build and typecheck clean). `readerRedesign.test.mjs` passed on its normal run, so no isolation re-run was needed. `crons.test.mjs` #4 is the known main baseline and I did not count it.

PR #137, opened 2026-09-09
Refresh session state on browser return

### LINKED ISSUE #124: Refresh session lists and status on return and reconnect
# Refresh session lists and status on return and reconnect

## Agreed scope

Implement one focused improvement: automatically refresh the session list and Overview/sidebar status when the operator returns to Agent Manager or the browser reconnects. The operator explicitly narrowed the broader connection-feedback proposal to this behavior. Do not add stale/reconnecting indicators, last-updated labels, a Retry button, or a general polling redesign.

This is a targeted freshness improvement, not a claim that every earlier reader problem has the same cause. Reader transcript recovery was already implemented in [PR #119](https://github.com/huggingface/agent-manager/pull/119).

## Current behavior and code pointers

- [App.tsx tree and metadata loops](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/App.tsx#L402): the tree already refreshes on visible `visibilitychange` and polls every 2.5 seconds. Metadata has no corresponding immediate foreground refresh; its timer uses 1.5 seconds in Overview and 8 seconds otherwise. Both loops skip regular reads while hidden. Neither loop explicitly handles browser `online` or back/forward-cache restoration.
- [API read helpers](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/api.ts): `getTree` and `getMeta` currently have no optional cancellation signal. App refreshes can overlap; late results have no request-generation check.
- [Existing reader lifecycle](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/lib/traceWindows.ts) and [reader request handling](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/lib/readerStore.ts) already cover visibility, `online`, persisted `pageshow`, cancellation and bounded reads. Preserve that behavior; do not add redundant app-triggered transcript or whole-file-summary reads.

Recheck main and open PR overlap before implementing. Source inspection establishes the missing triggers, not a reproduction of a specific production incident.

## Requirements

1. Refresh both `/api/tree` and `/api/meta` promptly on returning from a hidden tab, returning focus to a visible app window, restoring a persisted page, and receiving `online` while visible. Do not wait for the next ordinary polling interval or require a reload/navigation click. Ordinary initial loading must still work without a duplicate initial `pageshow` refresh.
2. Coalesce the related events from a single return/reconnect so they do not launch repeated requests or continually cancel each other. Start the first fresh read promptly. Events received while hidden must not restart background polling; catch up on the next visible return. Treat `online` as a refresh hint, not proof that the server is reachable.
3. A pre-return request must not block a fresh read indefinitely or later replace its result. Add only the local cancellation/request-generation/deadline handling needed to make these two resources reliable, including their existing poll and post-mutation refresh paths. If a mock/transport ignores abort, an obsolete request must still lose permission to publish and cannot retain the refresh slot forever. Do not introduce an app-wide request framework.
4. Keep the last successfully loaded list/cards/status visible during these reads and failures. Update each resource independently when it succeeds; one failed resource must not discard the other's result. Preserve selected session, open panes, drafts and scroll without remounting the app. Existing reconciliation for a session genuinely absent from a successful fresh tree must continue to work.
5. Preserve ordinary polling cadence, hidden-tab behavior, metadata payload-diffing and existing mutation-triggered tree refreshes. A failed return refresh must not permanently disable subsequent polling/recovery. Clean up listeners, timers and outstanding publication rights on unmount; remounts must not accumulate extra loops.
6. These are read-only refreshes. Do not reload the page, submit prompts, create/restart sessions, claim terminal control, or change PTY dimensions merely because the browser returns or reconnects.

## Required tests

Use controllable clocks and deferred/mock responses for races, plus browser coverage exercising the actual app wiring. Keep the tests in the normal discovered suite.

| Scenario | Required assertion |
| --- | --- |
| Hidden tab returns | Change tree and metadata fixtures while hidden. Returning starts both fresh reads without advancing the regular poll timer, in Overview and a session view; both update without reload. Hidden time does not add polling work. |
| Other return/reconnect paths | Visible window focus, persisted `pageshow`, and offline-to-online each refresh both resources. Initial mount and a burst of related events do not multiply requests. Online while hidden defers work until return. |
| Frozen or obsolete reads | Leave a pre-return read unresolved, including a mock that ignores abort; return still obtains new data. Resolve the old read after the fresh response: it cannot overwrite either resource. Exercise interaction with an ordinary poll and a post-mutation tree refresh. |
| Partial or continued failure | Tree succeeds/meta fails and vice versa; retain last-good content and accept independent success. A failed recovery read is followed by successful ordinary polling or a later return event; no permanently occupied request slot. |
| Lifecycle and existing behavior | Mount/unmount/remount, rapid view changes and repeated return cycles leave one intended set of listeners/timers. Existing polling rates, payload-diffing, selection reconciliation and post-mutation updates still work. |
| UI and reader integration | Keep selection, an unsent prompt/file draft, open panes and scroll through refresh. An active reader retains its existing recovery path; returning does not add duplicate app-level transcript reads, create a session or send a prompt/PTY action. |

Browser lifecycle events may need simulation in headless tests. State that limitation explicitly; exercise the real registered handlers, not a separate test-only implementation. Include a manual tab/app-switch and reconnect smoke check when a browser environment supports it.

## Exclusions and acceptance

- No connection banner, freshness label, Retry control, new status vocabulary, global retry/backoff service, server push/WebSocket migration, or refresh migration of unrelated panes/settings.
- No reader redesign or new transcript caching/parsing work; preserve #119 and its regression coverage.
- Use fixtures and disposable state only. Do not interact with live sessions, alter production configuration, or deploy/restart the app to test this issue.
- Run focused lifecycle/race tests, the default web suite and production web build/typecheck. Include exact commands/results; identify unrelated baseline failures separately. Run server tests if server code is changed, although no server change is expected.
- The PR is complete when returning/reconnecting refreshes both app resources promptly and safely, without added UI, lost local state or recurring extra network work. Document the small lifecycle contract and open one focused PR referencing this issue; merge/deployment require separate authorization.

Review coverage: the return/reconnect slice of B14/C06/U03 only. Broader freshness UI and general polling work remain unapproved.

lvwerra:
Closes #124

## What changed

- Refreshes both `/api/tree` and `/api/meta` immediately when a hidden tab becomes visible, a blurred visible window regains focus, a persisted page is shown, or the browser comes online while visible.
- Coalesces the companion events from one return, including browsers that report focus before visibility, while a hidden `online` event waits for the visible return.
- Gives tree and metadata separate latest-read managers. Foreground and post-mutation reads abort/retire old work; generation checks prevent late publication; a 12-second local deadline releases the slot even when a transport ignores abort. Ordinary polls remain single-flight.
- Keeps the existing 2.5-second tree and 1.5/8-second metadata cadence, hidden-tab guard, metadata payload diffing, tree reconciliation, and mutation refresh behavior. Failed reads preserve each resource's last good value independently.
- Threads optional abort signals only through `getTree` and `getMeta`; lifecycle cleanup retires listeners, timers, and outstanding publication rights.

## Lifecycle contract

Initial tree/meta reads are the only startup reads; ordinary non-persisted `pageshow` and startup focus noise do not duplicate them. Polls coalesce behind an in-flight read. A real foreground/reconnect event starts replacement reads promptly, and the first visible event consumes a hidden return so later companion events cannot duplicate it. Reader lifecycle remains owned by `traceWindows`/`readerStore` from #119.

## Agreed boundary

Implemented only app-level tree/list and metadata/status freshness. Deliberately left out stale/reconnecting UI, last-updated labels, Retry controls, polling redesign/backoff, server push, unrelated pane/settings refreshes, page reloads, reader transcript changes, and any prompt/session/trace/PTY action.

Open PRs #118 and #120 also touch `App.tsx` (and #118 touches `api.ts`); this branch does not incorporate their unrelated work and may need a mechanical rebase if either lands first.

## Verification

- `cd web && node test/appRefresh.test.mjs` — passed. Uses controllable clocks/deferred reads plus the real App wiring in Chromium. Covers all four return hints, hidden deferral, event bursts/order, stale abort-ignoring reads, deadline release, tree/meta partial failure and recovery, ordinary polls, a real post-rename refresh, unmount/remount, Overview and session views, retained drafts/panes/scroll, and the absence of non-read side effects.
- `cd web && npm test` — 24/24 discovered suites passed.
- `cd web && npm run typecheck` — passed.
- `cd web && npm run build` — passed (existing chunk-size advisory only).
- Mutation check: making explicit replacement wait behind the old request fails `foreground replacement does not wait for an old read` (`1 !== 2`). Restoring replacement semantics passes.
- Mutation check: removing the App return observer fails the Chromium immediate-return assertion (tree remains at the pre-return request count instead of incrementing with metadata). Restoring it passes.

Headless Chromium cannot force an OS-level tab/app switch or a real BFCache entry. The browser test therefore dispatches visibility/focus and persisted `PageTransitionEvent` through the production listeners (not a test-only refresh path). Reconnect is additionally exercised with Playwright's actual browser-context offline-to-online transition and passed. No live sessions or deployment were used.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -22,6 +22,7 @@ import { hiddenSessionIds } from './lib/overviewHidden';
 import { paneOwnsBack } from './lib/mobileBack';
 import { isPassive, isRemote, isShareable } from './types';
 import { EyeGlyph, EyeOffGlyph, GridGlyph, ListGlyph, SortGlyph } from './components/icons';
+import { createLatestRefresh, observeAppReturns } from './lib/appRefresh';
 
 // `?vvdebug=1` — a phone has no devtools, and the keyboard layout is a guess
 // when the app is embedded cross-origin. Read once: it never changes mid-run,
@@ -399,8 +400,11 @@ export default function App() {
   // treeLoaded gates the selection check below: until the first tree actually
   // arrives, "your agent isn't in this tree" only means the tree is still empty.
   const [treeLoaded, setTreeLoaded] = useState(false);
-  const refresh = useCallback(async () => {
-    try { setTree(await api.getTree()); setTreeLoaded(true); } catch { /* offline */ }
+  const treeRefresh = useRef<ReturnType<typeof createLatestRefresh<Tree>> | null>(null);
+  // Mutations and foreground recovery replace any possibly frozen request.
+  // Ordinary polls use the manager directly below and coalesce instead.
+  const refresh = useCallback((): Promise<Tree | null> => {
+    return treeRefresh.current?.refresh('replace') ?? Promise.resolve(null);
   }, []);
 
   // Hide/unhide a group (or one agent) in the Overview. Optimistic, because the
@@ -417,13 +421,21 @@ export default function App() {
 
   useEffect(() => {
     api.getClis().then(setClis).catch(() => {});
-    refresh();
-    // Skip polling while the tab is hidden; catch up immediately on return.
-    const t = setInterval(() => { if (!document.hidden) refresh(); }, 2500);
-    const onVisible = () => { if (!document.hidden) refresh(); };
-    document.addEventListener('visibilitychange', onVisible);
-    return () => { clearInterval(t); document.removeEventListener('visibilitychange', onVisible); };
-  }, [refresh]);
+    const manager = createLatestRefresh(
+      (signal) => api.getTree(signal),
+      (next) => { setTree(next); setTreeLoaded(true); },
+    );
+    treeRefresh.current = manager;
+    void manager.refresh('replace');
+    // Same cadence and hidden-tab rule as before. Foreground catch-up is shared
+    // with metadata below so visibility/focus/pageshow/online cannot drift.
+    const t = setInterval(() => { if (!document.hidden) void manager.refresh('poll'); }, 2500);
+    return () => {
+      clearInterval(t);
+      if (treeRefresh.current === manager) treeRefresh.current = null;
+      manager.dispose();
+    };
+  }, []);
 
   // Live per-session digests, polled CONTINUOUSLY in the background (not only
   // while the Overview is open) so opening it is instant. Faster cadence when
@@ -435,12 +447,15 @@ export default function App() {
   const [ages, setAges] = useState<Record<string, number>>({});
   const overviewActiveRef = useRef(false);
   overviewActiveRef.current = activeRef === 'overview';
+  const metaRefresh = useRef<ReturnType<typeof createLatestRefresh<Awaited<ReturnType<typeof api.getMeta>>>> | null>(null);
+  const refreshMeta = useCallback(() => {
+    return metaRefresh.current?.refresh('replace') ?? Promise.resolve(null);
+  }, []);
   useEffect(() => {
-    let alive = true;
     let lastPayload = '';
-    const load = () => api.getMeta()
-      .then((r) => {
-        if (!alive) return;
+    const manager = createLatestRefresh(
+      (signal) => api.getMeta(signal),
+      (r) => {
         setMetaReady(true);
         const payload = JSON.stringify(r.sessions);
         if (payload === lastPayload) return;
@@ -449,19 +464,29 @@ export default function App() {
         setAges(Object.fromEntries(r.sessions.map((s) => [
           s.id, Math.max(s.digest?.lastAssistantTs || 0, s.digest?.lastPromptTs || 0),
         ])));
-      })
-      .catch(() => {});
-    load();
+      },
+    );
+    metaRefresh.current = manager;
+    void manager.refresh('replace');
     // One self-scheduling timer whose delay adapts to the active view, so we
     // never tear down / recreate the loop when navigating.
     let t: ReturnType<typeof setTimeout>;
     const tick = () => {
-      if (!document.hidden) load();
+      if (!document.hidden) void manager.refresh('poll');
       t = setTimeout(tick, overviewActiveRef.current ? 1500 : 8000);
     };
     t = setTimeout(tick, 1500);
-    return () => { alive = false; clearTimeout(t); };
+    return () => {
+      clearTimeout(t);
+      if (metaRefresh.current === manager) metaRefresh.current = null;
+      manager.dispose();
+    };
   }, []);
+
+  // Browser returns arrive as different (often clustered) event shapes across
+  // desktop, mobile and bfcache restoration. Refresh both independent resources
+  // once; the reader owns its own equivalent lifecycle and is not touched here.
+  useEffect(() => observeAppReturns(() => Promise.allSettled([refresh(), refreshMeta()])), [refresh, refreshMeta]);
 
   // Archive threshold from the operator config; refresh when settings closes
   // (that's where it's edited).
@@ -724,9 +749,11 @@ export default function App() {
   // leaving you on a single view of a session that has moved.
   const doMove = (ref: string, to: MoveTarget) => api.move(ref, to)
     .then(async () => {
-      const next = await api.getTree().catch(() => null);
-      if (!next) return refresh();
-      setTree(next);
+      const next = await refresh();
+      // Preserve the old post-move fallback: if its first read fails, make one
+      // more ordinary replacement attempt even though there is nothing safe to
+      // navigate against yet.
+      if (!next) return refresh().then(() => undefined);
       const watching = activeRef?.startsWith('s:') ? activeRef.slice(2) : null;
       if (!watching) return undefined;
       const home = next.groups.find((g) => g.sessionIds.includes(watching));
--- a/web/src/api.ts
+++ b/web/src/api.ts
@@ -24,7 +24,7 @@ const json = (r: Response) => {
 };
 
 export const getClis = (): Promise<Cli[]> => fetch('/api/clis').then(json);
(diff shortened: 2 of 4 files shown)


lvwerra (commented):
Review of #137 against #124. Checked out `a2372ed`, ran it in headless Chromium, and re-ran the checks listed in the PR body. Main is still at `ef08e84`, the commit #124 cites, so the issue's pointers are current and the description of main's behaviour matches the code.

### 1. Medium — a superseded `refresh()` resolves `null` before any tree is published, and callers that navigate on it land on Overview

`refresh` in `web/src/App.tsx:406-408` is now `treeRefresh.current.refresh('replace')`. In `web/src/lib/appRefresh.ts:45` a later `replace` aborts the in-flight one, which rejects the local `canceled` race (`:51`), so the earlier caller's promise settles to `null` **immediately** — before the replacement read has returned anything. On main, `refresh()` was an independent `getTree()` that always ran to completion.

Callers that rely on the tree being fresh when their `await refresh()` returns:

- `createSession` `:664` → `setActiveRef('s:'+id)`
- `quickStart` `:719` → `setActiveRef('s:'+id)`
- `createGroup` `:743` → `setActiveRef('g:'+id)`
- `openSession` `:842` → `setActiveRef('s:'+pane.id)`
- `doMove` `:752-756` → uses `next` to follow a moved pane (a `null` here is treated as a failed read, so the follow-navigation is skipped and a second `refresh()` fires, which in turn aborts the return's read)
- `toggleDemo` `:395`

When the caller's read is superseded, `setActiveRef` runs against the still-stale tree, and the selection check at `:548-558` sees an unknown ref and resets to `'overview'`. The replacement read then arrives with the new session in it, but the selection is already gone.

**Trigger.** Any second `replace` during the caller's `/api/tree` roundtrip: a return/reconnect event (visibilitychange, blur+focus, persisted pageshow, `online`) or another mutation's refresh. Quickstart is the realistic case — the server boots the CLI for seconds, and a tab switch or focus return that lands in the tree roundtrip after it resolves does this. `online` flaps on mobile are another source.

**Reproduced** (headless Chromium, real App wiring, same harness shape as `appRefresh.test.mjs`, tree read held to model a slow network, both reads released afterwards):

```
PR #137
[control]     end activeRef=s:s99 pane=fresh-agent overviewShown=false
[interrupted] end activeRef=overview pane=null overviewShown=true sidebarHasFresh=true

main (ef08e84), same probe
[control]     end activeRef=s:s99 pane=fresh-agent overviewShown=false
[interrupted] end activeRef=s:s99 pane=fresh-agent overviewShown=false
```

Same result whether the interruption is a visibility return or a window blur+focus. Probe and output are in the operator workspace under `review-pr137/` if you want to run it.

This bites #124's requirement 4 ("preserve selected session ... through these reads"). The manager itself behaves as designed; the gap is that the *caller-facing* `refresh()` should not resolve until some fresh tree has been published — e.g. hand a superseded caller the superseding read's promise instead of `null`. Your call on the shape.

### Scope

Clean. Only tree/meta freshness, signals threaded only through `getTree`/`getMeta`, no indicators, labels, Retry, backoff, push, or reader changes. Nothing in the agreed scope is missing or partial.

### Claims checked

- `node test/appRefresh.test.mjs`: passes, 5/5 consecutive runs, ~11s each. Discovered by `run-suites.mjs`.
- `npm test`: 24 suites passed. `npm run typecheck`: passes.
- Revert `App.tsx`+`api.ts` to main, keep the test: fails (`return aborts both obsolete poll transports`).
- Remove only the `observeAppReturns` effect: fails 3/3 at `one return/reconnect burst performs one fresh read per resource`, as the PR body says.
- Remove the deadline from the race: the suite hangs on the frozen read rather than failing, which still fails the run.
- Remove the `mine !== generation` check (`appRefresh.ts:66`): suite still passes. Not a test gap — the abort race already rejects every superseded read, so that check is redundant in practice. Mentioning only because the PR body credits it with preventing late publication.

Nothing else found. The manual tab-switch/reconnect smoke check the issue asks for "when a browser environment supports it" was not done here either; the headless limitation is stated accurately in the PR body.

lvwerra:
Addressed the Medium finding in `44b42bd`; I agree with the reproduction.

`createLatestRefresh` now hands a superseded caller the replacing read's promise instead of settling it to `null`. The handoff is transitive if another replacement arrives, while ordinary poll coalescing, independent failures, deadlines, disposal, abort signaling, and latest-only publication keep their existing behavior.

I added two regressions to `appRefresh.test.mjs`:

- The focused race asserts that the superseded promise remains pending, then resolves to the value actually published by the replacement.
- The Chromium App harness now quickstarts a synthetic session, holds its post-create tree read, supersedes it with a hidden/visible return, and releases the replacement first. It asserts the new session remains selected and the app does not fall back to Overview.

Mutation check: changing the cancellation path back to `null` fails immediately at `the superseded caller waits for the replacing read`. The implementation was then restored.

Verification:

- `node test/appRefresh.test.mjs` — passed
- `npm test` — 24/24 web suites passed
- `npm run build` — typecheck and production build passed

The first aggregate attempt encountered an unrelated Chromium process exit in `filePreviewZoom.test.mjs`; that suite passed alone immediately afterward, and the full aggregate retry passed all 24 suites.

PR #138, opened 2026-09-09
Protect managed skills with file ownership and revision-checked mutations

### LINKED ISSUE #121: Managed skills: prevent accidental deletion and overwrites
# Managed skills: prevent accidental deletion and overwrites

## Goal and PR boundary

Make creating, updating, distributing and removing a managed skill one consistent, ownership-aware lifecycle. A manager operation must affect only the source and installed files it owns, preserve unrelated user installations, and require deliberate intent before deleting or replacing existing content.

**Agreed product decision:** prevention, not recovery. Do not add trash, backups, snapshots, restore actions, undo or version history. Correctly targeted and explicitly confirmed deletion is permanent. Atomic writes and safe retries of incomplete operations are correctness requirements, not a recovery feature.

This is the first PR-sized bundle from the September repository review. It covers B01, the installation/identity/ownership portion of B16, the foundation of F08, and the deletion-scope/partial-failure portion of U12. These IDs are bookkeeping; all implementation requirements are included below.

Keep this PR focused on the lifecycle service, its API integration, and UI safeguards for destructive actions and conflicts. Do not include a general file-manager rewrite, a new skill marketplace, frontmatter redesign, or the separate reusable editor-draft/autosave controller. Skills-specific revision checks and preserving the current buffer on failed saves are in scope.

## Starting points

Planning checked against `main` at `ef08e843d1c67fcc4e1c463415819ea351734d6d` on 2026-09-08. Recheck the latest branch before implementation.

- [Skill routes and distribution helpers](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/server/src/index.js#L1646): `skillPath`, `skillTargetDirs`, `parseSkillFile`, `distributeSkill`, `undistributeSkill`, `distributeAllSkills`, and `/api/skills`.
- [Generated environment skill](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/server/src/index.js#L1297): generated skills and boot redistribution must use the same contract.
- [SkillsEditor](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/SkillsEditor.tsx) and `web/src/api.ts`: mutation results, failure presentation and deletion confirmation.
- [Configuration](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/server/src/config.js): source directory and existing normalization conventions. Do not change the global `slugify` behavior for sessions/groups to solve a skills-only problem.

### 1. One identity and ownership contract

- Extract a small testable skills service with injectable source, state and destination roots. Route handlers and boot/generated-skill callers must share it.
- Resolve the source filename to one validated, nonempty installation ID. Use the same result for create, update, startup redistribution and removal; no different fallback names between operations.
- Reject invalid identifiers and ambiguous normalized-name collisions before mutating any source or destination. Account for case, extensions, punctuation/whitespace normalization and length truncation. Preserve the mapping for existing valid skills.
- Store a versioned, durable ownership manifest connecting the source to its installation ID and the exact managed files at each target. Record enough revision/content information to recognize externally modified files. Persist manifest changes atomically; a missing, corrupt or unwritable manifest must not silently authorize adoption or deletion.
- Own files explicitly, not an entire directory merely because its name matches. Never overwrite or remove another installation's files, extra support files or a configured root. Canonicalize configured roots, deduplicate aliased destinations, and validate recorded destinations again when used; an unchecked manifest path is not permission to write/delete there.
- Reject unexpected symlink redirection at managed source/destination entries. Preserve supported configured-root aliases without claiming that ordinary preflight path checks provide isolation from arbitrary hostile same-user filesystem races.

### 2. Explicit intent and stale-write protection

- Separate creating/uploading a new skill from updating the selected existing skill. Create/upload must be create-only by default. A matching filename or installation ID returns a visible conflict without changing existing bytes; offer a different name or cancel rather than silently replacing it. Users can open the existing managed skill and deliberately edit/save it.
- Use a server-checked revision/content tag for updates and deletion. A stale browser tab, external edit, or concurrent API operation must produce a conflict rather than overwrite or delete the changed version. Check recorded generated-file revisions as well; do not overwrite an externally modified installed copy just because its path is in the manifest.
- Keep ordinary explicit Save usable: updating the currently selected, unchanged-base managed skill does not need another confirmation for every save. No implicit replacement via upload or a selection change.
- Deletion requires a separate confirmation showing the exact skill name and affected managed installations and stating that deletion is permanent. Default focus must not be on the destructive action; Cancel/Escape does nothing. Opening the dialog must never submit deletion.
- Bind the confirmation to the displayed skill identity and revision, not a subsequently changed selection. Disable duplicate submission while the request is in flight. If the skill or target set changed after confirmation was prepared, require a refreshed view/confirmation rather than broadening the deletion silently.
- Enforce ownership, collision and revision requirements in the API/service, not only the UI. Automated callers must use the explicit create/update/delete contract too; a UI confirmation is protection against accidents, not authentication.

### 3. Safe mutation and honest outcomes

- Serialize conflicting operations within the manager process, including startup redistribution versus API updates, so collision checks and manifest commits cannot race each other.
- Preflight identity, ownership and known conflicts before writes. Publish replacements through operation-owned temporary files; a failed replacement must preserve the previous bytes. Temporary-file cleanup must touch only that operation's files.
- Delete only recorded, still-owned files. Preserve extra files and directories belonging to other tools/users; remove a directory only when it is an owned strict descendant and is empty. Never recursively remove a configured ownership root.
- A missing source is not enough evidence to remove an installed directory. Return a documented not-found/idempotent result when there is no valid ownership record. Cleanup after a partly completed deletion must be limited to its originally authorized, still-owned targets.
- Multi-target distribution is not a single filesystem transaction. Report per-target outcomes and retain enough operation/ownership metadata to safely retry an incomplete result; do not claim all targets updated or removed when any failed. Source persistence, target installation and manifest persistence must be distinguishable in the result. Do not retain historical content copies.
- Choose and document source-removal ordering so a failed target removal does not lose the identity needed for safe retry or cause startup to republish a skill whose deletion was already requested. A minimal pending-operation record is acceptable; a general job/recovery framework is not required.
- The UI must retain the selected source and entered content on failed saves, show conflicts/partial failures persistently, and offer an appropriate retry. On partial deletion, show the actual remaining state rather than claiming nothing changed or everything was removed. Retrying must never widen the authorized target set.

### 4. Existing installations and startup

- Define and document migration from installations without manifests. Name matching alone is not sufficient proof of ownership.
- A narrowly verified adoption is acceptable only for an unambiguous source with a byte-identical expected generated `SKILL.md`; record ownership of that file only. Preserve any additional files. Otherwise leave the existing installation untouched and surface a conflict requiring explicit resolution.
- Preserve current local-development distribution opt-in behavior (`SPACE_ID` or `AM_DISTRIBUTE_SKILLS`) and the configured Claude, ordinary home, Gemini and OpenClaw destinations. Do not add a new destination silently.
- Handle generated `environment.md` and ordinary source files consistently, including collisions and storage failures. One problematic skill should not erase unrelated skills or prevent the manager from serving its other functions; report degraded distribution instead.
- Restarting with a manifest or an interrupted mutation must be idempotent. Never treat a partly available directory as an authoritative instruction to erase the remaining installations.

## Required tests

Add discovered server unit/API suites and focused browser coverage. Test the service and its real callers, not just an unused path helper.

| --- | --- |
| Identity | Valid names remain compatible; empty/dot/path-like and normalization-empty names fail without writes; case/extension/punctuation/truncated-name collisions cannot replace each other. |
| Owned lifecycle | Create, update, redistribute on boot and permanently delete across every configured target; the exact manifest and installed bytes agree after restart; no content history/trash is created. |
| Overwrite protection | Same-name upload/create, normalized-ID collision, stale update/delete tags, two editing tabs, external source/installed-file modification. Conflicts preserve existing bytes; deliberate Save with the current revision succeeds. |
| Unrelated data | Seed independent skills and extra support files beside/inside a similarly named directory. Assert they remain byte-identical after successful, failed and repeated mutations. |
| Paths and aliases | Managed-entry and temporary-path symlinks, dangling symlinks, invalid manifest destinations, aliased roots and duplicate targets. No write/delete escapes the resolved owned target; roots survive. |
| Missing/legacy state | Missing source; absent, corrupt or inaccessible manifest; unique byte-identical legacy installation; modified or ambiguous legacy installation; generated environment skill. Adoption never follows name matching alone. |
| Failure/retry | Inject source-write, target-write, unlink, rename and manifest failures. Failed replacements preserve the previous file; partial deletion reports exactly what was removed; restart/retry never widens scope or republishes a pending deletion. |
| Concurrency | Simultaneous conflicting create/update/delete calls, duplicate deletes, startup redistribution racing an API update. No stale manifest overwrite, double ownership or lost newer update. |
| Development guard | Distribution disabled in local mode causes no target writes; explicit opt-in targets only the injected disposable roots. |
| UI | Upload collision leaves the existing skill unchanged; failed Save preserves the editor buffer; permanent deletion requires confirmation; Cancel/Escape/opening the dialog sends no deletion; selection/revision changes cannot retarget a pending confirmation; repeated clicks submit once; errors and partial results remain visible. |

### Test isolation is mandatory

Use disposable source/state/harness directories and child-process environment overrides. Clear inherited `SPACE_ID` and all real harness-home overrides, then opt into distribution only after every configured target resolves under the fixture root. `DATA_DIR` alone does **not** isolate installed skills. Never point these tests at live operator homes, installed skills, running sessions or the production manager. Do not launch real harnesses, publish data, deploy or restart production.

## Acceptance and handoff

- [ ] Every managed-skill entry point uses the same validated identity and ownership service.
- [ ] Unrelated bytes and all configured roots survive the full regression matrix.
- [ ] Creation/upload cannot silently overwrite; stale updates/deletes are rejected server-side.
- [ ] Permanent deletion is deliberate, correctly targeted and explicitly confirmed in the UI.
- [ ] Interrupted/partial operations have truthful results and tested, scope-limited retries.
- [ ] No trash, backup, restore, undo or version-history feature was added.
- [ ] Existing installations have a conservative, documented migration path.
- [ ] Tests exercise API/startup/UI integration as well as the service; new suites run through normal discovery.
- [ ] Run the focused skills suites, default server and web suites, and the production web build/typecheck; run focused browser tests for the changed UI. Report exact commands and results.
- [ ] If an existing unrelated test fails, reproduce it on the unchanged base and document it separately; do not silently omit it or fix unrelated behavior in this PR.
- [ ] Include brief ownership/mutation-contract documentation and screenshots of the changed conflict/confirmation states.
- [ ] Open one focused PR referencing this issue. Do not merge or deploy it without separate authorization.

Follow the repository's `SECURITY.md` for any security-sensitive discoveries during implementation; do not add exploit payloads, operator data or private traces to this public issue or its tests.

lvwerra:
Closes #121

Managed skill operations now share one ownership-aware lifecycle service. Create/upload is create-only; explicit saves and permanent deletion require a current revision. The atomic versioned manifest binds source root, filename, installation ID, unique creation identity and exact generated files. Unrelated installations and support files remain untouched.

The service serializes API, startup and generated-environment callers, validates content revisions and symlink redirection, and adopts legacy installations only when their generated file is unambiguous and byte-identical. Pending intents retain hashes and a fixed target set for incomplete operations. Deletion removes the source last and blocks startup republication. Deleting or customizing a generated skill pauses automatic regeneration; explicit recreation publishes the user's content while keeping that pause. Recreating identical content gets a new identity, invalidating old confirmations.

The editor preserves failed-save buffers, exposes conflicts and per-target partial outcomes, and supports safe retries. Its separate permanent-deletion dialog shows the exact name and installations, focuses Cancel, captures identity/revision, and prevents duplicate submission.

[Ownership, migration and API contract](docs/managed-skills.md) · [Screenshots: upload conflict, stale save, deletion confirmation and partial deletion](https://lvwerra-agent-artifacts.static.hf.space/managed-skills-121.html)

Deliberately excluded: trash, backups, snapshots, restore, undo, version history, marketplace changes, frontmatter redesign, global slug changes, a file-manager rewrite and a reusable draft/autosave controller. The existing general API audit-log policy and bucket-backup system are unchanged; no skills-specific content history was introduced. No new distribution destination, merge or production deployment. A fresh revision lets an operator review and explicitly save or delete an externally edited managed source. The accepted source hash is persisted before mutation, including scoped retries. A failed initial create cannot claim an unrelated source that subsequently appears. Externally modified installed copies and legacy ownership conflicts still require deliberate resolution outside this API; there is no force-adopt/overwrite endpoint. Generation cannot take over a user-created skill or overwrite a customized generated skill; the editor explains the generation pause.

Validation on `a8aed2d`, based on freshly fetched `ef08e843`:

- `npm test --prefix server -- skills` — passes: 57 service tests and 6 real API/startup tests, through normal discovery, plus the package's queued-prompt prelude. Includes sibling-file redirection within an allowed root and preservation of pre-existing empty installation directories; both precise guard mutations fail their regression tests. Also covers reviewed external source saves/deletions and pending retries, plus Settings/startup preservation of customized or recreated generated content.
- `node scripts/mutation-check-skills.mjs` — all 18 service mutations caught.
- Reverting the service to `0ec41e1` in a disposable copy makes exactly the three new API regression tests fail; the other three still pass.
- `node scripts/mutation-check-skills-ui.mjs` — all 3 browser mutations caught.
- `npm test --prefix web` — all 24 discovered suites pass, including the focused browser suite and real editor Save/Delete of externally edited sources.
- `npm run build --prefix web` — production TypeScript check and Vite build pass; existing chunk-size advisory remains.
- `npm test --prefix server` — stops at the **known baseline failure** in `server/test/crons.test.mjs` #4 after six discovered suites pass; later suites are not started by the default runner. Separately reproduced with `node server/test/crons.test.mjs` in an unchanged `ef08e843` worktree: it observes both `schedule` and `restart`, instead of only `restart`. No cron changes included. The 21 other suites pass separately. Migration and resize also pass in disposable test copies with unused ports (only the test port constants changed; application code is this commit), after a prior migration run encountered `EADDRINUSE` on 7893. Overall, 29/30 default server suites pass; only the known cron case fails.

API/browser fixtures clear inherited Space settings, credentials and harness-home overrides, verify all five disposable targets before opting into distribution, and launch no real harnesses. Full-suite runs also used disposable HOME/config roots with inherited Space settings removed. Publication uses temporary-file rename, compatible with roots lacking hard links. Path checks prevent accidental redirection; they do not claim isolation from hostile same-user filesystem races or multiple managers sharing one state root.

### DIFF
--- a/docs/managed-skills.md
+++ b/docs/managed-skills.md
@@ -1,0 +1,152 @@
+# Managed skills
+
+The manager owns a source file and explicitly recorded generated `SKILL.md`
+files. A matching directory name does not establish ownership. Support files,
+other installations and configured roots are never recursively deleted.
+Deletion is permanent. There is no trash, backup, snapshot, undo or content
+history in this service.
+
+## Identity and destinations
+
+`server/src/skills.js` is the shared lifecycle service for the HTTP routes,
+startup redistribution and generated `environment.md`. Its queue serializes
+these callers within one manager process, including their reads and revision
+checks. Do not run two managers against the same source/state roots.
+
+The installation ID retains the previous mapping: strip the last extension,
+lowercase, normalize punctuation/whitespace to `-`, trim edge hyphens, truncate
+to 40 characters. Invalid/path-like filenames and empty IDs are rejected.
+Distinct sources that collide after case, extension, normalization or truncation
+are rejected before either source or installation bytes change. The global
+session/group `slugify` function is unchanged.
+
+Distribution is still enabled only by `SPACE_ID` or `AM_DISTRIBUTE_SKILLS=1`.
+The destinations remain ordinary `HOME/.agents/skills`, Claude's configured
+skills directory, `HOME/.hermes/skills`, and the `.agents/skills` directories
+under configured Gemini and OpenClaw homes. There are no new destinations.
+The source, state and target roots are canonicalized; configured root aliases
+are supported and duplicate destinations are coalesced. Overlapping ownership
+roots are refused. Managed entries, including dangling symlinks and temporary
+entries, must not redirect through symlinks. Replacing an already resolved root
+with a redirect is also refused. These checks prevent accidental redirection;
+they do not isolate files against arbitrary hostile same-user filesystem races.
+
+## Ownership and existing installations
+
+The version 1 manifest is stored at `DATA_DIR/state/skills/skills-v1.json`.
+It binds the canonical source root and each source filename to its installation
+ID, current source digest, exact target roots/files, and generated-file digests.
+It records which skill directories the operation creates; adopted directories
+are never owned. Recorded paths are checked against current configured roots
+and reconstructed expected filenames before use. A changed source root, unknown
+manifest version, corrupt/unreadable manifest or invalid path disables mutations.
+An unwritable manifest cannot authorize a new operation.
+
+On first startup without a manifest, a unique valid source may be adopted.
+An existing destination is adopted **only** when its `SKILL.md` is byte-identical
+to the rendering of that source. Only that file is adopted, never its directory
+or support files. Missing destinations can be installed. A modified destination
+or ambiguous source makes the whole skill conflict before writes. Other skills
+continue independently, and the server reports degraded distribution without
+stopping its other functions. The editor explains when ownership is not
+established. Generated environment content follows the same checks, using the
+existing source to verify a legacy installation before publishing new content.
+
+Name matching alone cannot authorize deletion, even when the source is absent.
+After losing a manifest, removal returns not-found until ownership has been
+narrowly verified again. Never delete the manifest to force an overwrite. To
+resolve a legacy conflict, choose a noncolliding source name or deliberately
+resolve the independent installation outside the manager, then restart
+redistribution. For a file modified after ownership was established, the manager
+refuses to overwrite/delete it; restore the recorded current content or resolve
+its ownership deliberately outside this API. There is no force-adopt endpoint.
+A removed/redirected configured destination must be restored before operations
+on records that still reference it can proceed.
+
+## HTTP mutation contract
+
+All calls still require the usual operation origin (`?from=…`, or the existing
+operator header used by the web client). Origin attribution is not authentication.
+
+| Call | Contract |
+| --- | --- |
+| `GET /api/skills` | Lists source names and pending operations, including records whose source is missing. |
+| `GET /api/skills/:name` | Returns content, an opaque `revision`, ownership/pending state, and exact managed installation paths with their current presence. |
+| `POST /api/skills/:name` | Create only, with a plain-text body. Existing source or installation-ID collisions return 409. |
+| `PUT /api/skills/:name` | Explicit save, with a plain-text body and `If-Match` set to the exact returned `revision`. Missing tags return 428; stale tags return 409. |
+| `DELETE /api/skills/:name` | Permanent removal of the recorded files, with the same required `If-Match` revision. No record returns 404 and removes nothing. |
+
+Revisions include current source content, ownership/operation metadata, observed
+installation contents and the configured target set. A stale tab or changed
+confirmation cannot overwrite/delete the new state. Current revision tags do
+not override detected external changes to an owned installation. Repeated
+already-completed deletions return the same documented 404/no-mutation outcome.
+
+The UI uploads/creates via POST. Conflicts offer another name or cancellation;
+users open the managed skill to edit it. Failed saves retain the selected source
+and entered text. Refreshing its revision preserves that text and displays the
+current saved source for comparison before another explicit Save.
+
+Deletion has its own modal showing the exact filename and installation paths,
+with a permanent-deletion notice. Cancel receives initial focus. Opening it,
+Cancel and Escape do not mutate. Its captured identity/revision cannot follow a
+changed selection, and duplicate submissions are disabled immediately. A stale
+confirmation must be refreshed. Partial deletion displays remaining/absent
+installations before a scope-limited retry.
+
+## Atomic publication, partial outcomes and retries
+
+Before changing content, the service atomically publishes an operation record
+containing the intended source/target hashes and the exact target set. No
+historical content is stored. Source and generated replacements are written to
+exclusive operation-owned temporary files, flushed, checked, and renamed into
+place. A failed replacement preserves the preceding file. New files are checked for absence immediately before rename; the manager queue
+prevents concurrent create calls from passing that check for the same name.
+Publication does not require hard-link support on bucket-backed roots.
+Cleanup never removes a pre-existing or substituted temporary path.
+
+Multi-target distribution is not a filesystem transaction. HTTP 207 results
+carry `ok: false`, separate `source` and `manifest` outcomes, per-target
+`installed`/`removed`/`absent`/`failed`/`not-attempted` outcomes, and the current
+skill snapshot where available. A failed manifest commit is distinguished from
+successful content publication. A persisted intent lets the service recognize
+either the previous or intended hashes after interruption, without claiming
+ownership of differently modified files.
+
+Retry an incomplete create/save by GET followed by PUT with the same intended
+content and current revision. A create whose source write failed may have a
+pending ownership record and no source yet. POST remains create-only; it cannot
+silently turn that pending create into an overwrite. Startup can resume a pending
+save only when its intended source content is available. It leaves a missing
+source or different content alone. Incomplete operations never add destinations
+on retry; newly configured destinations wait for a subsequent ordinary save or
+startup redistribution after the operation finishes.
+
+Deletion persists its intent **before** unlinking targets and removes the source
+**last**, only after every recorded target file is absent. Empty directories may
+be removed only if the manager created them; a failed optional directory cleanup
+is returned as `directoryError`, and the directory is retained. Extra files
+always remain. A target-file failure retains the source and pending record.
+Startup never republishes a pending deletion and does not automatically finish
+it. GET followed by DELETE retries only the originally recorded targets, checking
+their contents again; an external replacement is a conflict. Even if source
+removal succeeded but the final manifest commit failed, the pending record is
+sufficient to finish that same deletion after restart.
+
+## Verification
+
+The discovered suites are `server/test/skills.test.mjs`,
+`server/test/skills-api.test.mjs` and `web/test/skills.browser.test.mjs`.
+The API/browser fixture clears inherited credentials, Space settings and harness
+homes, injects five disposable destinations, verifies them before opting into
+distribution, and starts no real harnesses. Its fault preload exists only in the
+test fixture and is not used by the application.
+
+Run focused discovery with `npm test --prefix server -- skills` and
+`npm test --prefix web -- skills`, and the full suites with `npm test` in each
+package. `npm run build --prefix web` includes the production typecheck.
+`node scripts/mutation-check-skills.mjs` and
+`node scripts/mutation-check-skills-ui.mjs` deliberately remove the important
+guards in disposable code copies and require regression failures. Set
+`AM_SKILLS_SCREENSHOTS` to a disposable output folder when running the browser
+suite to capture its conflict, confirmation and partial-deletion states.
--- a/docs/managed-skills.md
+++ b/docs/managed-skills.md
@@ -48,23 +48,42 @@ to the rendering of that source. Only that file is adopted, never its directory
 or support files. Missing destinations can be installed. A modified destination
 or ambiguous source makes the whole skill conflict before writes. Other skills
 continue independently, and the server reports degraded distribution without
-stopping its other functions. Deleting a generated skill also disables automatic
-generation of that name; an explicit create re-enables it. This stores only a
-name flag, so a later startup cannot recreate a deliberately deleted environment
-skill. The editor explains when ownership is not
-established. Generated environment content follows the same checks, using the
-existing source to verify a legacy installation before publishing new content.
+stopping its other functions. Deleting a generated skill disables automatic
+generation of that name. Explicit recreation publishes the user's content but
+keeps automatic generation disabled. Saving changed generated content also
+pauses regeneration, so Settings changes and startup cannot erase those edits.
+These decisions store only a name flag, never a content copy. The editor explains
+this behavior before editing a generated skill and shows when regeneration is
+paused. Ordinary explicit saves and redistribution still publish the customized
+skill; there is no automatic resumption that can overwrite it. Generation cannot
+take over an existing user-created managed skill. Legacy environment content
+follows the same adoption checks, using the existing source to verify its
+installation before publishing new generated content.
 
 Name matching alone cannot authorize deletion, even when the source is absent.
 After losing a manifest, removal returns not-found until ownership has been
 narrowly verified again. Never delete the manifest to force an overwrite. To
 resolve a legacy conflict, choose a noncolliding source name or deliberately
 resolve the independent installation outside the manager, then restart
-redistribution. For a file modified after ownership was established, the manager
-refuses to overwrite/delete it; restore the recorded current content or resolve
-its ownership deliberately outside this API. There is no force-adopt endpoint.
+redistribution. An externally modified installed copy still blocks saves and
+deletion, even with a fresh revision; resolve that independent installation
+deliberately outside this API. There is no force-adopt endpoint.
 A removed/redirected configured destination must be restored before operations
 on records that still reference it can proceed.
+
+An external edit to a managed **source** (including edits through Files or by an
+agent) invalidates old revisions and stops automatic republication. Open or
+refresh it in Skills, review the current contents, then explicitly Save or
+confirm permanent deletion. A matching current revision authorizes exactly those
+source bytes. The service persists their accepted hash with the operation intent
+before any mutation; installed-file ownership checks remain unchanged. A source
+edit after that snapshot still conflicts. Retrying a pending operation can
+likewise accept a freshly reviewed source, while retaining the original target
+set and, for saves, the original intended content. No historical bytes need to
+be restored to unblock an explicit operation.
+This does not claim a new file that appears after an initial source creation
+failed: without a previously owned source hash, only byte-identical intended
+content can be recognized as an interrupted publication.
 
 ## HTTP mutation contract
 
--- a/docs/managed-skills.md
+++ b/docs/managed-skills.md
@@ -48,17 +48,38 @@ to the rendering of that source. Only that file is adopted, never its directory
 or support files. Missing destinations can be installed. A modified destination
 or ambiguous source makes the whole skill conflict before writes. Other skills
 continue independently, and the server reports degraded distribution without
-stopping its other functions. Deleting a generated skill disables automatic
-generation of that name. Explicit recreation publishes the user's content but
-keeps automatic generation disabled. Saving changed generated content also
-pauses regeneration, so Settings changes and startup cannot erase those edits.
-These decisions store only a name flag, never a content copy. The editor explains
-this behavior before editing a generated skill and shows when regeneration is
-paused. Ordinary explicit saves and redistribution still publish the customized
-skill; there is no automatic resumption that can overwrite it. Generation cannot
-take over an existing user-created managed skill. Legacy environment content
-follows the same adoption checks, using the existing source to verify its
-installation before publishing new generated content.
+stopping its other functions.
+
+### Generated skills are read-only
+
+`environment.md` is derived from the Space configuration: the manager rebuilds
+it on every start and whenever secret descriptions or settings change. There is
+no authored content in it to protect, so it is **read-only**: `POST`, `PUT` and
+`DELETE /api/skills/environment.md` answer `403` with an explanation whatever
+revision is presented, and the editor offers no Edit, Save or Delete for it and
+says why. Keep your own instructions in a separate skill. The service is told
+which names are generated (`generated: ['environment.md']`); a record created
+by generation is read-only even before the name is configured.
+
+For a generated skill the **newest generated content is always authoritative**.
+The retry-with-the-same-content rule that protects hand-authored saves does not
+apply, because nobody can reproduce an earlier generation's bytes on purpose:
+when a generation finds the intent of an interrupted earlier one, it adopts the
+files that carry that intent's bytes (they were written by this manager), drops
+the intent, and publishes the new content. An installation that was modified by
+someone else is still refused by the ordinary ownership checks. Generation still
+cannot take over a user-created managed skill of the same name, and legacy
+environment content follows the same adoption checks, using the existing source
+to verify its installation before publishing new generated content.
+
+**Upgrading from the previous contract**, under which generated skills could be
+edited or deleted: a customized `environment.md` (recorded as
+`disabledGenerated`) is honoured for exactly one more generation — the editor
+shows that the edits are still in place and will be replaced next time — and
+replaced on the following one, after which the editor says the edits were
+replaced. Copy anything worth keeping into another skill during that window.
+A deleted `environment.md` is regenerated at the next generation; nothing is
+lost. Only a name flag is stored, never a content copy.
 
 Name matching alone cannot authorize deletion, even when the source is absent.
 After losing a manifest, removal returns not-found until ownership has been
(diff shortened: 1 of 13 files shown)


lvwerra (commented):
Reviewed by checking out `8b9f032`, running the suites, and probing the service directly with disposable roots. **No correctness defect found in the shipped behaviour** — every destructive path I attacked refused correctly. Two findings, both about the tests protecting that behaviour rather than the behaviour itself.

What I verified independently, all of it holding:

- Deletion leaves unrelated data byte-identical: a sibling skill's `SKILL.md`, a loose file in the target root, and a user file *inside* the owned directory all survive; the owned directory is kept because it is no longer empty.
- Truncation collisions (two names differing only after char 40) are refused before any write; a revision from one skill cannot delete another; deletion without a revision is 428; an externally modified installed file blocks deletion and the hand-edited bytes stay.
- A tampered manifest is not authority: an unconfigured root is refused, a path outside the id directory fails validation, and a corrupt manifest disables mutations rather than authorising them. `precious.md` survived every variant.
- **Migration is real.** I reproduced `main`'s `distributeSkill()` output byte-for-byte for six legacy skills, then ran boot redistribution against it: five adopt with bytes unchanged, the sixth (a name `main`'s own `SKILL_RE` rejected) is reported and left alone. The old `slugify` mapping and the new `skillId` agree on every valid name I tried — they differ only on degenerate names like `_.md`, which now fail loudly instead of silently colliding on the `'skill'` fallback. That is the issue's intent, not a regression.
- No stale references to the deleted `distributeSkill`/`undistributeSkill`/`distributeAllSkills`/`parseSkillFile`/`skillPath` remain (one prose comment in `remote.js:77` still names `parseSkillFile`; cosmetic, not raised as a finding).
- Claims check out: `skills.test.mjs` 48/48, `skills-api.test.mjs` 3/3, `mutation-check-skills.mjs` 10/10 killed, `mutation-check-skills-ui.mjs` 3/3 killed, `npm test --prefix web` 24 suites, screenshots page returns 200. The `crons.test.mjs` #4 failure is the known baseline one.

---

## 1. A corrupted manifest can delete **another skill's** installed file, and no test would notice

`server/src/skills.js` — `load()` validates `t.path !== path.join(t.root, r.id, 'SKILL.md')`. That line is the only thing stopping a manifest record from naming a file inside a *configured* root that belongs to a different skill. `targetsValid()` does not help (the root really is allowed) and `entry()` does not help (the path really is below the root).

Remove that one clause and this happens:

```
create mine.md and other.md
edit the manifest: mine.md's target.path -> <t1>/other/SKILL.md (and copy other's hash)
remove('mine.md')
  shipped code       -> refused: "Invalid ownership manifest…"; other's file survives
  clause removed     -> delete proceeded; other skill's installed file is GONE
```

The suite passes 48/48 with the clause removed. The existing `outside-path` / `outside-root` cases at `skills.test.mjs:178-179` both point *outside* the target root, which `entry()`/`allowed.includes()` catch on their own — so they pass either way and do not exercise this clause.

The issue lists this exact case as required coverage: *"Paths and aliases: … invalid manifest destinations … No write/delete escapes the resolved owned target"*, and *"validate recorded destinations again when used; an unchecked manifest path is not permission to write/delete there."*

Suggested case: tamper a manifest target `path` to a sibling skill's installed file **inside an allowed root**, assert the mutation is refused and the sibling's bytes are untouched.

## 2. The `dirOwned` rule is untested; a regression removes directories the manager never created

`server/src/skills.js` — in `remove()`, `if (t.dirOwned)` is what limits `rmdirSync` to directories this manager created. Change it to `if (true)` and the suite still passes 48/48, but:

```
another tool creates <t1>/mine/ first
create mine.md   (dirOwned recorded false — correctly)
remove mine.md
  shipped code      -> pre-existing directory survives
  guard removed     -> pre-existing directory is removed
```

The closest existing assertion is `skills.test.mjs:96-106`, which records `dirOwned === false` and then checks `helper.py` survives deletion. That directory survives because it is *non-empty*, so `rmdir` fails with `ENOTEMPTY` regardless of the guard — which is why the mutant lives. The case the guard actually exists for, a pre-existing directory that *is* empty after the owned `SKILL.md` is removed, is not covered.

The issue is explicit: *"remove a directory only when it is an owned strict descendant and is empty."*

Suggested case: pre-create the target directory with nothing else in it, create and then delete the skill, assert the directory still exists.

---

### Probed and deliberately not raised

Four other guards also survive mutation, but each is backed by a second check, so removing one causes no damage and I do not think they need tests: the delete-loop `actual !== t.hash` re-check (already covered by `verified()` in the same serialized operation), `!(deleting && src === null)` (the later *"Source changed during deletion"* check still refuses and the hand-edited source survives — I confirmed), `path.isAbsolute(t.root)` (covered by `allowed.includes`), and the `entry()` symlink check (the `!s.isFile()` check on the same line catches a symlinked managed entry — also confirmed).

### Scope

I found nothing implemented that the issue excluded — no trash, backup, snapshot, restore, undo or version history; `disabledGenerated` is a suppression flag, not retained content. Nothing in the agreed scope looked missing or quietly partial. The externally-modified-source dead end (a skill can then be neither saved nor deleted through the manager until its recorded bytes are restored) is real, but it is the issue's stated conservatism and is documented at `docs/managed-skills.md:62-65`, so I am not raising it.

lvwerra:
Both findings are accepted; neither is disputed. Pushed `0ec41e1dfa130e913b06c01a9c960185b7cbc992` to the existing `fix/managed-skills-ownership` branch. This commit changes only the service tests and mutation runner; shipped service behavior is unchanged.

1. **[Manifest destination inside an allowed root](https://github.com/huggingface/agent-manager/blob/0ec41e1dfa130e913b06c01a9c960185b7cbc992/server/test/skills.test.mjs#L209):** the new test creates `demo.md` and `other.md`, redirects demo's recorded destination to other's installed file, and copies the sibling's actual hash. This avoids accidentally testing root containment or external-edit detection instead of identity. A fresh service must reject the read/delete sequence with 503, leaving the manifest, both sources, and both skills' installations across all five roots byte-identical.
2. **[Pre-existing empty directory](https://github.com/huggingface/agent-manager/blob/0ec41e1dfa130e913b06c01a9c960185b7cbc992/server/test/skills.test.mjs#L108):** the new test pre-creates one empty installation directory, creates the skill, and verifies persisted `dirOwned` flags are `[false, true, true, true, true]`. After a service restart and successful deletion, the pre-existing directory must remain empty, the four manager-created directories must be removed, and every configured root must survive. There is no support file that could mask an unconditional `rmdir`.

Both exact guard regressions are now permanent entries in `scripts/mutation-check-skills.mjs`. Before adding the tests, both mutations **survived with 48/48 passing**. Afterward, both are **killed**, and the complete mutation run is **12/12 killed**. Mutants run in disposable copies; production source is never edited by the runner.

Validation on the committed content, with disposable HOME/config/harness roots and no inherited Space distribution settings:

- `npm test --prefix server -- skills`: **50/50 service tests and 3/3 API/startup tests passed**, plus the queued-prompt pre-step.
- `node scripts/mutation-check-skills.mjs`: **12/12 killed**.
- `npm test --prefix web`: **24/24 discovered suites passed**, including the real SkillsEditor/API browser test.
- `node scripts/mutation-check-skills-ui.mjs`: **3/3 killed**.
- `npm run build --prefix web`: **TypeScript check and production build passed**.
- `npm test --prefix server`: six discovered suites passed before the **known main failure in `crons.test.mjs` #4** stopped discovery. Running the remaining **23 suites** separately passed, for **29/30 default server suites passing** overall; the known cron failure is not attributed to this change.

For round two, the complete change is the two named regression tests and two corresponding mutations in commit `0ec41e1`; no runtime guard, migration policy, deletion behavior, or recovery feature changed.

lvwerra (commented):
Review round 2 of #138 against #121, on `0ec41e1` (base `ef08e84`; `origin/main` is still `ef08e84`, so no issue pointer went stale). Round one probed the service on local disk and found the two test gaps the author has since closed. This round went after the behaviour itself through the real server and its callers, and after the one thing round one could not see from `/tmp`: whether the new atomic writer works on the storage production uses.

**No wrong-target write or unconfirmed deletion found.** Every attempt to make a manager operation touch a file it does not own, or delete without the required confirmation, was refused. Two findings on behaviour follow, most severe first. Neither loses data on its own; the first strands data, the second replaces operator-typed content without a conflict.

### Findings

**1. Medium — an externally edited managed source is permanently stranded: boot stops republishing it, and the editor can neither save nor delete it even from a fresh revision.**

Round one saw this and deferred it as documented conservatism. I am raising it because "outside the manager" includes two supported manager workflows, and because the issue's own model has a way out that the PR does not.

- The environment skill the manager generates tells every agent: "Reusable skills … live in `/data/workspaces/skills/` and are published into every agent's skills directory automatically." An agent that edits an existing skill there is following instructions. The Files pane, browsing the workspaces root, can edit `skills/*.md` with the manager's own text editor. Both produce `sourceHash` mismatches.
- Reproduced through the real server fixture: create `demo.md`, overwrite the source on disk, then `GET` (fresh revision, `problem: "Source was modified outside the manager"`), `PUT` with that revision → 409, `DELETE` with that revision → 409. Restart: boot logs degraded distribution and every installation stays at the old content indefinitely. `PUT` after the restart → 409. The operator's only documented recovery is to "restore the recorded current content", but the recorded hash is opaque, and the docs forbid editing the manifest. There is no way forward from the UI at all.
- The issue requires an external edit to "produce a conflict rather than overwrite" — the first refusal is correct. But it also says "Users can open the existing managed skill and deliberately edit/save it" and "Keep ordinary explicit Save usable … unchanged-base". Once the operator has refreshed, the revision covers the current bytes, so the stale-write concern `verified()` guards is already satisfied by `match()`. Refusing at that point is not stale-write protection; it is a lockout. On `main`, boot republished edited skills, so the agent-authoring workflow the product advertises regresses from "propagates on restart" to "silently frozen, log line only".

This is the `src !== r.sourceHash` clause in `verified()` running unconditionally in `write()` and `remove()` after `match()` has already passed. Where a caller supplied a matching current revision, that clause is protecting against nothing the revision does not already cover.

**2. Low — `environment.md` regeneration replaces operator-typed content with no conflict.** Reproduced two ways through the real server: (a) delete the generated skill, `POST` a user skill named `environment.md`, then save Settings (`PUT /api/config`) — the user's source and all five installations are replaced by generated text, `skillDistribution.ok: true`; (b) edit the generated skill via `PUT` with a fresh revision (200), then save Settings — same replacement. `generate()` calls `write()` with `boot: true`, which skips `match()`, and flips the record to `generated` on the first regeneration, so an explicit create re-arms overwriting. Path (b) is pre-existing on `main`; path (a) is new to the "explicit create re-enables" design. Both are the one remaining route where a manager operation replaces content the operator typed, with no revision, no conflict and no confirmation, which is the class this issue set out to close. The editor does not mark `environment.md` as reserved or generated. Low because the content is a generated notice and the name is effectively reserved, but the typed text is gone permanently by design.

### Verified this round

- **Round-one fixes.** Both new tests are present at `skills.test.mjs:108` and `:209`; `scripts/mutation-check-skills.mjs` kills 12/12 including the two new guards; `mutation-check-skills-ui.mjs` kills 3/3.
- **Production storage.** Replayed `atomic()`'s exact syscall sequence on the `/data` bucket mount: `O_EXCL|O_NOFOLLOW` open, `fstat` inode, write, `fsync`, close, `lstat` inode equal to the `fstat` inode, `rename` over the destination. All succeed, so the inode-identity check does not spuriously fail publication in production.
- **Real routes and callers.** Create-only `POST` refuses an existing name and a normalized collision; `PUT`/`DELETE` without `If-Match` return 428, stale tags 409; two `PUT`s with one revision commit exactly one; an externally modified installed file blocks `PUT` and `DELETE` and the `/api/secrets` generated path; partial deletion returns 207 with per-target outcomes, survives restart as `pending: delete`, refuses `POST` during the pending window, and a retry removes only the recorded targets while the sibling support file survives. A dot-file and a subdirectory in the source folder are reported per name without stopping other skills or startup.
- **Claims.** `npm run typecheck` and `npm run build` pass. Server suites: all pass individually except `crons.test.mjs` #4, the known-red case. `skills.test.mjs` 50/50, `skills-api.test.mjs` 3/3. The web runner stopped at `readerRedesign.test.mjs` ("an append does not move a manual reader"); that suite passed 2/2 when rerun on the branch and failed 1/2 on clean `main`, so it is a pre-existing flake, not attributable here. Every other web suite passes individually, including `skills.browser.test.mjs` (5 PASS lines).
- **Scope.** Nothing excluded was built: no trash, backup, snapshot, restore, undo or version history; `disabledGenerated` and `pending` hold names and hashes only. Nothing in the agreed scope is missing beyond finding 1's recovery path.

lvwerra:
Both round-two findings are accepted. Pushed `a8aed2d1d30176a4f77a2f99392bfb767f04d902` to the existing branch.

**Medium — externally edited sources:** GET's matching revision now authorizes the exact source bytes it returned for an explicit Save or confirmed deletion. The service records that accepted source hash with the operation intent before changing files. A stale revision still conflicts, and a fresh source revision cannot authorize overwriting or deleting an externally modified installed copy. Startup leaves unreviewed source edits untouched; the editor's message and generated environment instructions now explain how to review and publish them.

This also works after an interrupted operation: a fresh revision can authorize the current source for retry, while the original installed targets and intended save content remain fixed. No old source contents need to be reconstructed from a hash to proceed. This authorization applies only to an already-owned source: if an initial create failed before owning any source and an unrelated file appears, even a fresh revision cannot claim it. That boundary has a dedicated regression and mutation check.

**Low — generated environment overwrites:** explicit edits to generated content pause automatic regeneration, and deleting/recreating it no longer re-arms automatic replacement of user content. Ordinary saves and redistribution continue publishing the customized skill. The editor identifies generated content before editing and reports the pause afterward. Generation also refuses to take over an existing user-created managed skill. Both `/api/config` and `/api/secrets`, plus restart, are covered. No automatic resumption or recovery/history mechanism was added.

Regression evidence:

- New service cases verify fresh versus stale source revisions, installed-file protection, failed publication/deletion and scoped retries, and preservation of customized/recreated generated skills. The new assertions produced seven failures on the previous service and pass with the fix.
- The real API suite verifies the lockout after restart, explicit Save/Delete, and both Settings-triggered overwrite paths. Restoring the previous service in a disposable copy makes **exactly the three new API tests fail**; the existing three pass.
- The real SkillsEditor browser test now opens an externally edited source, saves reviewed content to all five installations, then confirms deletion after another external edit.
- Mutation checks now cover source acceptance, failed-create ownership, fixed pending-save content, and generated-source ownership as well as every earlier guard: **18/18 service mutations and 3/3 UI mutations caught**.

Validation on the committed content, using disposable state and harness roots:

- `npm test --prefix server -- skills`: **57/57 service tests and 6/6 API/startup tests passed**, plus the queued-prompt pre-step.
- `node scripts/mutation-check-skills.mjs`: **18/18 killed**.
- `node scripts/mutation-check-skills-ui.mjs`: **3/3 killed**.
- `npm run build --prefix web`: **typecheck and production build passed**.
- `npm test --prefix web`: **24/24 discovered suites passed**, including `readerRedesign.test.mjs` and all six browser scenarios.
- `npm test --prefix server`: six suites passed before the **known baseline `crons.test.mjs` #4 failure** stopped the default runner. The 21 remaining non-port-bound suites passed separately. Migration and resize passed in isolated disposable test copies with unused ports; only their port constants were changed, and the server source remained this commit. This rerun resolves a prior `EADDRINUSE` collision on migration port 7893. **29/30 default server suites pass overall**; the known cron failure is unchanged and is not attributed to this PR.

No findings are disputed. The changes retain permanent deletion and conservative automatic publication; no trash, backups, snapshots, undo, or version history were added.

lvwerra:
Pushed `b65ec23`: **the generated `environment.md` is now read-only, and the newest generation always wins.** Operator decision; addresses the collision between regeneration and the managed-skill contract.

**Why.** `environment.md` is rebuilt on every boot and on every secrets/settings change; nothing in it is authored. Under the contract as merged, an interrupted generation left a `pending` intent whose bytes only the *old* inputs could produce. The next boot's generation had different content, `finishWrite` refused it ("retried with the same content"), boot logged degraded distribution, and agents never saw the new variable — unrecoverably, since the operator holds no old bytes to retry with. The rule is right for hand-authored skills (unchanged for them); it is wrong where there is no authored content to lose.

**What changed** (`server/src/skills.js`, `index.js`, `SkillsEditor.tsx`, docs):
- `createSkillsService({ generated: ['environment.md'] })`. Those names — and any record created by `generate()` — are read-only: `POST`/`PUT`/`DELETE /api/skills/environment.md` answer **403** with an explanation, whatever revision is presented. The editor shows that explanation, disables Edit and offers no Save or Delete. Enforced at the API, not only in the UI.
- **Regeneration always wins.** Decision: the pending record is *cleared on the generated path*, not "never recorded" — the intent is still what lets `verified()` recognise a half-written target as ours. `generate()` adopts every file that carries the interrupted intent's bytes (they were written by this manager), drops the intent, and publishes the new content. Files modified by someone else are still refused by the ordinary ownership checks (test kept).
- The `customized` / `disabledGenerated` pause machinery is **deleted** rather than special-cased: with read-only enforced, nothing can produce those states any more. The manifest keeps the `disabledGenerated` field only for reading legacy state.

**Decision on an already-edited `environment.md`** (`disabledGenerated` on a live deployment): *honour once, then replace, and say so both times.* The first generation after upgrade leaves the edit in place and the editor says it is read-only and will be replaced next time — copy anything worth keeping. The following generation replaces it and the editor says the edits were replaced; the notice clears after one more cycle. Only a name flag is stored (`legacyEdit: kept|replaced`), never a content copy. A legacy *deletion* simply regenerates (nothing to lose).

**Tests** (each fails when its guard is reverted; new entries in both mutation runners):
- `skills.test.mjs`: read-only at the service (create/update/remove 403 before and after generation; ordinary skills untouched); interrupted generation superseded by different content (target-write failure and source-write failure variants); externally modified installation still refused; legacy customization honoured once → replaced → notice cleared; legacy deletion regenerates. 60 tests.
- `skills-api.test.mjs`: read-only through the real routes; the exact walk-through from the issue through `PUT /api/config` — a generation fails on one target, the inputs change, the next generation succeeds with no pending, a restart with a leftover interrupted generation regenerates cleanly, and the log never contains "An incomplete save must be retried". Replaces the two tests that asserted the old editable/paused behaviour.
- `skills.browser.test.mjs`: the editor shows the explanation, Edit disabled, no Save/Delete, API 403.
- `docs/managed-skills.md`: new "Generated skills are read-only" section and the upgrade behaviour; API table updated.

**Verification on `b65ec23`:** unit 60/60, API 5/5, browser suite 7/7; `mutation-check-skills.mjs` 19/19 killed (three new), `mutation-check-skills-ui.mjs` 4/4 killed (one new; one pre-existing mutant survived once under load ≈18 and was killed on rerun); all 29 other server suites green individually, only the known main-red `crons` #4 fails; web `npm test` and production build pass. On the integration branch this should turn the two `settings-write` checks about the derived skill converging after a failed write green; I did not touch that branch.

No merge or deployment.

PR #140, opened 2026-09-09
Reader/terminal switching: uncover the terminal when it paints, keep the reader's place

### LINKED ISSUE #127: Reader/terminal switching: reduce terminal delay and preserve scroll position

lvwerra:
Closes #127

Both symptoms were operator-reported, so I reproduced and measured before changing anything. Neither was where it looked, and the measurement changed what I fixed.

**Evidence page:** https://lvwerra-agent-artifacts.static.hf.space/reader-terminal-switch.html

## Reproduction and root cause

I instrumented the switch through the real `TerminalPane` lifecycle with a scripted transport (it opens, sends a canonical `grid`/`restore` control frame, then the retained ANSI) and a scripted trace API. Every phase the issue asked to separate, on the first visit to a shell session:

| phase | ms |
|---|---|
| xterm constructed, socket opened | 19 |
| transport open | 20 |
| canonical restore frame | 22 |
| retained output delivered | 25 |
| **screen painted** | **46** |
| **cover removed** | **179** |

**Terminal, cause 1.** Everything that could plausibly own a 1–2 second wait owns about fifty milliseconds. The gap between painting and uncovering is the delay, and it is the cover's own probe: `setTimeout(…, 150)` after each output burst, asking on a delay a question `term.write()`'s completion callback answers precisely.

**Terminal, cause 2, and the worse one.** That probe reads the screen's **upper two-thirds**. That is the right question for a harness still booting — an agent TUI paints its input bar first and loads its history seconds later — and the wrong one for re-attaching to a session the backend holds a snapshot of. A screen that *is* a bottom-anchored input bar never satisfies it, so the cover sat until its 20-second safety cap. Measured: **20,009 ms of hiding a screen that painted at 10 ms**, on the first visit and on every repeat.

**Reader.** Not a switching bug at all. A reader following the latest output was **339 px short of the bottom on first open**, before any mode switch, with "At latest" already off. Following the end converges over several passes — pin to the end, render the rows that lands on, measure them, pin again — because heights start as a 220 px estimate. In between the content below is still estimated, so the distance to the bottom is briefly large while `scrollTop` has not moved at all. `onScroll` re-derived the follow latch from that distance, and one such sample ended following permanently. The mode switch only made it visible, because a round trip is a fresh mount and a fresh convergence.

## Lifecycle design

`TerminalPane.tsx` — the probe moves into the `write()` completion callback, and the upper-region rule now applies only until the canonical snapshot has been written. The signal is already on the wire: `t:'restore'` promises one authoritative snapshot in the very next frame (`server/src/index.js:2968` sends the frame, then `restore.ansi`), and a session with nothing to restore is sent no such frame — so a new or starting terminal keeps the cold rule *and* its 20-second cap untouched. A `grid` frame that repaints also probes.

My first attempt kept a module-level set of "sessions already seen in this page". I removed it: the wire says it better, and this way there is no retained state to bound or release.

`ConversationView.tsx` — following now ends only when the scroller moves **up**, never because the distance to its bottom grew. Attributing a scroll event to the wheel that caused it is not available: under automation the scroll is dispatched *before* the wheel handler runs, which the fixture demonstrates. So the rule uses the physical difference between the two events instead of trying to attribute them. The test drives the two halves of a gesture in both orders for that reason.

No server change. No hidden transport, no retained emulator, no new cache, no polling change.

## Timing evidence

Same fixture, same measurement code either side, seven consecutive warm switches per row. "Usable" means live content on screen with nothing covering it.

| fixture | before (first / p50 / range) | after (first / p50 / range) |
|---|---|---|
| shell prompt at the top | 179 / 171 / 165–176 | 42 / **36** / 17–45 |
| bottom-anchored input bar | 20009 / **20012** / 20007–20026 | 40 / **13** / 10–45 |
| alternate-screen TUI | 173 / 161 / 160–170 | 42 / **40** / 11–46 |
| 2000 lines of retained scrollback | 170 / 168 / 159–178 | 35 / **36** / 26–39 |
| 240 ms injected transport latency | 190 / 195 / 189–202 | 60 / **60** / 53–70 |

The last row is the control: injected latency still costs about what it costs. Only the invented delay is gone.

**Regression budget: 1500 ms to a usable terminal**, asserted for the first visit and five repeats on every fixture. Measured worst case is 46 ms, so it fails on the regression rather than on a slow runner. It deliberately cannot see a 150 ms regression, so the probe *mechanism* is pinned by source assertion instead — a wall-clock budget loose enough not to flake is too loose to catch it.

Reader scenarios, before → after: at the end with no new output, 339 px then 372 px short → **0 px**; ten exchanges arriving during the terminal visit, 0 px → **0 px** (already worked); deliberately scrolled back, same message and offset → **unchanged**; scrolled back then Latest then a round trip, 372 px short → **0 px**; switch inside the 150 ms save debounce, one exchange adrift → **same message, 24 px**.

## Scope

Implemented: responsive terminal activation (both causes), and reader intent through the real round trip — bottom with and without new output, manual position, rapid intent changes, and the readiness/lifecycle/multi-pane matrix.

**Deliberately left**, with reasons:

- **No `latest()` intent guard.** I wrote one (recording the follow intent at the press rather than inferring it from the scroll it causes, for "a stale remembered manual position must not override a newer explicit jump to Latest") and then removed it: mutation-testing showed no scenario fails without it, because the existing capture-on-scroll path already satisfies the requirement. Shipping a line no scenario needs is not better than not shipping it. The requirement is covered by test either way.
- **Two virtualization behaviours left alone**, both pre-existing and outside this boundary: a large jump into unmeasured rows (a `Home` press) drifts by ~20 exchanges as estimates resolve, and a restored anchor lands within ~24 px rather than exactly. The test documents 40 px of tolerance.
- **No read-state change.** Nothing here infers acknowledgement from a mode switch or a restored scroll flag.
- **No reader rewrite, no terminal-control redesign.** The PR #119 Reader/PTY separation is preserved: `docs/reader-architecture.md` contract 7 ("Terminal and reader are mutually exclusive rendering surfaces") is asserted rather than relaxed.

One thing I would flag as outside the boundary rather than fix here: the terminal is fully torn down and rebuilt on every mode switch, so local selection and scrollback position are lost each time. That is the documented tradeoff from #119 and the measurements say it now costs tens of milliseconds, so it is not worth reopening — but if the operator misses selection across switches, that is the change to spec, not this one.

## Tests and commands

New: `web/test/modeSwitch.test.mjs` — 25 assertions, real React, real `TerminalPane`, real xterm, scripted transport and trace API. Auto-discovered by `run-suites.mjs`, so it runs in the default set. Never starts an agent, opens no real socket, sends no prompt, reads no transcript.

Each fix was mutation-checked by reverting it and counting failures:

| mutation | failures |
|---|---|
| cover reads only the upper two-thirds again | 2 |
| probe back on a 150 ms timer | 2 |
| follow latch back to geometry only | 6 |

```
cd web && npx tsc --noEmit                    # clean
cd web && npm test                            # 24 suites passed (incl. modeSwitch)
cd web && npm run test:render                 # rc=0
cd web && npm run build                       # built in 4.97s
cd server && node test/terminal-modes.test.mjs  resize.test.mjs
                  test/revive.test.mjs  test/repin.test.mjs   # all pass
```

Unrelated baseline failure, on clean `ef08e84` and untouched here: `server/test/crons.test.mjs` #4, `run on restart fires once for enabled running jobs`.

### DIFF
--- a/web/src/components/TerminalPane.tsx
+++ b/web/src/components/TerminalPane.tsx
@@ -686,18 +686,24 @@ export default function TerminalPane({
     // (banner, resumed history) seconds later — only the upper region tells us
     // the pane is genuinely ready. Shell prompts paint at the top anyway.
     let restoring = false;  // a restore frame was seen: the next frame is its snapshot
-    let restored = false;   // that snapshot has been parsed onto the screen
-    const screenHasContent = () => {
+    let restored = false;   // and it carried a screen, so this is a reattachment
+    const screenHasContent = (whole = restored) => {
       try {
         const buf = term.buffer.active;
         // The upper-region rule is a rule about a harness BOOTING, and it only
         // applies while we are still waiting to find out what the screen looks
-        // like. Once the canonical snapshot has been written the screen IS the
-        // answer, wherever its content sits, so the whole screen counts. Read
-        // the upper region before that and a bottom-anchored TUI never
+        // like. Once an authoritative snapshot has been written the screen IS
+        // the answer, wherever its content sits, so the whole screen counts.
+        // Read the upper region after that and a bottom-anchored TUI never
         // qualifies: the cover sat until the 20-second cap on every switch
         // (#127, measured at 20.0s against 13ms of actual paint).
-        const rows = restored ? term.rows : Math.max(2, Math.floor(term.rows * 2 / 3));
+        //
+        // A restore FRAME is not that evidence on its own. attach() calls
+        // ensureRunning() before restore() (server/src/runner.js), so a session
+        // this request just started is sent a restore frame too — carrying an
+        // empty snapshot of a terminal that has painted nothing yet. Its
+        // content is what tells the two apart, and `restored` is set from that.
+        const rows = whole ? term.rows : Math.max(2, Math.floor(term.rows * 2 / 3));
         for (let y = 0; y < rows; y++) {
           const line = buf.getLine(buf.baseY + y);
           if (line && line.translateToString(true).trim().length >= 2) return true;
@@ -792,7 +798,10 @@ export default function TerminalPane({
         restoring = false;
         term.write(d, () => {
           schedulePreview();
-          if (canonical) restored = true;
+          // An empty canonical snapshot is a terminal that was just started for
+          // this request, not a screen to trust: stay in cold-boot mode and let
+          // the upper-region rule wait for the harness to paint something real.
+          if (canonical && screenHasContent(true)) restored = true;
           if (bootLive && screenHasContent()) endBoot();
         });
       };
--- a/web/src/components/TerminalPane.tsx
+++ b/web/src/components/TerminalPane.tsx
@@ -685,11 +685,20 @@ export default function TerminalPane({
     // TUIs paint their bottom input bar first and load the actual content
     // (banner, resumed history) seconds later — only the upper region tells us
     // the pane is genuinely ready. Shell prompts paint at the top anyway.
+    let restoring = false;  // a restore frame was seen: the next frame is its snapshot
+    let restored = false;   // that snapshot has been parsed onto the screen
     const screenHasContent = () => {
       try {
         const buf = term.buffer.active;
-        const upper = Math.max(2, Math.floor(term.rows * 2 / 3));
-        for (let y = 0; y < upper; y++) {
+        // The upper-region rule is a rule about a harness BOOTING, and it only
+        // applies while we are still waiting to find out what the screen looks
+        // like. Once the canonical snapshot has been written the screen IS the
+        // answer, wherever its content sits, so the whole screen counts. Read
+        // the upper region before that and a bottom-anchored TUI never
+        // qualifies: the cover sat until the 20-second cap on every switch
+        // (#127, measured at 20.0s against 13ms of actual paint).
+        const rows = restored ? term.rows : Math.max(2, Math.floor(term.rows * 2 / 3));
+        for (let y = 0; y < rows; y++) {
           const line = buf.getLine(buf.baseY + y);
           if (line && line.translateToString(true).trim().length >= 2) return true;
         }
@@ -698,11 +707,9 @@ export default function TerminalPane({
     };
     let bootLive = false;
     let bootTimer: ReturnType<typeof setTimeout> | null = null;   // safety cap
-    let bootCheck: ReturnType<typeof setTimeout> | null = null;   // throttled content probe
     const endBoot = () => {
       bootLive = false;
       if (bootTimer) { clearTimeout(bootTimer); bootTimer = null; }
-      if (bootCheck) { clearTimeout(bootCheck); bootCheck = null; }
       setBooting(false);
     };
     const connect = () => {
@@ -712,6 +719,7 @@ export default function TerminalPane({
       // A reconnect keeps the already-rendered xterm visible. On a fresh page,
       // `booting` instead exposes the saved preview until canonical restore has
       // painted real content.
+      restoring = false; restored = false;
       const needsCover = !screenHasContent();
       setBooting(needsCover);
       bootLive = needsCover;
@@ -738,6 +746,7 @@ export default function TerminalPane({
           try {
             const m = JSON.parse(d.slice(MODE_CTRL.length));
             if (m.t === 'grid' || m.t === 'restore') {
+              if (m.t === 'restore') restoring = true;
               controllerRef.current = !!m.controller;
               setHasInputControl(!!m.controller);
               const applyGrid = () => {
@@ -760,6 +769,7 @@ export default function TerminalPane({
                     followingBottom = true;
                   }
                   schedulePreview();
+                  if (bootLive && screenHasContent()) endBoot();
                 } catch { /* ignore */ }
               };
               // Writes are asynchronous. A queued empty write is a barrier so
@@ -772,14 +782,19 @@ export default function TerminalPane({
           } catch { /* ignore */ }
           return;
         }
-        term.write(d, schedulePreview);
-        // Probe shortly after each burst (throttled; write() is async).
-        if (bootLive && !bootCheck) {
-          bootCheck = setTimeout(() => {
-            bootCheck = null;
-            if (bootLive && screenHasContent()) endBoot();
-          }, 150);
-        }
+        // The write callback fires when this data has been parsed onto the
+        // screen, which is exactly the question the cover is waiting on. The
+        // old 150ms probe timer asked it late and kept painted content hidden
+        // for that long on every switch.
+        // One canonical snapshot follows a restore frame. Everything after it is
+        // live output.
+        const canonical = restoring;
+        restoring = false;
+        term.write(d, () => {
+          schedulePreview();
+          if (canonical) restored = true;
+          if (bootLive && screenHasContent()) endBoot();
+        });
       };
       ws.onclose = (e) => {
         // A real process exit: stop here and let the user relaunch. Anything
@@ -976,7 +991,6 @@ export default function TerminalPane({
       closedByUs = true;
       if (retry) clearTimeout(retry);
       if (bootTimer) clearTimeout(bootTimer);
-      if (bootCheck) clearTimeout(bootCheck);
       if (resyncTimer) clearTimeout(resyncTimer);
       persistPreview();
       ro.disconnect();
--- a/web/src/components/conversation/ConversationView.tsx
+++ b/web/src/components/conversation/ConversationView.tsx
@@ -38,6 +38,8 @@ export default function ConversationView({
 }) {
   const scroller = useRef<HTMLDivElement>(null);
   const following = useRef(true);
+  /** Where this scroller was at the previous scroll event — see onScroll. */
+  const lastTop = useRef(0);
   const [atLatest, setAtLatest] = useState(true);
   const touched = useRef(false);
   const [query, setQuery] = useState('');
@@ -225,8 +227,25 @@ export default function ConversationView({
       onWheel={interact} onTouchStart={interact} onPointerDown={interact} onKeyDown={interact}
       onScroll={(event) => {
         const el = event.currentTarget;
-        following.current = !q && el.scrollHeight - el.scrollTop - el.clientHeight < 48;
-        setAtLatest(following.current); virtual.onScroll(); capture();
+        // Following is an intent, so only a move AWAY from the end may end it,
+        // and "away" means this scroller moved UP — not that the distance to
+        // its bottom grew. Those are different events and telling them apart is
+        // the whole of #127: a reader following the end converges over several
+        // passes (pin to the end, render the rows that lands on, measure them,
+        // pin again), and in between the content below is still an estimate, so
+        // the distance is briefly large while scrollTop has not moved at all.
+        // Re-deriving the latch from that distance is what parked a following
+        // reader a few hundred pixels short of the end. Attributing the scroll
+        // event to the wheel that caused it is not an option: the scroll can be
+        // dispatched before the wheel handler runs.
+        const far = el.scrollHeight - el.scrollTop - el.clientHeight >= 48;
+        const movedUp = el.scrollTop < lastTop.current - 1;
+        if (following.current ? (movedUp && far) : (!q && !far)) {
+          following.current = !following.current;
+          setAtLatest(following.current);
+        }
+        lastTop.current = el.scrollTop;
+        virtual.onScroll(); capture();
         clearTimeout(settle.current); settle.current = setTimeout(() => { if (position.current) rememberReading(session.id, position.current); }, 150);
         if (touched.current && !q && el.scrollTop < 250 && !loadingEarlier) void loadOlder();
       }}>
--- a/web/test/modeSwitch.test.mjs
+++ b/web/test/modeSwitch.test.mjs
@@ -51,7 +51,7 @@ await build({
     import React from 'react'; import {createRoot} from 'react-dom/client'; import {flushSync} from 'react-dom';
     import TerminalPane from './src/components/TerminalPane';
     const CTRL = '\\x00\\x00AM:';
-    window.sockets = []; window.live = 0; window.sends = []; window.marks = [];
+    window.sockets = []; window.live = 0; window.sends = []; window.wsSends = []; window.marks = [];
     const mark = (name) => window.marks.push({ name, t: performance.now() });
     window.mark = mark;
     let config = {}, root;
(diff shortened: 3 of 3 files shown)


lvwerra (commented):
1. **High — a new/stopped terminal is uncovered on the first bottom-only startup frame, before the harness is actually ready** (`web/src/components/TerminalPane.tsx:688-700, 749, 795`). The new `restored` flag is not evidence that this was a warm reattachment. On the real server, `attach()` calls `ensureRunning()` and then `handle.restore()`; `server/src/runner.js:2102-2111` returns a snapshot even for the terminal it just started, and `server/src/index.js:2970-2977` therefore sends `t:'restore'` plus that initially blank ANSI snapshot. Its callback sets `restored = true`, so the next live paint is checked across the whole screen. If a cold TUI paints only its bottom input bar before loading its banner/history, that frame now removes the cover—the cold upper-two-thirds rule this PR says it preserves no longer applies.

   I reproduced this by extending only the PR's fixture to use `running:false`/`everStarted:false`, then delivering the real sequence: blank canonical restore at 0 ms, bottom-only startup frame at 500 ms, upper-screen content at 1,000 ms. This branch removed the cover at **531 ms**, before the upper-screen content arrived, and the added check failed. The existing fixture hard-codes every session as running/started (`web/test/modeSwitch.test.mjs:112-114`), while its source assertion at 407-411 encodes the false assumption that a new/starting session receives no restore frame. Please distinguish a cold start from a warm authoritative restore and add this server-faithful case.

2. **Medium — the required terminal acceptance matrix is only partially exercised, despite the PR's scope claim.** `FakeSocket` never disconnects/reconnects, every session is hard-coded running, and the browser stays at one 1000×700 desktop viewport with the default zoom. The `usable()` metric at `web/test/modeSwitch.test.mjs:198-209` checks only painted text plus absence of the cover; it does not measure the separately required input-ready state. Thus genuine new/stopped sessions, reconnect after interruption, browser hide/return, mobile/zoom changes, and another viewer's control/geometry are not covered. The two-pane assertion only checks socket session IDs. The cold-start failure above is one concrete bug this missing matrix concealed.

3. **Low — the reported latency comparison is internally inconsistent.** The evidence page labels the row “30ms transport latency” and reports an after median/range around 60 ms, while its next sentence and the PR body call those same numbers **240 ms**. The committed test actually injects 120 ms open + 120 ms restore and measured 273–276 ms in my runs, with no matching before distribution. Please correct or rerun the before/after table so the required same-fixture timing evidence supports the stated condition.

Verification performed on clean `af7f384`: focused mode-switch suite passed; default web suite passed 24/24; render suite, typecheck, and production build passed; the four named server tests each passed when run individually. Reverting the reader follow-latch change produced the six mode-switch failures claimed. The cold-start extension above produced the new failure and was then removed; the checkout is clean and I did not push changes.

lvwerra:
Thanks — the High is real, I reproduced it, and it is fixed in `7e3e956`. Pushed to the same branch.

## 1 (High) — fixed

You are right, and my PR body's claim was simply wrong. I checked the server rather than take it on trust: `attach()` calls `ensureRunning()` before `restore()` (`server/src/runner.js:2058-2059`), and `restore()` snapshots unconditionally (`:2102-2111`), so a session started for that very request is sent `t:'restore'` too — carrying an empty snapshot of a terminal that has painted nothing. Treating the *frame* as evidence of a warm reattachment let the next cold frame, typically a bottom-only input bar, remove the cover. That is a worse experience than the delay I set out to remove, so thank you for catching it.

The flag is now set only when the canonical snapshot **itself carries a screen**:

```js
if (canonical && screenHasContent(true)) restored = true;
```

A warm reattachment always carries one — that is why we are re-attaching. A cold start carries nothing, so `restored` stays false, the upper-region rule and its cap remain in force, and the cover waits for the harness to paint something real. The bottom-anchored-TUI case the PR exists to fix is unaffected, because that snapshot has content.

I built your scenario into the suite as a permanent case: `running:false`/`everStarted:false`, empty canonical restore, bottom-only startup frame at 400 ms, upper-screen repaint at 900 ms. It asserts the cover does not lift on the bottom-only frame. **Reintroducing the defect fails three checks and lifts the cover at 429 ms**, matching your 531 ms; with the fix it lifts at 910–914 ms across repeated runs, on the repaint.

One detail worth recording for whoever reads this next: my first version of that assertion checked the banner was in the DOM at the instant the cover went away, and failed consistently. The probe reads xterm's *buffer*; its DOM rows land a frame later. The assertion now checks the lift time against the frame schedule and confirms the banner arrives separately — a test bug, not a product one, but it would have been a confusing false failure to inherit.

## 2 (Medium) — fixed

Fair, and the cold-start bug is exactly what the missing matrix concealed. The fixture no longer hard-codes the session state, its socket can deliver a scripted frame sequence and a transient drop, and the suite now covers:

- **cold/new sessions** — the case above, with `running`/`everStarted` off.
- **reconnect after interruption** — a 1006 drop; recovers on its own in ~1.2 s with exactly one live transport.
- **input readiness, measured separately** — `usable()` was indeed only paint-plus-no-cover. Keystrokes are now typed into the terminal and asserted to reach the transport as input frames, reported alongside the paint time.
- **viewport and zoom** — the readiness measurement repeats at a 390×844 phone viewport and at 150% zoom.
- **another viewer of the same session** — two panes on one session id; both attach and both paint, so a second viewer does not blank the first.

Two limits I will state rather than overclaim. The fixture models the transport, not the backend's controller/geometry arbitration, so "another viewer's control and geometry remain correct" is covered only as far as attachment and painting go — the real arbitration is exercised by the existing server suites. And browser hide/return is still not covered here; I left it because this container cannot genuinely background Chromium, which I hit on #74 and could not get around then either.

## 3 (Low) — fixed, and you were right about which numbers were which

The row was measured on a 15+15 ms fixture and labelled 240 ms in both the PR body and the evidence page. I re-measured it properly at the injection the committed test actually uses, same fixture on both sides, seven warm switches:

| | before (first / p50 / range) | after (first / p50 / range) |
|---|---|---|
| no injected latency | 181 / 170 / 165–171 | 58 / **36** / 12–37 |
| 240 ms injected (120 open + 120 restore) | 410 / 403 / 402–413 | 273 / **274** / 269–283 |

So the guard overhead that goes away is ~129 ms, and 34 ms of real work sits on top of the injected delay, which is untouched. The evidence page is corrected and now also carries section 4 describing this review finding and the 429 ms → 910 ms correction, so the page does not keep asserting the claim you disproved.

## Re-run

```
cd web && npx tsc --noEmit                 # clean
cd web && npm test                         # 24 suites passed
cd web && npm run test:render              # passed
cd web && npm run build                    # built
cd server && node test/terminal-modes.test.mjs  resize.test.mjs
                  test/revive.test.mjs  test/repin.test.mjs   # all pass
```

`server/test/crons.test.mjs` #4 is the known baseline failure on main and is not mine. Mutation checks re-run on the updated tree: reintroducing the reviewed defect → 3 failures; restoring the upper-two-thirds rule unconditionally → 5; putting the probe back on a 150 ms timer → 2; reverting the reader follow-latch change → 6.

PR #141, opened 2026-09-09
Files: safe replacements and dozens of reader attachments

### LINKED ISSUE #122: Files: safe replacements and dozens of reader attachments

lvwerra:
Closes #122

### Reader attachment batches

- Removes the five-file limit from picker/repeated additions and server resolution without replacing it with another per-prompt count cap. The existing 100 MiB/file, 500 MiB/session, and 200 stored files/session safeguards remain server-enforced.
- Replaces the 20 uploads/minute admission failure with bounded work: one shared FIFO per originating session starts at most three browser transfers; the server streams and serializes quota/publication work per session. Different sessions remain independent.
- Preserves selection order and distinct IDs for equal display names, while still preferring `DataTransfer.files` over its duplicate `items` representation.
- Adds aggregate counts, a bounded/scrollable detail list, per-file progress, queued/active cancellation, retry for transient errors, and always-visible failure rows when collapsed. Size/storage/validation failures retain their reason without offering a futile retry. Send remains unavailable until every remaining file is confirmed.
- Wires reader drop as well as its picker/paste paths. Overview, quick-create, reader, and terminal keep their existing session-specific delivery behavior.

### Workspace replacement and folder safety

- Makes workspace upload create-only by default. A collision returns the exact relative path, full streamed SHA-256 revision, and an opaque replacement token. The Files pane presents a separate Replace dialog with Cancel focused by default; Escape/backdrop cancellation performs no replacement.
- Binds explicit replacement to the exact resolved destination and content revision. A change before or during upload returns a fresh conflict and requires a new decision.
- Streams every upload to a random exclusive same-directory temporary, revalidates, and only then publishes. Aborts and injected publication failures remove only the operation-owned temporary and preserve the prior bytes. Concurrent creates never overwrite one another; final-entry and parent redirect symlinks are rejected.
- Shares the owned-temp/destination-lock primitive with the existing text writer while retaining its current content-tag API and UI state contract.
- Blocks rename/move/delete for the exact or ancestor directory of configured workspaces and shared skills, using real paths and path-segment containment. Defaults, stopped sessions, shared/nested workspaces, aliases, sibling lookalikes, and safe rename/move no-ops are covered.

## Diagnosis

Before the fix I reproduced three independent failures on `ef08e84`:

- attachment upload 21 returned 429 even though the files were tiny;
- the server rejected six IDs at prompt resolution even if all six had uploaded;
- workspace upload silently overwrote an existing file, and interrupting that replacement removed the old destination entirely.

The last case was the data-loss path: the old route opened the destination itself for truncating output, then unlinked that destination on request failure.

## Deliberately left out

This does not redesign editor save/autosave state, add trash/version history, relocate agent workspaces, introduce a general upload service, add remote-agent attachments, promise durable exactly-once prompt delivery, or redesign the reader. There is no Replace All path. The documented hard-link fallback and the cross-process agent-write/rename gap are not described as transactional guarantees.

## Verification

- `web: npm run typecheck`
- `web: npm run build`
- `web: npm test` — 25/25 discovered suites passed
- `web: node test/attachmentBatches.test.mjs && node test/fileUploadReplace.test.mjs`
- `server: node test/attachments.test.mjs && node test/safe-write.test.mjs && node test/files-safety.test.mjs`
- `server: SCREENSHOT_PORT=7922 ISSUE122_SCREENSHOT=/tmp/issue122-reader-50.png node screenshot-input.test.mjs` — all checks passed, including real XHR/server uploads and 50 ordered IDs in the final request
- `server: npm test` — stops at the known-red `test/crons.test.mjs` test 4. The same failure reproduced on an untouched `ef08e84` worktree. I then ran the 23 remaining discovered non-manual suites explicitly; all passed.

Measured in browser coverage: max 3 active transfers for one session and 6 total across two sessions. Measured at server storage: max 1 stream per session and at least 2 across independent sessions. The tests also exercise 6, 30, 50, and 75 files; same-name files; overlapping additions; partial failure; offline/quota errors; queued/active/late cancellation; retry without reuploading successful siblings; narrow layout; create/replace races; mid-stream content changes; failed publication; temp collisions/symlinks; and folder dependencies.

Mutation checks confirmed the focused suites fail when the old five-file resolver cap is restored, final replacement revalidation is removed, ancestor containment is reduced to equality, or the browser transfer bound is removed.

[50-file reader evidence at 390 px](https://lvwerra-agent-artifacts.static.hf.space/issue-122-files-evidence.html)

### DIFF
--- a/docs/screenshot-input.md
+++ b/docs/screenshot-input.md
@@ -122,7 +122,8 @@ same paste/drop behavior:
 - A files-only submission uses `Please inspect the attached file.` (or `files`
   for several) as its text. Image-only server delivery retains the more specific
   `screenshot` wording.
-- At most five files may be attached to one prompt.
+- There is no per-prompt file-count limit. The separate server safeguards remain
+  100 MiB per file, 500 MiB and 200 stored attachment files per session.
 - Choosing, pasting, or dropping a file starts its upload immediately. Each
   chip shows transferred bytes, percentage, and server-confirmed success.
 - The send button is disabled until every upload has succeeded.
@@ -137,6 +138,19 @@ Pending files remain browser `File` objects for retry even after the server has
 stored them. `URL.createObjectURL()` supplies local previews and is revoked when
 a chip is removed or the component unmounts. Successful files are session-owned
 and follow that session's existing deletion/pruning lifecycle.
+
+The browser runs at most three attachment transfers for one session at once;
+overlapping picker, paste and drop additions join the same FIFO. Another session
+has its own queue. The server serializes publication/quota checks within a
+session, so concurrent requests cannot each pass a stale 500 MiB/200-file check.
+There is no automatic retry and no 20-per-minute rejection: transient failures
+stay beside their file with Retry, while size/storage/validation failures keep
+their reason and require removal rather than repeating the same invalid request.
+A lost response can leave one stored file whose id the browser never learned;
+the lifetime byte/file quotas
+bound those orphans, session removal clears them, and the seven-day orphan-store
+pruner handles sessions that disappeared after a crash. This is deliberately not
+an exactly-once transport claim.
 
 ### 5.2 Rendered terminal reader
 
@@ -338,12 +352,11 @@ Errors:
 | `404` | unknown session/attachment |
 | `413` | empty, larger than 100 MiB, or over the session quota |
 | `415` | a claimed PNG/JPEG/GIF/WebP is malformed |
-| `429` | per-session upload backstop exceeded |
 | `500` | durable write failed |
 
-Use a small backstop such as 20 uploads/minute/session. This is not the security
-boundary—the private Space is—but protects the terminal process from accidental
-paste/drop loops.
+The per-session stream lock, browser's three-transfer queue, and byte/file quotas
+are the load bounds. The old 20-uploads-per-minute counter rejected legitimate
+small-file batches at file 21 without limiting active work, so it was removed.
 
 ### 7.2 Preview
 
@@ -510,10 +523,11 @@ Add a small module, `web/src/lib/attachments.ts`, containing:
 - `transferMayContainFile(DataTransfer)`;
 - duplicate suppression across `items` and `files`;
 - local preview creation/revocation; and
-- sequential upload helpers that report progress and preserve per-file errors.
+- bounded per-session upload helpers that report progress and preserve per-file errors.
 
-Use sequential uploads initially. Five files is the maximum, bucket writes are
-the bottleneck, and simpler ordering makes chip status deterministic.
+The browser starts no more than three uploads for one session; the server streams
+and publishes them one at a time under its session quota lock. Selection order is
+kept independently of completion order.
 
 Add a reusable `Attachments` chip row used by Sidebar and Overview. The
 terminal imports only the transfer/upload helpers.
--- a/docs/workspace-file-uploads.md
+++ b/docs/workspace-file-uploads.md
@@ -1,0 +1,57 @@
+# Workspace file uploads
+
+Workspace uploads are create-only unless the operator confirms one exact
+replacement. This is intentionally different from reader attachments: reader
+files always receive unique server ids, so equal display names remain separate.
+
+## Collision and replacement API
+
+`POST /api/files/:id/upload?path=<folder>&name=<name>` streams bytes. With no
+replacement header it creates a new file. If that destination exists, the server
+does not consume it as authorization and returns `409`:
+
+```json
+{
+  "error": "\"report.txt\" already exists",
+  "code": "file-exists",
+  "name": "report.txt",
+  "path": "reports/report.txt",
+  "revision": "sha256:<bytes>:<digest>",
+  "replaceToken": "<opaque token>"
+}
+```
+
+The Files pane shows the exact name and workspace-relative destination. Replace
+is a separate action in a dialog whose focused/default action is Cancel. Escape
+and backdrop dismissal also cancel. There is no Replace All path.
+
+After confirmation the browser repeats the upload with
+`X-AM-Replace-Token`. The token authenticates the resolved destination and a
+streamed SHA-256 identity of the existing content; it is not an mtime check and
+does not buffer the existing file. Tokens are process-local, so a server restart
+requires a fresh collision/decision. If the file or destination folder changed,
+the server returns `409 replacement-stale` with a fresh identity/token and the
+old choice cannot publish.
+
+## Publication and failure scope
+
+Every body first streams to a random, exclusive `.part` file beside the resolved
+destination. The parent/root relationship and replacement identity are checked
+again after streaming. Only then is the temporary published. Aborts, validation
+errors and failed replacement renames remove that operation's temporary and
+leave existing destination bytes alone. Final-entry symlinks are refused;
+parent aliases must resolve inside the same workspace root and must not retarget
+during the request.
+
+Manager-originated writes to one resolved destination are serialized. Create-only
+publication uses a hard link for atomic no-replace semantics. On storage adapters
+without hard links it falls back to `COPYFILE_EXCL`: a competing creator is still
+never overwritten, but the adapter cannot promise that a brand-new file is
+invisible until the exclusive copy finishes. Replacement uses same-directory
+rename after revalidation. An external agent can still write in the small gap
+between the final content check and the filesystem rename; that cross-process,
+storage-adapter race is not presented as a transaction or exactly-once guarantee.
+
+The text editor keeps its existing content-tag API and UI state. It only shares
+the unique temporary-file and per-destination publication primitive; autosave and
+save-state redesign remain outside this work.
--- a/server/screenshot-input.test.mjs
+++ b/server/screenshot-input.test.mjs
@@ -324,6 +324,7 @@ try {
   await readerPicker.setInputFiles({
     name: 'interrupted.bin', mimeType: 'application/octet-stream', buffer: Buffer.alloc(256 * 1024),
   });
+  await readerReply.fill('keep this draft while the file fails');
   const interruptedChip = page.locator('.pane-reader .cxv-composer .image-chip', { hasText: 'interrupted.bin' });
   await interruptedChip.filter({ has: page.locator('.image-chip-retry') }).waitFor({ state: 'visible' });
   const interruptedText = await interruptedChip.locator('.image-chip-meta').textContent();
@@ -331,6 +332,9 @@ try {
     !!interruptedText?.includes('connection was interrupted')
       && await interruptedChip.locator('.image-chip-retry').isVisible(),
     JSON.stringify({ interruptedText }));
+  check('an unresolved file blocks partial Send without losing the draft',
+    await readerReply.inputValue() === 'keep this draft while the file fails'
+      && await page.locator('.pane-reader .ov-send').count() === 0);
   await interruptedChip.locator('.image-chip-retry').click();
   await interruptedChip.filter({ has: page.locator('.image-chip-meta', { hasText: 'uploaded' }) }).waitFor();
   const attachmentDir = path.join(DATA_DIR, 'state', 'attachments', id);
@@ -340,6 +344,32 @@ try {
   const interruptedRemoved = await waitFor(() => !fs.existsSync(attachmentDir)
     || !fs.readdirSync(attachmentDir).some((name) => name.endsWith('-interrupted.bin')));
   check('removing a successful unsent chip deletes its stored file', interruptedStored && interruptedRemoved);
+  await readerReply.fill('');
+
+  await page.evaluate(() => Object.defineProperty(navigator, 'onLine', { configurable: true, value: false }));
+  await page.route(`**/api/sessions/${id}/attachments`, (route) => route.abort('internetdisconnected'), { times: 1 });
+  await readerPicker.setInputFiles({
+    name: 'offline.bin', mimeType: 'application/octet-stream', buffer: Buffer.alloc(1024),
+  });
+  const offlineChip = page.locator('.pane-reader .cxv-composer .image-chip', { hasText: 'offline.bin' });
+  await offlineChip.filter({ has: page.locator('.image-chip-retry') }).waitFor();
+  check('offline failure is distinguished and remains retryable',
+    (await offlineChip.locator('.image-chip-meta').textContent())?.includes('device is offline'));
+  await offlineChip.getByRole('button', { name: 'Remove offline.bin' }).click();
+  await page.evaluate(() => { delete navigator.onLine; });
+
+  await page.route(`**/api/sessions/${id}/attachments`, (route) => route.fulfill({
+    status: 413, contentType: 'application/json',
+    body: JSON.stringify({ error: 'session attachment storage is full (500 MB max)' }),
+  }), { times: 1 });
+  await readerPicker.setInputFiles({
+    name: 'quota.bin', mimeType: 'application/octet-stream', buffer: Buffer.alloc(1024),
+  });
+  const quotaChip = page.locator('.pane-reader .cxv-composer .image-chip', { hasText: 'quota.bin' });
+  await quotaChip.filter({ has: page.locator('.image-chip-meta', { hasText: '500 MB max' }) }).waitFor();
+  check('storage quota failure stays visible and does not offer a futile retry',
+    await quotaChip.locator('.image-chip-retry').count() === 0);
+  await quotaChip.getByRole('button', { name: 'Remove quota.bin' }).click();
 
   // Hugging Face's ingress may answer before Express with an HTML error page.
   // Preserve the status as a useful diagnosis instead of reducing it to "413".
@@ -350,10 +380,11 @@ try {
     name: 'proxy-limit.bin', mimeType: 'application/octet-stream', buffer: Buffer.alloc(1024),
   });
   const proxyChip = page.locator('.pane-reader .cxv-composer .image-chip', { hasText: 'proxy-limit.bin' });
-  await proxyChip.filter({ has: page.locator('.image-chip-retry') }).waitFor();
+  await proxyChip.filter({ has: page.locator('.image-chip-meta', { hasText: 'HTTP 413' }) }).waitFor();
   const proxyText = await proxyChip.locator('.image-chip-meta').textContent();
   check('a non-JSON proxy rejection keeps an actionable HTTP reason',
-    !!proxyText?.includes('proxy rejected this file as too large') && proxyText.includes('HTTP 413'),
+    !!proxyText?.includes('proxy rejected this file as too large') && proxyText.includes('HTTP 413')
+      && await proxyChip.locator('.image-chip-retry').count() === 0,
     JSON.stringify({ proxyText }));
   await proxyChip.getByRole('button', { name: 'Remove proxy-limit.bin' }).click();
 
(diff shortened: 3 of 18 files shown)


lvwerra (commented):
Review round 1 of #141 against #122. Checked out `0c8cef5` (based on `ef08e84`; `origin/main` is still `ef08e84`, so none of the issue's line pointers were stale). Ran it rather than reading the diff alone: all suites, the manual screenshot-input integration, the real Files pane in a browser against the PR server, a filesystem probe on the production bucket mount, and 19 mutations against the new tests.

**No data-loss or wrong-target defect found.** The replacement half holds up under everything I could throw at it. Three low-severity findings follow, most severe first; the rest of this comment is the evidence for the empty part of the review.

### Findings

**1. Low — Replace button does nothing when the collision carries no token.** When the destination disappears between the collision and the confirmation, `receiveWorkspaceFile` takes the `else if (replaceToken)` branch and `collision()` throws `replacement-stale` with `replaceToken: null` ("no longer exists — upload it again"). `WorkspaceUploadError` still classifies that as a collision, so the row shows **Replace…**, the dialog opens, and `confirmWorkspaceReplace` returns early on the missing token: the Replace button is inert with no feedback. The right next action is a plain create, but `Retry` is only rendered for `error` rows. Recovery is Cancel and re-select the file, so nothing is lost. Trigger: upload `a.txt` over an existing `a.txt`, delete the workspace file, click Replace…, then Replace.

**2. Low — the "late success racing cancellation" assertion in `web/test/attachmentBatches.test.mjs` passes without the code it targets.** I removed the `if (controller.signal.aborted) { deleteAttachment… }` block in `uploadPendingAttachments` and the suite still passed. The fixture's `late-success.txt` resolves after ~8 ms, before the click lands; by then `update()` has already set `attachment` on the ref, so the click hits `discardPendingAttachment`, which deletes an ordinary uploaded chip. The delete the assertion counts comes from that path, not from the race. The product code is unchanged from main and reads correctly; only the test's claim is unsupported. Making the fixture resolve only after it has observed `abort` would exercise the real case.

**3. Low — the cross-session independence assertion is satisfied by a global queue.** Replacing the per-session queue key with a single global one kept `attachmentBatches.test.mjs` green because it asserts `maxTotal >= 2`, which a global bound of 3 also meets. The code is genuinely per-session; the PR body's "6 total across two sessions" is measured in `attachments.test.mjs` on the server but not asserted for the browser scheduler. Asserting `maxTotal > ACTIVE_UPLOADS_PER_SESSION` would pin it.

Informational, not a finding: after confirming Replace, the row keeps a Cancel button while the token-bound upload is in flight. If it is pressed after the server has renamed but before the response arrives, the row reads "canceled" although the replacement happened. The user had already confirmed that exact replacement, so no unintended write occurs.

### What was verified

**Replacement safety, real server and real UI.** Booted the PR build with a scratch `DATA_DIR`, created a Files session, and drove the pane with Playwright: a collision left `ORIGINAL-BYTES` intact and reported the exact path; the dialog opened with Cancel focused; Enter on the default and Escape both left the file untouched and wrote nothing; Replace published the new bytes; changing the file on disk between the collision and Replace produced `replacement-stale`, kept the agent's bytes, and left no `.part` file. The raw 409 body matches the documented shape. Mixed batches show every outcome; the fresh sibling file landed while the collision waited.

**Bucket mount semantics.** Probed the production `/data` FUSE mount: `link()` fails with `ENOTSUP`, which is in the module's fallback list; `open('wx')`, `copyFile(COPYFILE_EXCL)` (refusing an existing target with `EEXIST`) and `rename` over an existing file all behave as `safe-write.js` assumes. So the create-only and replace paths work on the storage production actually uses, not just local disk.

**Claims.** `npm run typecheck`, `npm run build`, and `npm test` (25 web suites, three render suites skipped as usual) pass. Every server suite passes individually except `crons.test.mjs` #4, the known-red case. `SCREENSHOT_PUBLIC_DIR=… SCREENSHOT_PORT=7923 node screenshot-input.test.mjs` reports "all checks passed", including the 50-file reader batch with real XHR uploads and 50 ordered ids in the final input request; the screenshot shows the bounded list above an intact composer at 390 px.

**Mutations (19 run, 16 killed).** Killed: restoring the five-id resolver cap; silent overwrite without a token; removing the post-stream content revalidation (unit suite); unlinking the destination on failure (the original data-loss bug); reducing `contains` to equality; dropping realpath in `dependedOnDir`; accepting symlink destinations; dropping the mid-upload folder revalidation; removing the per-session server lock; unbounded browser concurrency; restoring a five-file slice; hiding failures when collapsed; moving default focus to Replace; making Escape replace; auto-replacing on collision; reusing a stale token. Survived: the two above (#2, #3), plus removing the final revalidation against the HTTP suite alone, which is fine because the unit suite catches it.

**Scope.** Everything is inside the agreed boundary: no Replace All, no trash/version history, no autosave redesign, no remote-agent attachments, no new per-prompt count cap (6/30/50/75 exercised server-side and in the browser). The 20/minute counter was removed rather than made burst-aware; the issue allowed "replace or adjust", and the serialization plus byte/file quotas are documented as the bound. Reader drop wiring is within "picker, drag/drop and clipboard". `TerminalPane` no longer imports the removed constant and no five-file wording remains in `web/src`, `server/src`, or `docs`. Quick-create and reader both gate Send on every attachment being confirmed.

lvwerra:
Round-one follow-up pushed in `387dbed`. I agree with all three findings; no pushback.

- **Missing replacement token:** `FilesPane` now treats `replacement-stale` without a fresh token as a normal retryable error. The row offers **Retry**, which sends a create-only request with no replacement token, instead of opening an inert replacement dialog. The browser test now drives the exact collision → destination deleted → Replace → Retry sequence and asserts the request tokens are `[null, old-token, null]`.
- **Late success after cancellation:** the attachment fixture no longer wins the race before the click. `late-success.txt` publishes its synthetic server result only after observing `abort`; removing the production `controller.signal.aborted` cleanup now times out waiting for the required delete and fails the suite.
- **Cross-session queue independence:** the fixture holds the two session batches open long enough to overlap, reads the actual exported per-session limit, and requires total concurrency to exceed that limit. Replacing the session key with one global key now fails with `maxTotal: 3`.

Verification on `387dbed`:

- `web: npm test` — 25/25 suites passed.
- `web: npm run typecheck` — passed.
- `web: npm run build` — passed.
- `server: node ../scripts/run-suites.mjs test/safe-write.test.mjs test/files-safety.test.mjs test/attachments.test.mjs` — 3/3 passed.
- `SCREENSHOT_PUBLIC_DIR=... SCREENSHOT_PORT=17923 node screenshot-input.test.mjs` — all checks passed, including the 50-file reader batch and ordered 50-id send.
- `server: npm test` — reached only the documented known-red `test/crons.test.mjs` #4 after the first six suites passed; no new failure.

Mutation checks were run separately for all three review points: restoring the tokenless collision UI, removing late-cancel deletion, and replacing the per-session key with a global key each fails its corresponding test.

lvwerra:
Review round 2 of #141, on `387dbed`. Second independent pass, aimed at the destructive half: I tried to make the code overwrite or destroy something the operator had not confirmed, rather than re-checking what round one already covered. Checked out and run, not read: a live server on a scratch `DATA_DIR` with four Files sessions (one nested, one rooted at the workspaces root), ~60 adversarial probes against the real routes, and mutations against the three fixes pushed since round one.

**No finding. I could not get a byte of unconfirmed data written or deleted.** Everything below is evidence; two notes at the end are observations, not defects.

### The three round-one fixes hold, and are pinned

Each one was reverted in place and the suite it belongs to was re-run:

| reverted | result |
| --- | --- |
| tokenless `replacement-stale` classified as a collision again | `fileUploadReplace.test.mjs` fails — "a missing destination has no replacement token and must not offer an inert Replace action" |
| the `controller.signal.aborted` delete after a late success | `attachmentBatches.test.mjs` fails (times out waiting for the delete) |
| the per-session queue key replaced by one global key | `attachmentBatches.test.mjs` fails with `maxTotal: 3` |

The Retry that replaced the inert Replace sends **no** token, so the recovery path is create-only and cannot overwrite the file that reappeared in the meantime.

### Trying to overwrite something unconfirmed

Against the live server, always asserting the bytes on disk afterwards:

- **Token replay.** A token minted for `keep.txt` refused for `other.txt`; refused for `twin-b.txt` whose content is byte-identical to `twin-a.txt` (the HMAC covers the resolved path, not just the revision); refused for the same file after the content moved; a token minted in the root-level pane refused when spent from a pane one level down. In every case the target kept its original bytes.
- **Path spellings.** `sub`, `./sub`, `sub/`, `sub/.`, `sub/../sub`, `/sub`, `sub//` all still see the collision — no spelling slips past create-only into a silent overwrite.
- **Aiming out of the workspace.** `../beta`, `%2e%2e%2fbeta`, `..%2Fbeta`, `....//beta`, `/../beta`, `sub/../../beta`, `..`, `/` all refused; the neighbouring workspace's file and the root-level file untouched.
- **Symlinks.** A file symlink pointing outside is refused and its target untouched; a dangling symlink is refused and nothing is created at its target; a directory symlink pointing outside is refused and nothing is planted. A directory symlink pointing *inside* is accepted and canonicalises to the same destination and the same token as the real spelling — which is the right answer.
- **Seven live races**, each with the upload's body held open mid-stream: two confirmed replacements with the same token (exactly one 200, the loser `replacement-stale`, content is one whole version); two creates of the same new name (one wins, one `file-exists`); the file rewritten by another writer mid-stream (replacement refused, the other writer's bytes survive); the destination deleted mid-stream (refused, file stays gone rather than being re-created); a competing creator landing mid-stream (create refused, competitor's bytes survive); the operator cancelling mid-stream (previous bytes intact); the destination folder deleted mid-stream (refused, nothing written at the root instead). No `.am-upload-*.part` survived any of them.

### The editor writer, which this PR moved onto the same primitive

Round one verified the upload path; this is the other writer to the same files.

- A save with the current tag lands; the same tag replayed is refused and the file is unchanged.
- **New protection over main:** if the file becomes a symlink between opening and saving, the save is refused and the file it points at outside the workspace is untouched. Main's writer would have renamed over the link.
- Editor save racing a confirmed replacement of the same file: never both 200, never mixed content, and the file always holds the bytes of whichever writer was told it succeeded.

### Folder dependencies, live rather than mutated

Delete/move/rename of another session's exact workspace: all 409, work inside intact. Delete/rename of the folder *above* a workspace: 409, everything under it survives. A sibling whose name merely starts the same way (`inner-notes` beside the `inner` workspace) is still deletable. Reaching a protected workspace through a symlink inside the pane does not slip past the check. The shared skills folder is refused and its file survives. `store.list()` is unfiltered, so archived sessions count as dependencies too.

### Suites on this commit

`web: npm test` 25/25 (three render suites skipped as usual, `EXIT=0`). `web/test/attachmentBatches.test.mjs`, `web/test/fileUploadReplace.test.mjs`, `server/test/safe-write.test.mjs`, `server/test/files-safety.test.mjs`, `server/test/attachments.test.mjs` all pass individually.

### Two notes, not findings

1. **The destination lock is held for the whole upload stream.** Measured: an editor save of the same file sat unanswered for the duration of a slow upload, then correctly got 409. That is the mechanism working, but on a large file over a slow link the save request hangs for as long as the transfer, with nothing in the UI saying why it is waiting.
2. **The two writers key their lock differently** — the editor on `resolveSafe`'s lexical path, uploads on `realpath(dir) + name`. With a directory symlink inside a workspace the same file therefore takes two different locks; measured, the editor save ran concurrently with the upload instead of queueing behind it. Nothing is lost today because both writers re-validate after staging and the loser is refused — the exposure is the same microsecond window as the documented cross-process gap. It is worth knowing only because it would become a real hole if that post-stage revalidation were ever dropped on the grounds that the lock already covers it.

Scope is unchanged from round one's assessment: no Replace All, no new per-prompt count cap, no trash/history, no autosave redesign.

PR #142, opened 2026-09-09
Filter credentials from full API audit logs

### LINKED ISSUE #132: Filter credentials from full API audit logs and document what is retained
## Goal and agreed scope

Keep the operator's full API-call history useful while filtering recognizable credentials from newly written audit records. Preserve full non-secret prompt, file and response content; do not replace it with hashes-only summaries, truncate it, or introduce a route allowlist. Correct the generated instructions and API-log documentation so they accurately describe this policy.

The operator approved this narrowly scoped bundle. Existing log files are not to be deleted, rewritten or migrated. No new retention/purge settings, general logging redesign or sharing-policy changes.

This public issue is a safeguards/compatibility specification, not a vulnerability report. Follow [SECURITY.md](https://github.com/huggingface/agent-manager/blob/main/SECURITY.md) for private security coordination. Do not include private review reproductions, real credentials, operator log contents or live endpoint details in public issues/PRs.

## Requirements

| Area | Required behavior |
| --- | --- |
| Preserve full logging | Keep all currently captured non-secret content, including plain-text prompts, nested JSON, file-write text and recorded results/errors, with their existing structure and useful attribution/outcome fields. No arbitrary capture limit, truncation, opt-in payload mode or hashes-only replacement. Do not start capturing previously unlogged headers, bodies, streaming traffic or additional read endpoints. Preserve accepted-call and resolved-wait selection behavior. |
| Filter before persistence | Apply one clearly owned credential-filtering policy to the detached audit representation before it is serialized/appended. Cover logged request/query/result strings and audit metadata that can contain user-supplied text; credentials must not remain in an unfiltered sibling field, raw copy, encoded representation or fallback. UI-only masking is insufficient. Keep the actual request body, file contents, response, session objects and agent input unchanged. |
| Structured and inline credentials | Preserve protection for sensitive-named fields and add content-aware filtering for recognizable credentials in ordinary strings. Define a bounded, tested set of supported cases: supported provider-token formats, explicit credential assignments/authorization text, complete recognizable private-key blocks, and exact known configured secret values where safely available. Verify provider formats from official documentation when implementing. Ordinary words are not automatically passwords; do not claim arbitrary unlabeled secrets can always be identified. |
| Matching and false positives | Operate on parsed string values where available so normal JSON escaping, quotes and multiline content are handled correctly. Replace the credential material, retaining surrounding ordinary text; for recognizable private-key blocks remove the whole block, not just a marker. Document support and limitations for embedded encoded text and opaque binary data. If the existing Buffer representation remains supported, it must not bypass the supported text/known-value filtering simply by base64-encoding the original bytes. Do not add general binary decoding, archive inspection or arbitrary recursive format guessing. Use precise boundaries rather than erasing every long identifier, URL, code sample or ordinary word. |
| Known-value scope | Use only a deliberately selected, already available server-side source of configured credentials. Do not scan workspaces, installed harness credential files, historical logs or the whole home directory, query external services, or disclose the secret set to the browser. Define safe treatment of empty/short values, overlaps, duplicates and rotation so matching does not replace ubiquitous substrings or keep obsolete secrets indefinitely. Use dependency injection and fake values in tests, never real environment credentials. |
| Checksums and representation | For new filtered text, derive any retained checksum and text length from the stored filtered value, with clear version/metadata semantics. Do not retain the original secret-bearing text's hash, a secret preview, or a separate per-secret fingerprint as an accidental alternate representation. Preserve normal hashes for unchanged text. Where the UI groups repeated calls, distinguish identical logged/filtered payloads from identical original requests: two different secrets may become the same placeholder. Keep old records readable without rewriting them; do not imply older records used the new filter. |
| Failure and performance | Make filtering deterministic and idempotent, with bounded matcher work proportional to the supported input size. No network calls, repeated disk secret discovery, pathological regex backtracking or raw-tail bypass after a scanning cutoff. Preserve the existing non-fatal audit behavior: a sanitizer/storage failure must not change or replay a completed domain operation or send a second HTTP response. On sanitizer failure, never fall back to raw capture; emit only a minimal safe failure record/diagnostic with internally generated identifiers and a fixed reason, not unfiltered labels, paths, bodies or exception text. Preserve exactly-once recording across response finish/close. |
| Truthful documentation | Update the generated environment instructions, relevant code/API type comments, API-log wording and documentation to state that full non-secret API content is retained with best-effort credential filtering. Explain supported filtering and its limits, that the audit log remains sensitive/private, and that deleting a source session/file does not delete its existing audit copy. State explicitly that this change protects new records and does not clean historical logs or backups. Do not describe filtered logs as safe to publish or automatically free of all sensitive data. |

The scope is intentionally prospective: do not inspect or clean the operator's historical records as part of implementation/testing. Any historical cleanup requires a separately approved action and exact targets.

## Required tests

Use harmless synthetic credentials and isolated storage. Verify newly persisted JSONL bytes and decoded record fields, not merely a redaction count or rendered UI. Also assert that the intended recipient still receives the exact original input/response.

| --- | --- |
| Supported credential matrix | Each documented token/credential family in ordinary prompt text, nested objects/arrays, file-write content, result/error strings and supported query/metadata fields. Include multiple/repeated/overlapping matches and a credential under both a sensitive key and an ordinary text key. Matched material is absent from all newly written representations; surrounding text remains intact. |
| Escapes and multiline content | Quotes, backslashes, Unicode, actual newlines, supported escaped/embedded text forms, multiple complete private-key blocks and incomplete-block behavior as documented. Assert removed material, not only disappearance of a header. Exercise the supported Buffer/text representation without treating base64 as sanitization. |
| Known-value filtering | Injected configured secrets with and without recognizable prefixes, duplicate/overlapping values, empty/short values and changed secret configuration. No real credentials or secret-file discovery. Sensitive fields remain protected even where short-value matching in ordinary prose is intentionally unsupported. |
| Full-content preservation | Short and long non-secret prompts, ordinary code, URLs, IDs, numbers/booleans/null, deep/nested supported payloads and existing multi-megabyte fixtures. Preserve every ordinary byte/character and applicable shape without new truncation or route-dependent omission. Include near-matches to catch overly broad filtering. |
| No domain mutation | Freeze or compare original request/result/session objects; exercise real middleware with fake domain handlers. Requests delivered to fake agents, file writes and actual HTTP responses are unchanged even when their audit copies are filtered. Test success, application error, client close and finish/close races with exactly one appropriate audit record. |
| Hash and UI semantics | Checksums/lengths agree with the retained text. Identical unfiltered inputs remain comparable; different credential values producing the same filtered text are not described as proof of identical original requests. Test mixed legacy/new records, API-log table/hover details and rendered/source views, preserving the existing one-line rows and compact layout. |
| Failure handling | Inject sanitizer failure and append failure. No raw-value fallback, secret-bearing diagnostic, duplicate record, second response or change to the completed operation's outcome. A later healthy audit event can still be recorded. Run these through the real middleware, not only the pure matcher. |
| Prospective behavior | Seed harmless legacy records, append new records, and verify the legacy prefix remains byte-for-byte unchanged. API/UI can still read old and new entries. Do not silently migrate/delete history or imply old records are sanitized. Assertions about absence of new credentials must distinguish appended records from the deliberately preserved fixture history. |
| Documentation and performance | Test the generated instruction output under an isolated destination and assert its capture/privacy claims match behavior. Measure before/after filtering and append cost plus memory/event-loop impact on representative small and multi-megabyte inputs and adversarial near-matches. Document practical matcher limits without silently truncating full payloads or bypassing filtering. |

- Ensure tests are discovered by normal runners. Run focused operations/filter tests, relevant server suites, the real-browser API-log suite, normal server/web suites, and frontend typecheck/production build. Report exact commands, measurements and unrelated baseline failures separately; do not let an unrelated aggregate failure hide whether the new focused tests ran.
- Preserve existing tests for full prompts/file bodies, large entries, origin/target snapshots, resolved versus timed-out waits, and one record on finish/close. Do not weaken those assertions merely to simplify filtering.
- Include a concise filtering-policy table, metadata/version compatibility notes and examples using fake values in the implementation documentation. Filtering is a safety aid, not a confidentiality guarantee.

## Integration and exclusions

Planning inspected main at `ef08e843d1c67fcc4e1c463415819ea351734d6d`; recheck before implementing. Start with `server/src/operations.js`, `server/test/operations.test.mjs`, the generated environment text in `server/src/index.js`, `web/src/api.ts`, `web/src/components/ApiLog.tsx` and `web/test/apiLog.test.mjs`. Keep structural extraction limited to the filter/test seam needed for this change; do not turn the documentation correction into an instruction-template framework rewrite.

Coordinate request/response capture ordering with [#130](https://github.com/huggingface/agent-manager/issues/130) and visibility handling with [#131](https://github.com/huggingface/agent-manager/issues/131). Preserve their rules for rejected-request diagnostics and protected access; do not add raw body capture or an authentication bypass. Check generated-instruction overlap with [PR #109](https://github.com/huggingface/agent-manager/pull/109).

No historical log deletion/rewrite, retention/purge UI, quotas/capture caps, older-history pagination, new audit endpoints, general API/error framework, sharing/export-redaction changes, automatic secret rotation, new login/auth system or storage recovery. Reusing a suitable pure helper is allowed, but changes to sharing behavior require separate scope. No live log/credential reads, installed skill rewrites, agent prompts/restarts, notifications, public probes, paid jobs or deployments during development/testing. Open one focused PR, coordinating sensitive security evidence privately; merge/deployment require separate authorization.

lvwerra:
Closes #132

## What changed

- Added one owned `credentials-v1` persistence filter for new API audit records. It filters sensitive structured fields and property names, documented provider token shapes, explicit authorization/credential assignments, complete private-key blocks, and exact values from the existing server-side injected-secret catalog.
- Applied the filter only while building the detached audit copy. Requests delivered to handlers and agents, file contents, responses, and session metadata remain unchanged.
- New records are version 2 and state whether the credential filter was applied or failed. A sanitizer failure writes only a fixed minimal record; an append failure emits one fixed diagnostic and never retries or changes the completed response.
- Derived text lengths, checksums, and Buffer representations from retained filtered values. The API-log repeat wording now says that it compares retained payloads, and its detail view shows record/filter metadata.
- Traversed and serialized detached audit records iteratively, so valid deeply nested JSON keeps full attribution and content without an arbitrary depth limit.
- Corrected Settings, API types/comments, generated environment instructions, and added `docs/api-audit-log.md` with the retention policy, supported rules, limitations, failure behavior, and v1/v2 compatibility.

## Deliberately unchanged

- Full non-secret prompt, nested JSON, file-write, response, and error content is still retained without truncation, hashes-only replacement, route allowlists, or capture caps.
- Accepted mutations and resolved waits are still the only selected events. This does not capture headers, new bodies, streaming traffic, or additional reads.
- Existing JSONL records and backups are not read, rewritten, migrated, purged, or relabeled. Records without filter metadata remain readable as legacy v1 entries.
- This does not add retention settings, sharing/export changes, authentication, secret rotation, workspace/home/history scanning, recursive archive/encoding inspection, or network lookups.

## Policy boundaries

Filtering is a safety aid, not a confidentiality guarantee. Arbitrary unlabeled secrets and unsupported formats can remain, and the audit log stays sensitive/private. Exact configured values are read per new record from the existing injected-secret catalog, remain server-side, ignore empty/short values, deduplicate overlaps longest-first, and do not stay cached across rotation. Opaque Buffers receive direct ASCII-shape/exact-value filtering only; archives and nested encodings are not guessed.

## Verification

- `node ../scripts/run-suites.mjs audit-instructions operation-credentials operations` — 3/3 focused server suites passed.
- `node test/operation-credentials.test.mjs` — persisted-byte matrix, domain immutability, v1/v2, failures, exactly-once, deep/large/adversarial and Buffer cases passed.
- Mutation checks: bypassing the filter fails on the first synthetic provider token; restoring an unbounded assignment-label expression fails the dash-chain scaling assertion; broadening ambiguous bare labels fails persisted ordinary-code checks; removing word-aware structural matching fails the cited field variants; reverting iterative JSONL serialization fails the 5,000-level append assertion.
- `npm test` in `web/` — 23/23 suites passed, including the real-browser API-log test at desktop and phone widths with mixed v1/v2 rows.
- `npm run build` in `web/` — typecheck and production build passed.
- `npm test` in `server/` — the runner discovered and passed the new instruction suite, then stopped only at the documented clean-main failure in `server/test/crons.test.mjs` case 4. The remaining server suites were run explicitly and passed; the queued-prompts prelude also passed.

Latest full-run measurement on this host: 2.10 MiB representative input and 2.20 MiB adversarial near-matches; baseline summary 3.4 ms, filtered summary 47.1 ms, adversarial filter 69.1 ms, 1,000/4,000-segment dash chains 0.2/0.6 ms, event-loop delay 47.4 ms, filter plus append 51.5 ms, heap delta 10.4 MiB. Thresholds are intentionally generous and the test prints the observed values for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ on local POSIX storage and is restored/checkpointed by
 database backups rather than copying live WAL files. See
 [`docs/agent-state-checkpoints.md`](docs/agent-state-checkpoints.md).
 Scheduled prompts are documented in [`docs/cron-jobs.md`](docs/cron-jobs.md).
+API audit retention and credential filtering are documented in
+[`docs/api-audit-log.md`](docs/api-audit-log.md).
 
 ## Architecture
 
--- a/docs/api-audit-log.md
+++ b/docs/api-audit-log.md
@@ -1,0 +1,81 @@
+# API audit log: retained content and credential filtering
+
+Agent Manager appends accepted mutating API calls, plus resolved agent waits, to
+`operations.jsonl` under the private data directory. The log is meant to answer
+who asked whom to do what and what happened, so it retains full non-secret
+request, prompt, file-write, result and error content. It has no route allowlist,
+payload cap, hashes-only mode or retention setting.
+
+This is a **sensitive private audit log**, not an export-safe artifact. Deleting
+a source session or file does not delete the copy already recorded here. Keep
+the Space and its bucket private.
+
+## New-record filtering policy
+
+Version 2 records carry:
+
+```json
+{"version":2,"audit":{"credentialFilter":{"policy":"credentials-v1","status":"applied"}}}
+```
+
+The filter operates on a detached audit representation immediately before JSON
+serialization and append. The handler, agent input, file content, response and
+session objects keep the original values.
+
+| Rule | What `credentials-v1` replaces |
+|---|---|
+| Sensitive structured fields | Values under normalized credential names such as `authorization`, `password`, `token`, `api_key`, `clientSecret`, `private_key`, `subscription` and `endpoint`. A short value is still removed when its field is explicitly sensitive. Similar ordinary names such as `tokenizer` and `secretary` are retained. |
+| Provider/standard shapes | Hugging Face `hf_…`; Anthropic `sk-ant-…`; OpenAI `sk-…` including `sk-proj-…`/`sk-svcacct-…`; OpenRouter `sk-or-v1-…`; GitHub `ghp_`/`gho_`/`ghu_`/`ghs_`/`ghr_` and `github_pat_`; AWS `AKIA…` access-key ids; Google `AIza…`, `AQ.…` and `ya29.…`; three-part JWT-shaped strings. Boundaries and minimum lengths prevent short examples and ordinary identifiers from matching. |
+| Explicit text | Values in authorization text (`Authorization: Bearer …`, Basic or Token) and credential assignments such as `OPENAI_API_KEY="…"`, `password: …`, query-string `token=…`, and the same shapes in an escaped embedded JSON string. The label and surrounding ordinary text remain. |
+| Private keys | A complete matching `BEGIN … PRIVATE KEY` through `END … PRIVATE KEY` block, including PKCS, RSA, EC, OpenSSH and PGP-style labels. The whole block becomes one placeholder. An incomplete marker is retained because there is no safe block boundary to remove. |
+| Configured values | Exact values, at least eight non-whitespace characters long, from the Space's existing injected-secret catalog. Duplicates are removed, overlaps run longest-first, and one URL-encoded form is matched. The catalog is read for each new record, so rotation does not retain an obsolete matcher. Values are never sent to the browser. |
+| Buffers | UTF-8 Buffers use the text policy. Other Buffers use a one-byte mapping so recognizable ASCII credentials cannot evade filtering merely because the stored representation is base64. The retained byte length, checksum and base64 are all derived from the filtered bytes. |
+
+The provider prefixes follow provider-owned references: [Hugging Face user
+tokens](https://huggingface.co/docs/hub/security-tokens), [OpenAI API
+keys](https://help.openai.com/en/articles/6882433-incorrect-api-key-provided),
+[Anthropic SDK credential examples](https://github.com/anthropics/anthropic-sdk-python/blob/main/examples/workload_identity.py),
+[OpenRouter keys](https://openrouter.ai/docs/api/reference/authentication),
+[GitHub token formats](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github),
+and [Google's Gemini tooling recognizers](https://github.com/google-gemini/gemini-skills).
+
+Every replacement is the fixed string `[redacted]`. No rule name, secret hash,
+preview or per-secret fingerprint is retained.
+
+## What the filter cannot promise
+
+An arbitrary unlabeled password has no reliable shape. If it is not a supported
+token, explicit assignment/private-key block, sensitive structured field or one
+of the currently injected values, it remains. Ordinary base64/hex strings,
+archives, compressed data and nested encodings are not recursively decoded.
+Opaque binary receives only the direct ASCII checks described above. The filter
+does not scan workspaces, CLI credential files, the home directory, historical
+logs or backups, and it makes no network calls.
+
+These deliberate limits avoid erasing long IDs, URLs, code samples and ordinary
+words while keeping work bounded by the retained input, the fixed rule set and
+the small injected-value set. There is no scan cutoff or raw unscanned tail.
+The focused test prints baseline/filter/event-loop timings for a representative
+multi-megabyte value; filtering remains synchronous with the existing append,
+so that measured duration is also the event-loop delay for that write.
+
+Filtering is best effort, not a confidentiality guarantee. Do not publish an
+audit record merely because its metadata says the filter ran.
+
+## Checksums, failures and compatibility
+
+Text `chars` and `sha256`, and Buffer `bytes`/`sha256`, describe the **stored
+filtered value**. Unchanged text keeps its ordinary checksum. Different original
+credentials can both become `[redacted]`, so an equal v2 checksum proves only
+that the retained payloads match; the API-log UI says this explicitly.
+
+If sanitizing a new record throws, Agent Manager writes one minimal v2 audit
+failure entry containing generated identifiers and fixed diagnostic text—never
+the raw path, labels, payload or exception. If append itself fails, it logs one
+fixed server diagnostic and does not retry an append that may have partially
+landed. Neither failure changes or replays the completed API operation, and the
+next healthy event can still be recorded.
+
+Records without `audit.credentialFilter` are legacy records. This change is
+prospective: existing JSONL bytes and backups are not inspected, rewritten,
+migrated or deleted, and older entries must not be assumed filtered.
--- a/docs/api-audit-log.md
+++ b/docs/api-audit-log.md
@@ -60,7 +60,9 @@ credential label, structured sensitive field or configured-value match when
 that material must be recognized. These deliberate limits avoid erasing long
 IDs, URLs, code samples and ordinary words while keeping work bounded by the
 retained input, the fixed rule set and the small injected-value set. There is no
-payload scan cutoff or raw unscanned tail.
+payload scan cutoff, traversal-depth limit or raw unscanned tail. Payload
+traversal and JSONL serialization use heap-backed work lists so deeply nested
+valid JSON does not fall through to a generic filter-failure entry.
 The focused test prints baseline/filter/event-loop timings for a representative
 multi-megabyte value; filtering remains synchronous with the existing append,
 so that measured duration is also the event-loop delay for that write.
--- a/docs/api-audit-log.md
+++ b/docs/api-audit-log.md
@@ -26,7 +26,7 @@ session objects keep the original values.
 |---|---|
 | Sensitive structured fields | Values under normalized credential names such as `authorization`, `password`, `token`, `api_key`, `clientSecret`, `private_key`, `subscription` and `endpoint`. A short value is still removed when its field is explicitly sensitive. Similar ordinary names such as `tokenizer` and `secretary` are retained. |
 | Provider/standard shapes | Hugging Face `hf_…`; Anthropic `sk-ant-…`; OpenAI `sk-…` including `sk-proj-…`/`sk-svcacct-…`; OpenRouter `sk-or-v1-…`; GitHub `ghp_`/`gho_`/`ghu_`/`ghs_`/`ghr_` and `github_pat_`; AWS `AKIA…` access-key ids; Google `AIza…`, `AQ.…` and `ya29.…`; three-part JWT-shaped strings. Boundaries and minimum lengths prevent short examples and ordinary identifiers from matching. |
-| Explicit text | Values in authorization text (`Authorization: Bearer …`, Basic or Token) and credential assignments such as `OPENAI_API_KEY="…"`, `password: …`, query-string `token=…`, and the same shapes in an escaped embedded JSON string. The label and surrounding ordinary text remain. |
+| Explicit text | Values in authorization text (`Authorization: Bearer …`, Basic or Token), bounded compound credential assignments such as `OPENAI_API_KEY="…"`/`access_token=…`, quoted bare sensitive fields in embedded JSON, uppercase environment-style bare assignments, and the same quoted shapes in an escaped embedded JSON string. Parsed query-string `token` fields are covered structurally. The label and surrounding ordinary text remain. |
 | Private keys | A complete matching `BEGIN … PRIVATE KEY` through `END … PRIVATE KEY` block, including PKCS, RSA, EC, OpenSSH and PGP-style labels. The whole block becomes one placeholder. An incomplete marker is retained because there is no safe block boundary to remove. |
 | Configured values | Exact values, at least eight non-whitespace characters long, from the Space's existing injected-secret catalog. Duplicates are removed, overlaps run longest-first, and one URL-encoded form is matched. The catalog is read for each new record, so rotation does not retain an obsolete matcher. Values are never sent to the browser. |
 | Buffers | UTF-8 Buffers use the text policy. Other Buffers use a one-byte mapping so recognizable ASCII credentials cannot evade filtering merely because the stored representation is base64. The retained byte length, checksum and base64 are all derived from the filtered bytes. |
@@ -52,9 +52,15 @@ Opaque binary receives only the direct ASCII checks described above. The filter
 does not scan workspaces, CLI credential files, the home directory, historical
 logs or backups, and it makes no network calls.
 
-These deliberate limits avoid erasing long IDs, URLs, code samples and ordinary
-words while keeping work bounded by the retained input, the fixed rule set and
-the small injected-value set. There is no scan cutoff or raw unscanned tail.
+Assignment labels are bounded to 128 characters, which keeps malformed
+dash-joined input linear rather than repeatedly scanning the same suffix.
+Ambiguous lowercase, unquoted bare assignments such as `key=value` are retained
+because they are common in source code, YAML and prose; use a compound
+credential label, structured sensitive field or configured-value match when
+that material must be recognized. These deliberate limits avoid erasing long
+IDs, URLs, code samples and ordinary words while keeping work bounded by the
+retained input, the fixed rule set and the small injected-value set. There is no
+payload scan cutoff or raw unscanned tail.
 The focused test prints baseline/filter/event-loop timings for a representative
 multi-megabyte value; filtering remains synchronous with the existing append,
 so that measured duration is also the event-loop delay for that write.
--- a/server/src/audit-credentials.js
+++ b/server/src/audit-credentials.js
@@ -51,10 +51,8 @@ const AUTHORIZATION = new RegExp(
 // at every word boundary and allowed unlimited `word-` prefixes; on a long
 // dash-joined line it repeatedly re-walked the suffix (quadratic event-loop
 // work) even though the line contained no assignment operator at all.
-const ASSIGNMENT_VALUE = new RegExp(
-  String.raw`((?::|=)\s*)(${VALUE})`,
-  'g',
-);
+const ASSIGNMENT_OPERATOR = /[:=]/g;
+const VALUE_AT = new RegExp(VALUE, 'y');
 const COMPOUND_ASSIGNMENT_SUFFIXES = [
   'apikey', 'accesstoken', 'refreshtoken', 'authtoken', 'apitoken',
   'clientsecret', 'secretkey', 'privatekey',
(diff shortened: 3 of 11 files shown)


lvwerra (commented):
Review of #142 against #132, round one. Checked out `0cef0b7`, ran the server and web suites, and probed the filter with synthetic inputs. Main is still at `ef08e84`, the commit #132 cites, so the pointers are current. Per SECURITY.md the exact probe inputs and outputs are not in this comment; they are in the operator's private workspace under `review-pr142/`.

### 1. High — the assignment matcher backtracks quadratically and runs on the server's event loop

`CREDENTIAL_LABEL` in `server/src/audit-credentials.js` starts with `(?:[a-z0-9]+[_-])*`, and `ASSIGNMENT` anchors it with `\b`. On a whitespace-free run of many short alphanumeric segments separated by hyphens, every segment start is a word boundary, so the engine re-walks the rest of the chain from each one. Underscores do not trigger it (`_` is a word character), which is why the PR's adversarial fixture stays fast.

Measured on this host, `createCredentialFilter([])` alone:

| input (one line) | size | time |
| --- | --- | --- |
| 1 000 dash-joined UUID-shaped ids | 36 KB | 0.4 s |
| 2 000 | 72 KB | 2.6 s |
| 4 000 | 145 KB | 12 s |
| 20 000 | 723 KB | 196 s |
| same 4 000 ids newline-separated | 145 KB | 3 ms |

The filter runs synchronously inside the `finish`/`close` handler in `operationMiddleware`, so one accepted file write or prompt of that shape stalls every other request on the Space for that long. #132 requires "bounded matcher work proportional to the supported input size" and "no pathological regex backtracking"; `docs/api-audit-log.md` currently states work is "bounded by the retained input". The prefix loop needs a shape that cannot restart per segment (or the label needs to be matched without an unbounded prefix).

### 2. High — bare `key`, `token`, `secret`, `password` labels erase ordinary code and data

`CREDENTIAL_LABEL` ends in `|token|secret|key)` and `ASSIGNMENT` fires on any of them followed by `:` or `=`. The unquoted `VALUE` branch then consumes to the next whitespace. Through `summarizePayload(..., 'content', filter)`, i.e. a file write:

- Python `sorted(items, key=lambda x: x[1])` → `key=[redacted] x: x[1]`
- JSX `<li key={item.id}>{item.name}</li>` → `<li key=[redacted]` (the rest of the tag is gone)
- JSON `{"key": "north", "secret": false, "token": 42}` → all three values `[redacted]`
- YAML `key: value` → `key: [redacted]`
- prose `The password: is required…` → `The password: [redacted] required…`

Agents in this repo write React and Python constantly, so this is routine content, not an edge case. #132: "Use precise boundaries rather than erasing every long identifier, URL, code sample or ordinary word" and "Preserve every ordinary byte/character". The docs say the limits "avoid erasing long IDs, URLs, code samples and ordinary words", which is not what the filter does. The compound labels (`api_key`, `client_secret`, `access_token`, …) and the quoted-value branches are fine; it is the four bare words with the to-whitespace value that need tightening.

### 3. Medium — structural field protection is narrower than main for several common credential names

`isSensitiveAuditKey` replaces main's substring regex with an exact set plus a suffix list. Suffixes include `apikey`, `accesskey`, `privatekey` but not plain `key`, and `secret`/`subscription`/`password` only as suffixes or exact names. Verified through `summarizePayload` with the filter attached: a body of `{"secret_key": "plain-value-1234"}` is persisted in full. Also kept now, redacted on main: `SECRET_KEY`, `secretKey`, `subscriptionKey`, `Ocp-Apim-Subscription-Key`, `passwordHash`, `endpointUrl`, `subscription_id`. The inline text rule does include `secret[_-]?key`, so the two halves of the policy disagree. #132: "Preserve protection for sensitive-named fields". Freeing `tokenizer`, `maxTokens`, `sinceTokens` is intended and fine.

### Scope

Clean. Full content is retained (no truncation, hashes-only or allowlist), only the detached audit copy is filtered, existing JSONL is untouched (legacy-prefix byte check passes), no retention or sharing changes, instructions and docs updated. The known-value source is the existing curated catalog in `index.js` (build-time keys, SPACE_/AM_/KUBERNETES_ prefixes and a fixed non-secret set excluded); on this Space it resolves to token names plus one variable below the 8-character floor, so no ubiquitous-substring problem here.

### Claims checked

- `node ../scripts/run-suites.mjs audit-instructions operation-credentials operations` (server): 3 suites passed.
- `npm test` (server): runner stops at the known-red `crons.test.mjs` case 4; the 7 suites before it pass (new `audit-instructions` included); the 19 after it, run individually, all exit 0.
- `npm test` (web): 23 suites passed, including the real-browser API-log suite with mixed v1/v2 rows. `npm run typecheck` and `npm run build`: pass.
- Mutation: making `createCredentialFilter` return its input unchanged fails `operation-credentials` on the first synthetic provider token, as the PR body says.
- Docs vs behaviour: the instruction text and `docs/api-audit-log.md` describe retention, prospective scope and the failure path accurately; the two sentences quoted in findings 1 and 2 do not match the filter today.

lvwerra:
Round-one fixes are pushed in `26aabf6` and `0b08936`. I accept all three findings; no disputes.

1. **High — quadratic assignment matching: fixed.** The filter now scans `:` / `=` operators once from left to right and inspects at most 128 label characters immediately before each operator. It no longer starts an unbounded prefix expression at every word boundary. The new scaling check uses harmless generated dash-joined text: the final implementation measured 0.2 ms for 1,000 segments and 0.6 ms for 4,000 segments in the full non-cron run. Mutation-checking it with the prior unbounded shape failed as intended (about 2.1 s / 18.0 s).

2. **High — ordinary code/data erased by bare labels: fixed.** Ambiguous lowercase unquoted bare labels are retained, and bare `key` is no longer treated as an inline or structural credential by itself. Compound credential labels remain supported; quoted embedded sensitive fields and uppercase environment-style bare assignments remain supported. A persisted file-write regression test now preserves the cited Python, JSX, JSON, YAML, and prose shapes byte-for-byte, while separate assertions pin the supported explicit-assignment cases. Broadening the bare-label decision again makes this test fail. The policy table and limitations now describe this boundary.

3. **Medium — structural coverage narrower than main: fixed.** Structural key classification now combines normalized exact/suffix checks with separator/camel-case word parsing. The persisted record test covers `secret_key`, `SECRET_KEY`, `secretKey`, `subscriptionKey`, `Ocp-Apim-Subscription-Key`, `passwordHash`, `endpointUrl`, and `subscription_id`; all redact, while `tokenizer` and `secretary` remain ordinary. Removing the word-aware branch makes the persisted-byte test fail.

Verification after the final matcher change:

- Focused server suites (`audit-instructions operation-credentials operations`): 3/3 pass.
- All 29 non-cron server suites pass. The first migration run encountered the documented shared port 7893 collision; its retry passed, as did the two suites it initially prevented from starting.
- The aggregate server runner still stops only at the known-red `crons.test.mjs` case 4.
- Web: 23/23 suites pass; typecheck and production build pass.
- Representative final full-run measurement: 2.10 MiB filtered in 49.2 ms, 2.20 MiB adversarial input in 65.0 ms, event-loop delay 49.4 ms, filter plus append 51.9 ms, heap delta 12.3 MiB.

All reproductions and new fixtures use synthetic values only; no private probe payloads or deployment data were copied into the branch or this comment.

lvwerra (commented):
Round two, independent pass on `0b08936`. Checked out and ran the branch. **One finding, Medium.** Round one's three fixes hold up under fresh probing, and the areas round one left unexamined — checksum provenance, the Buffer representation, idempotency, known-value edge cases, the failure path and the UI's repeat semantics — all check out.

Per SECURITY.md, probe inputs are described by class only. Everything below used harmless synthetic values.

## 1. Medium — a caller can make its own operation unattributable by nesting the request body deeply

This PR removes `MAX_DEPTH` from `summarizePayload` (`server/src/operations.js`) and replaces it with a `WeakSet` that handles cycles but does not bound depth. The traversal is recursive, so a sufficiently nested body overflows the stack. `JSON.parse` and the default `express.json()` body limit both accept nesting far deeper than the traversal survives, so this is reachable from an ordinary request: a body around 10 KB is enough.

The overflow lands inside the `try` in the record handler, so the catch writes the minimal failure record. That record is designed for a *sanitizer* failure, where withholding everything is exactly right. Here the record was otherwise fine, and what is withheld is the attribution. Driven through the real middleware with fake domain handlers:

| body nesting | audit record |
| --- | --- |
| 10 | `status: applied`, origin, target, method and path all recorded |
| 1 000 | `status: applied`, all recorded |
| 5 000 | `status: failed`, `origin: null`, `target: null`, `method: AUDIT`, `path: /audit/credential-filter` |

In every case the operation itself completed and answered normally, with exactly one `json` call and exactly one audit record — the non-fatal and exactly-once guarantees are intact, and nothing leaks. What is lost is who asked whom to do what. On main the same body logged normally, with `[max-depth]` markers below level 20.

The consequence worth weighing is that the choice belongs to the caller. An agent that wants its prompt delivered but not attributed can pick the shape of its own body; the prompt still reaches the target, and the log records only an anonymous filter failure at that timestamp. In a change whose purpose is to make the audit log safe to keep in full, a caller-selectable path to an unattributable entry seems worth closing.

`#132` rules out an "arbitrary capture limit", so restoring a depth of 20 is not the answer. Three shapes that do not reintroduce one, in the order I would consider them:

- Bound the traversal at a depth far beyond any real payload — a few hundred — and emit a marker for that branch, so a deep body degrades one subtree instead of the whole record. This is main's mechanism at a depth no genuine payload reaches.
- Make the traversal iterative, so depth costs heap rather than stack.
- Distinguish a traversal overflow from a filter failure and keep the identifying fields in the former. `origin` and `target` are resolved server-side from the roster, not from request text, so they are safe to retain when the thing that failed was not the filter.

`sanitizeMetadata` has the same unbounded shape, but its inputs are small server-built objects, so I do not think it is reachable.

This slipped past the existing coverage because the failure-handling test injects a *filter* failure (`filterFactory` throwing), which is the case where nulling everything is correct. A case that overflows the traversal on an otherwise-clean record would have caught it, and would fail against the current code.

## Round one's fixes, verified rather than re-litigated

- **Quadratic matching.** Gone. 3 MiB of ordinary text filters in 29 ms; the multi-megabyte and adversarial figures in the author's reply reproduce.
- **Ordinary code and data erased by bare labels.** I ran a 20-case corpus of Python, JSX, JSON, YAML, TypeScript, CSS, SQL, shell, prose, URLs, timestamps, UUIDs and near-miss token shapes, including all five shapes round one cited. **All 20 survive byte-for-byte.** The documented boundary is exactly what the filter does: compound labels, quoted embedded fields and uppercase environment-style assignments redact; ambiguous lowercase unquoted `token:`/`secret:`/`password=` are retained, and the limitations section says so.
- **Structural coverage.** The word-aware classification is a superset of main's for the names round one listed and narrower only where intended.

## What I checked that round one did not

All clean:

- **Checksums.** `chars`, `sha256` and Buffer `bytes` are all derived from the stored filtered value; the hash of the original secret-bearing text is not retained anywhere.
- **Buffer representation.** A credential inside a UTF-8 Buffer and inside an opaque binary Buffer is filtered, and the stored base64 decodes to the filtered bytes — base64 is not an escape hatch.
- **Idempotency.** Twelve shapes including already-redacted text, quoted placeholders and trailing punctuation: `filter(filter(x)) === filter(x)` throughout.
- **Known-value handling.** Regex-special characters are matched literally, the URL-encoded derivative works, the eight-character floor and whitespace-only values behave as documented, and a short value under an explicitly sensitive field is still removed.
- **Leak sweep.** A synthetic provider token and a synthetic configured value placed in a plain string, a nested object, an array, an object *key*, a UTF-8 Buffer and a binary Buffer: absent from the serialized record in every case.
- **No truncation, hashes-only or route allowlist.** A 3 MiB prompt is retained character-for-character; deep-but-safe nesting, numbers, booleans, null and arrays keep their shape.
- **No domain mutation.** The request object is byte-identical after summarizing.
- **UI semantics.** The repeat tooltip, hover card and header comments now say equal checksums mean equal *retained* payloads and explicitly disclaim identical originals.
- **Documentation.** I re-checked the specific claims in `docs/api-audit-log.md` against behaviour — incomplete private-key markers retained, complete blocks removed whole, multiple blocks, the 128-character label bound, the configured-value rules. They match. The generated instruction text is truthful about full retention, best-effort filtering, prospective scope and the log remaining sensitive.

## Test runs

- `node ../scripts/run-suites.mjs audit-instructions operation-credentials operations` (server): 3/3 pass.
- `npm test` (web): **first run failed on `readerRedesign.test.mjs`. Re-ran it in isolation twice: passes both times, and the full suite passes on retry, 23/23.** That is the known contention flake on this box, not a finding — flagging it so nobody chases it.
- `node test/apiLog.test.mjs` (the browser API-log suite for this change): passes.
- `npx tsc --noEmit` and `npm run build` (web): pass.
- `server/test/crons.test.mjs` #4 remains the known-red baseline on main.

## Scope

Agrees with round one. Full non-secret content is retained, only the detached audit copy is filtered, existing JSONL is untouched, and there is no retention, purge, sharing or auth change. The known-value source stays the existing server-side catalog and never reaches the browser.

lvwerra:
Final-round fix is pushed in `d5124ad`. I accept the Medium finding; no disagreement.

I replaced the recursive request/result traversal with an explicit heap-backed work stack. It preserves the full nested shape, uses the existing active-path `WeakSet` semantics for actual cycles, and introduces no depth marker or capture limit.

I also replaced native `JSON.stringify` at the JSONL append seam with an iterative serializer for the already-filtered representation. This second half is necessary: after making traversal iterative, native serialization still overflows on the same valid deep structure and drops the record during append.

The new real-middleware regression case sends a harmless synthetic body nested 5,000 levels and asserts:

- the handler receives the original object unchanged and responds exactly once;
- exactly one JSONL record is appended with filter status `applied`;
- origin, target, method, and path remain attributable;
- every nested level is retained;
- a synthetic recognized token at the deepest leaf is filtered;
- shared branches remain shared content and only an actual cycle becomes `[circular]`.

Mutation check: reverting only the append seam to native `JSON.stringify` makes the new test fail because no deep record is appended. Reverting the traversal to the reviewed recursive implementation fails earlier by producing the generic filter-failure entry.

Verification after the final change:

- Focused server suites: 3/3 pass.
- All 29 server suites other than the known-red `crons.test.mjs` suite pass.
- Web: 23/23 suites pass on the first run.
- Web typecheck and production build pass.
- No private probe values or deployment identifiers were used or copied into this comment.

PR #143, opened 2026-09-09
Save settings immediately, and never let an older answer speak for a newer edit

### LINKED ISSUE #123: Save settings immediately and make file-editor saves reliable

lvwerra:
Closes #123

Settings had a 900 ms debounce, so a change waited for a typing pause and closing the panel dropped whatever had not fired. Both settings writes then overlapped freely, and the response that came back last won regardless of which edit it described. The file editor stays explicit-save, but had the same ordering flaw with worse consequences: a save that landed after the user typed again cleared the newer draft and reported "saved" over text that was only in the buffer, and a second Save during a write was dropped entirely — so "save and close" could close over uncommitted text.

Evidence, including the red baseline on unmodified main, the six save states and the mutation table: https://lvwerra-agent-artifacts.static.hf.space/save-ordering-123.html

**Round 2** (`b84cdce`) acts on all eight review findings; the reply comment covers each one. The largest was a file response that could still speak for a file it no longer belonged to after navigation or unmount.

## Reproduced first

`web/test/saveRaces.test.mjs` parks every write and lets the test choose the order the answers come back in, so nothing here depends on a timer. On `main` at ef08e84 it is red in sixteen places — both halves of the issue, observed before anything was changed. It is green here, and now carries 95 checks; `server/test/settings-write.test.mjs` carries 30.

## What this does

**Immediate settings, explicit files, one ordering rule for both.** Both debounces are gone: a valid change is sent as it is made — no blur, no close, no Apply. A write already in flight retains the latest pending value in one slot (not a queue of keystrokes) and sends it the instant the write settles, with no second delay. Typing in a file still writes nothing.

**No older answer may speak for a newer edit.** One controller (`web/src/lib/saveQueue.ts`) gives both policies the same guarantee: a response may clear a draft, advance a base or say "saved" only if nothing newer was asked for while it was away. A file write carries the session and path it belongs to, so a success that lands after navigation — or after the viewer is gone — releases the remembered draft only if that draft *is* the text committed, and otherwise keeps it and rebases it onto the version that just landed. Another file's draft, buffer, metadata and base are never touched.

**Two writers cannot silently overwrite each other.** `/api/config` and `/api/secrets` return the revision of the committed bytes and require it back as `?base=…`; a stale writer gets `409` with what is actually stored, and the operator chooses (*Keep mine* / *Use theirs*). Nothing is merged or forced automatically.

**A damaged settings file is not an empty one.** Unreadable settings are reported on the read and refused on the write, with the original bytes untouched — the previous behaviour handed the next ordinary edit permission to replace a hand-written config with defaults.

**Truthful endpoints.** Atomic temp+rename, no `{ok:true}` for a write that did not happen, and the committed value and revision in the response.

**Saves outlive their panel.** The settings savers live in `web/src/lib/settingsSaves.ts` for the life of the page. Closing Settings cannot cancel work already asked for; a late failure is reported in the app chrome with a way back and a Retry; reopening shows the value still owed rather than fetching over it; `beforeunload` warns while anything is outstanding.

**A finite failure path.** Every save has a bounded window for an answer. On timeout the slot is released and the state is *not confirmed* — a lost answer may mean the server committed — and the resource is read back before anything is sent again: recognised as committed, or sent again through the ordinary precondition. Never a forced overwrite to recover.

**Input semantics.** An incomplete numeric field is shown, marked and explained rather than saved as a reset to 0. Server normalization is not adopted into a field the user may be typing in.

**Failures stay actionable.** Any file failure — not only a conflict — carries Discard and Retry in the info strip; conflicts keep Reload/Overwrite.

**Derived work is a separate outcome.** The generated `environment` skill is rebuilt after the response, coalesced, from committed state; its failures are reported (`derived`) instead of being swallowed into a state that reads as fully applied. Skill distribution beyond that file remains #121's.

No new backups, trash, version history or restore feature; no diff or recovery UI; no application-wide store migration. The existing unsaved-file draft behaviour is preserved and is what the ordering tests assert on. #121, #122 and PR #120's areas are untouched.

`docs/saving.md` documents the immediate-versus-explicit policy, the revision and error semantics with curl examples, pending-operation lifetime, and what a browser can and cannot promise during shutdown.

## Testing

- `web/test/saveRaces.test.mjs` (95 checks): file ordering, queued explicit saves, save-and-close, refusal/overwrite/Retry, cross-file and post-unmount responses, settings immediacy and coalescing, conflicts both ways, lost answers and reconciliation, incomplete numeric input, close/reopen, the app-level late failure, and a guard against the viewer re-reporting itself forever.
- `server/test/settings-write.test.mjs` (30 checks): committed value and revision, whole-JSON on disk with no temp file left behind, the two-tab race, storage failure answered as failure with the previous bytes intact, damaged settings preserved and refused, and the derived update reported separately from persistence.
- Ten mutation checks, one per guard — each listed with what it breaks on the evidence page.
- Full web suites (24), `npm run test:render`, every server suite, `tsc --noEmit`, `vite build`.
- `server/test/crons.test.mjs` #4 fails on this branch and identically on stashed clean `main` at ef08e84. Pre-existing, untouched.

### DIFF
--- a/docs/api-errors.md
+++ b/docs/api-errors.md
@@ -1,0 +1,159 @@
+# HTTP validation and failures
+
+This is the compatibility inventory for the Express **4** request boundary.
+The routes and successful payloads are unchanged. HTTP failures retain the
+top-level `error` string used by copied shell/agent clients, and add `code`:
+
+```json
+{"error":"name: must not be empty","code":"invalid-input","details":[{"field":"name","message":"must not be empty"}]}
+```
+
+`error` is display text, not HTML; clients branch on `code`, not English.
+Expected exceptions use `ApiError`. Arbitrary exceptions (including library
+errors carrying their own `status`) do not authorize exposing their messages.
+Unexpected errors return a generic 500. Legacy local 5xx JSON replies are also
+sanitized before the audit recorder sees them. An optional `requestId` is a
+server-generated UUID, never a request-supplied identifier. No exception body,
+stack, upstream response or new secret-bearing diagnostic is logged here.
+
+## Status and code contract
+
+| Status | Codes | Safe extra data / meaning |
+| --- | --- | --- |
+| 400 | `invalid-input`, `bad-request`, `invalid-json`, `invalid-path` | `details: [{field,message}]`; no rejected values echoed |
+| 400 | `origin-required`, `unknown-origin` | Existing attribution requirement, unchanged ordering |
+| 400 | `bad-repo`, `not-a-bundle` | Existing import refusals |
+| 400 | `request-aborted` | Upload/body interrupted before a response was committed |
+| 403 | `locked` | `reason: public-space`; health/info/visibility remain accessible |
+| 403 | `forbidden`, `no-access`, `no-hf-token`, `backup-unavailable` | Expected access/capability refusal |
+| 404 | `not-found`, `api-not-found` | Missing item versus nonexistent API route |
+| 404 | `no-trace`, `unsupported-harness`, `trace-not-user-conversation` | Expected Reader availability, not a failed fetch substituted with empty content |
+| 409 | `conflict`, `not-archived`, `input-not-ready`, `invalid-workspace`, `remote-agent`, `cron-unavailable`, `backup-running`, `no-transcript` | Action cannot run in the current state; not permission to replay it |
+| 409 | `file-changed` | Existing `mtime` preserved; browser also retains `tag`/`currentTag` when supplied by file-save endpoints |
+| 409 | `redaction-blocked` | Existing `hits` rule/count map preserved |
+| 409 | `conflict` on paused remote messages | Existing `stop` and `reason` retained, plus additive `error`/`code` |
+| 413 / 415 / 429 | `payload-too-large` / `unsupported-media-type` / `rate-limited` | Body limits, media/encoding policy, existing rate limits |
+| 5xx | `internal-error` | Generic safe message; optional bounded `requestId`; a settings write may name its validated data-root-relative file in `details: [{field:"path",message}]` |
+
+`reason`, `hint`, `hits`, `mtime`, `tag`, `currentTag`, `retryAfter`, and
+`details` are optional, not universally present. A 5xx never carries an absolute
+path or exception text; the settings filename is relative to the durable data
+root (for example, `am-config.json`). The browser retains these
+allowlisted fields in `ApiError.data`, with bounded strings/collections. It
+retains `status` even for an unreadable proxy response. No automatic retry,
+new timeout policy, mutation replay or delivery guarantee is added.
+
+The audit recorder keeps one outcome. If the connection closes before the
+server finishes responding, `ok:false` and `incomplete:true` distinguish it from
+success; `status:499` is audit-only when no headers were committed. This is not
+evidence that the operation was rolled back or that a replay would be safe.
+Request-body capture and filtering are unchanged.
+
+## Route / response / mutation inventory
+
+Paths below are relative to `/api`. The handler inventory was checked against
+main `ef08e843d1c67fcc4e1c463415819ea351734d6d`, rather than relying on planning
+line numbers. `apiRoutes` covers every API registration in `index.js`, including
+the generated `mkdir`/`touch` routes. Global JSON parsing, privacy lock and
+operation attribution stay in their original order; validation runs after
+route-specific parsers and before the final action. The separate browser-origin
+and audit-filter work is not reimplemented here.
+
+| Routes | Forms and responses | Mutation / compatibility notes |
+| --- | --- | --- |
+| GET `health`, `info`, `visibility`, `clis`, `tree`, `sessions`, `next-name` | JSON | Health/public lock exceptions unchanged; next-name validates CLI |
+| GET `usage`, `operations`, `traces`, `meta`, `meta/:id` | JSON | Provider enum; operation limit 1–1000, existing timestamp cursor; optional provider/digest subsystem fallbacks remain local |
+| POST `push/subscribe`, `push/unsubscribe`, `notify` | JSON command | Subscription structure, endpoint and notification fields validated before dispatch; notification zero-device/per-device partial outcomes remain **200 `ok:false`** |
+| POST `sessions` | JSON creation, 201 | CLI required; optional name/prompt/path; blank name retains generated default, blank path retains workspace root; group must exist; optional prompt still starts asynchronous work |
+| PUT `sessions/:id` | JSON selected-field update | Display name only, no folder rename |
+| POST `sessions/:id/input`, `attachments/insert` | JSON command | Text / attachment ID types checked before resolution or agent input; existing attachment limits and insertion acknowledgement unchanged |
+| POST `sessions/:id/attachments` | Raw bytes, 201 JSON metadata | Any file media type, including JSON, bypasses JSON parser; existing quotas, image validation and cancellation retained |
+| DELETE `sessions/:id/attachments/:attachmentId`; GET `…/raw` | JSON command; binary stream | Stream errors before headers are JSON, after headers terminate; disconnect destroys file source |
+| GET `agents`, `agents/:id`, `…/tail`, `…/subagents`, `…/subagents/:agentId` | JSON | Tail line integer validation with existing 2000-line clamp (including clients requesting more), trace query validation; roster/trace distinctions retained |
+| GET `agents/:id/wait` | JSON long poll | State list, timeout 1–300 s, settle 0–60 s; existing wait/deadline behavior retained, disconnect stops polling |
+| POST `agents`, `agents/:id/prompt`, `agents/:id/stop` | Text body **or** JSON `prompt` / `text`; `from` body/query, launch CLI/name/path/group query aliases | Preserve curl's non-JSON default content type as literal prompt text, not form decoding. Sender, target, installed CLI, self-target and group resolution checks remain before action |
+| GET `remote/:name/ping`, `…/messages`, `sessions/:id/remote`; POST `remote/:name/hello`, `sessions/:id/remote/paused` | JSON | Remote peer strings/paused boolean; scalar sequence cursor; existing remote-origin fallback unchanged |
+| POST `remote/:name/messages`; GET `…/prompt`, `…/stream` | Text/JSON message command; plain-text instructions; heartbeat/NDJSON long poll | Paused poll's **200 stop result** is deliberate; paused message is 409. Stream sends connected/heartbeat/one JSON line, never JSON error appended after bytes. Wait remains 5–1800 s, no global short timeout |
+| POST `demo`, `welcome/seen`, `relaunch`, `update`; GET `update/check` | JSON commands/status | Demo boolean. `no-space`, `no-token`, `cooldown`, `busy`, `upstream-unreachable` and upstream refusal capability results remain **200 `ok:false`**; thrown request failures now return safe 500 |
+| GET/PUT `config` | JSON **replacement** | Omitted sections/fields keep existing replacement defaults; booleans are not coerced; explicit zero, false and backup `exclude:[]` preserved. No durable-store/recovery rewrite |
+| GET/PUT `secrets` | JSON notes-map replacement | Notes contain arbitrary user-defined keys with string values, including empty strings; no global field stripping |
+| GET `backup/status`; POST `backup/run` | JSON status / starts asynchronous Job | Missing prerequisites, already-running and bad IDs explicitly classified; status subsystem's optional lookup fallbacks unchanged; no Job lifecycle redesign |
+| GET `skills`, `skills/:name`; PUT/DELETE `skills/:name` | JSON listing/read; text replacement / delete | PUT must decode to a string, including valid empty text; JSON objects cannot silently overwrite a skill with empty text; distribution remains best effort |
+| GET `files/:id`, `…/preview`, `folders` | JSON | Scalar paths; existing lexical/realpath, file-kind and size checks retained |
+| GET `files/:id/raw`, `…/download`, `trace/:id/download` | Binary/download stream | Existing sandbox/content-disposition headers; explicit callback forwarding; no JSON appended after headers |
+| PUT `files/:id/write` | Text replacement, JSON result | Valid empty text allowed; existing `base` revision precondition, text-size and path checks retained; JSON objects rejected before writing |
+| POST `files/:id/mkdir`, `…/touch`, `…/rename`, `…/move`; DELETE `…/entry` | JSON commands | Single filename / destination types validated before filesystem changes; root/dependency/existing-entry checks retained |
+| POST `files/:id/upload` | Raw bytes → JSON completion | Raw JSON files also bypass body parsing. One completion/error path; abort unpipes/closes destination and releases listeners. Existing overwrite policy unchanged (separate file-protection work) |
+| POST `overview/hidden`, `move`; POST/PUT/DELETE `groups[/id]` | JSON | Hide boolean/ref; live move targets; group PUT is a selected-field patch, not replacement. `layout:null` means auto; explicit layout is 1–3 rows/columns; session ordering retains live membership checks |
+| GET/POST `crons`; PUT/DELETE `crons/:id`; POST `…/run` | JSON; creation 201, manual fire 202 | Create requires name/agent/prompt/five-field schedule/time zone. PUT patches selected fields, nested agent/schedule objects still replace as before. Strict runOnRestart/state types. 202 means accepted, not finished; later failure stays in `last` with safe text. No target identity/scheduling behavior change |
+| GET/POST `sessions/:id/share`; GET/POST `share/access` | JSON | Visibility/name/user lists validated; redaction 409 and successful share's `granted`/`grantErrors` retained. Partial grant errors get a safe message/code; best-effort access-list lookups remain unchanged |
+| POST `trace/import`; GET `trace/bundles` | JSON | Bare repo or pasted dataset URL still supported; known domain codes kept. No safe re-import/transaction changes |
+| GET `trace/:id`, `…/location`, `files/:id/trace`; PUT `trace/:id/source` | JSON trace pages/windows/summary; selected source update | Trace modes mutually exclusive; safe nonnegative integer cursors, v1/v2; existing page/window-size clamps retained (500 turns / 8 MiB). Source kind enum, ref/target checks before storing |
+| POST `sessions/:id/archive`, `…/unarchive`, `…/stop`; DELETE `sessions/:id` | JSON commands | Existing archive/delete and `ifNeverStarted=1` rules unchanged; optional attachment cleanup remains isolated |
+| `/ws` | WebSocket terminal protocol | Not JSON wrapped; framing, origin admission and terminal lifecycle unchanged |
+| Other `/api` paths; static assets and SPA paths | API 404 JSON; ordinary static/SPA responses | API errors never fall through to an HTML application page; non-API failures remain Express/static concerns |
+
+## Validator conventions
+
+- Validate at the route boundary before any domain action, not only in TS.
+  Schema helpers distinguish missing values from null/empty/false; no generic
+  truthiness or numeric-string conversion for JSON. JSON-only payloads require
+  `application/json`; bodyless commands may omit it. Malformed JSON, unsupported
+  encodings and parser limits get deliberate JSON errors. Express's strict
+  parser rejects scalar JSON documents (including bare `null`).
+- Queries are scalar strings: duplicate/bracket-structured parameters are
+  rejected. Numeric query fields use complete integer syntax, safe integer
+  bounds and route limits. Existing documented clamping of trace size budgets
+  stays in the domain reader. Paths reject null characters and retain existing
+  containment checks; do not introduce a second path resolver.
+- Unknown fields retain each endpoint's existing handling, usually ignored.
+  No input object is stripped or rewritten by validation. Notes-map extension
+  keys remain supported. Intentional empty group names on creation, empty text
+  file replacement and null group layout keep their special meaning.
+- Schedule validation is reused, with a public validation error instead of
+  arbitrary parser exception text. Store schemas/durability, scheduler identity,
+  import replacement, async work receipts and delivery retries are separate work.
+
+## Adding a handler and client
+
+```js
+// server/src/index.js — register via api, not app, and add its input case to
+// createValidator. Use ApiError only for a message known to be public.
+api.post('/api/example', async (req, res) => {
+  if (!available()) throw new ApiError(409, 'example-unavailable', 'Try after the current task finishes.');
+  res.json(await performValidatedAction(req.body));
+});
+```
+
+Promise rejections and synchronous throws enter one Express error path. For
+event emitters, explicitly forward errors and remove timers/listeners on close;
+the promise wrapper cannot catch unrelated future callbacks. Downloads pass
+their callback error to `next`; `pipeResponse` and `remoteStream` demonstrate
+the before/after-headers distinction. A late failure after a completed response
+is consumed, not sent again. Do not change status or append JSON to committed
+stream bytes.
+
+```ts
+// web/src/api.ts — the existing local fetch keeps attribution; json uses the
+// single decoder. For a text success, use decodeResponse(r, 'text').
+export const example = (body: ExampleInput) =>
+  fetch('/api/example', { method: 'POST', headers: HEADERS, body: JSON.stringify(body) }).then(json);
+// UI: catch ApiError; display .message, branch on .code, inspect .data.
+// Keep the draft and existing explicit failure actions; never replay here.
+```
+
+`decodeResponse` reads a failure body once, capped at 64 KiB, cancels oversized
+reads, and never shows a proxy's HTML/status text. Empty success is allowed;
+malformed expected-JSON success is `unreadable-response`. Legacy `{error:string}`
+still works. AbortError identity remains intact for Reader cancellation;
+`timeout`, `offline`, and `network-error` are distinct transport failures with
+no HTTP status unless headers were already received. XHR remains the upload
+transport: same decoder, progress, explicit cancellation and upload deadline.
+TraceUnavailable and RedactionBlocked remain subclasses with their existing
+constructors/fields; generic HTTP failures are not Reader empty states.
+
+Tests: `server/test/api-boundary.test.mjs` exercises the real Express stack with
+failure injection and fake action counters; `api-http.test.mjs` exercises the
+production routes with isolated roots and mocked external calls. The discovered
+web decoder and control tests exercise real fetch/XHR, Reader/file/share/group
+controls and draft retention. No test sends real agent work or notifications.
--- a/docs/managed-skills.md
+++ b/docs/managed-skills.md
@@ -1,0 +1,196 @@
+# Managed skills
+
+The manager owns a source file and explicitly recorded generated `SKILL.md`
+files. A matching directory name does not establish ownership. Support files,
+other installations and configured roots are never recursively deleted.
+Deletion is permanent. There is no trash, backup, snapshot, undo or content
+history in this service.
+
+## Identity and destinations
+
+`server/src/skills.js` is the shared lifecycle service for the HTTP routes,
+startup redistribution and generated `environment.md`. Its queue serializes
+these callers within one manager process, including their reads and revision
+checks. Do not run two managers against the same source/state roots.
+
+The installation ID retains the previous mapping: strip the last extension,
+lowercase, normalize punctuation/whitespace to `-`, trim edge hyphens, truncate
+to 40 characters. Invalid/path-like filenames and empty IDs are rejected.
+Distinct sources that collide after case, extension, normalization or truncation
+are rejected before either source or installation bytes change. The global
+session/group `slugify` function is unchanged.
+
+Distribution is still enabled only by `SPACE_ID` or `AM_DISTRIBUTE_SKILLS=1`.
+The destinations remain ordinary `HOME/.agents/skills`, Claude's configured
+skills directory, `HOME/.hermes/skills`, and the `.agents/skills` directories
+under configured Gemini and OpenClaw homes. There are no new destinations.
+The source, state and target roots are canonicalized; configured root aliases
+are supported and duplicate destinations are coalesced. Overlapping ownership
+roots are refused. Managed entries, including dangling symlinks and temporary
+entries, must not redirect through symlinks. Replacing an already resolved root
+with a redirect is also refused. These checks prevent accidental redirection;
+they do not isolate files against arbitrary hostile same-user filesystem races.
+
+## Ownership and existing installations
+
+The version 1 manifest is stored at `DATA_DIR/state/skills/skills-v1.json`.
+It binds the canonical source root and each source filename to its installation
+ID, a unique creation identity, current source digest, exact target roots/files, and generated-file digests.
+It records which skill directories the operation creates; adopted directories
+are never owned. Recorded paths are checked against current configured roots
+and reconstructed expected filenames before use. A changed source root, unknown
+manifest version, corrupt/unreadable manifest or invalid path disables mutations.
+An unwritable manifest cannot authorize a new operation.
+
+On first startup without a manifest, a unique valid source may be adopted.
+An existing destination is adopted **only** when its `SKILL.md` is byte-identical
+to the rendering of that source. Only that file is adopted, never its directory
+or support files. Missing destinations can be installed. A modified destination
+or ambiguous source makes the whole skill conflict before writes. Other skills
+continue independently, and the server reports degraded distribution without
+stopping its other functions.
+
+### Generated skills are read-only
+
+`environment.md` is derived from the Space configuration: the manager rebuilds
+it on every start and whenever secret descriptions or settings change. There is
+no authored content in it to protect, so it is **read-only**: `POST`, `PUT` and
+`DELETE /api/skills/environment.md` answer `403` with an explanation whatever
+revision is presented, and the editor offers no Edit, Save or Delete for it and
+says why. Keep your own instructions in a separate skill. The service is told
+which names are generated (`generated: ['environment.md']`); a record created
+by generation is read-only even before the name is configured.
+
+For a generated skill the **newest generated content is always authoritative**.
+The retry-with-the-same-content rule that protects hand-authored saves does not
+apply, because nobody can reproduce an earlier generation's bytes on purpose:
+when a generation finds the intent of an interrupted earlier one, it adopts the
+files that carry that intent's bytes (they were written by this manager), drops
+the intent, and publishes the new content. An installation that was modified by
+someone else is still refused by the ordinary ownership checks. Generation still
+cannot take over a user-created managed skill of the same name, and legacy
+environment content follows the same adoption checks, using the existing source
+to verify its installation before publishing new generated content.
+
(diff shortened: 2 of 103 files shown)


lvwerra (commented):
Request changes — blocking correctness and agreed-scope gaps.

1. **[Correctness — blocking, data loss] `web/src/components/FilesPane.tsx:676-685,723` lets an older file success erase a newer draft after navigation/unmount.** The path cleanup sets `pending.current = null`, so the still-running success callback can no longer tell that its submitted text is stale and executes `remember(sessionId, { draft: null })`. I reproduced this in the PR's browser harness: save H, type I while H is parked, close the file view, then resolve H; the remembered I draft becomes `null`. The pane close at line 1161 also bypasses `leaveView`, and `saving` is excluded from `unsaved` at 987, so this is reachable through normal UI navigation. It can also clear the draft for a different file opened in the same session before the old response lands. The committed suite only tests the response while the original viewer remains mounted, so it misses the destructive case the issue specifically calls out.

2. **[Correctness / scope adherence — blocking, cross-client lost settings] `server/src/index.js:1229-1263` and `1684-1693` still accept unconditional whole-resource replacements with no revision precondition.** I ran the branch server in disposable state with two copies of the same config: tab A changed the artifacts Space and received 200; stale tab B changed only the Jobs threshold and also received 200; the final GET had tab B's old artifacts Space, silently erasing A. `/api/secrets` has the same contract. This is explicitly required by #123 (including two-tab/API coverage and defined patch-vs-replacement semantics), despite the PR body saying it is out of scope. `web/src/api.ts:125-130,200-203` carries no revision either, and `SettingsView.tsx:217-219` discards the normalized committed response, so the client cannot maintain a server base or reconcile what actually committed.

3. **[Correctness / scope adherence — blocking, destructive recovery] `server/src/index.js:1183-1192` treats malformed or unreadable settings as empty defaults, after which an ordinary edit overwrites the original bytes.** I reproduced this with a malformed `am-config.json`: GET returned defaults, changing the Jobs threshold returned 200, and the malformed file (including its prior artifacts value) was replaced by the default-shaped config. `loadSecretNotes` does the same. #123 explicitly says malformed/inaccessible settings must be preserved and answered as an actionable error; this was not a deliberate exclusion available to the implementation.

4. **[Correctness / scope adherence — blocking] Settings save state does not outlive the Settings component.** The two savers are component-local at `SettingsView.tsx:186-229`; after unmount, `saveQueue.ts:90-103` deliberately suppresses the only status update while retaining a failed job inside the now-unreachable hook. I reproduced change → close Settings → reject the parked request: no persistent “not saved” indication or Retry remained, and reopening fetched the old server value into a new saver. This directly contradicts the required app-level late-failure indication, Retry route, and reopen behavior, and also contradicts the PR's claim that navigation does not discard pending changes. There is likewise no requested unload warning anywhere in the changed path.

5. **[Scope adherence — blocking] `web/src/lib/saveQueue.ts:81-105` has no finite timeout or lost-response reconciliation.** If a request never settles, `active.current` remains true forever, every later settings edit is trapped in the single slot, and the UI remains “saving…” indefinitely. If the server committed before the response was lost, a blind retry also has no revision/read-back mechanism to distinguish that from a failed commit. Both the timeout path and this reconciliation are explicit #123 requirements and required tests, not follow-up policy work.

6. **[Correctness] `SettingsView.tsx:481-485` persists a temporarily incomplete number as zero.** In the real browser harness I entered 17 and then cleared the field; the controlled input immediately displayed `0` and a PUT with `askAboveUsd: 0` was dispatched. That is exactly the accidental reset #123 says to prevent: an incomplete numeric edit must stay visible with validation and must not be marked/sent as a valid default.

7. **[Scope adherence] A non-conflict file failure has no visible Retry action.** At `FilesPane.tsx:1181-1183`, the error branch renders only the error text and suppresses the Save controls. I parked a file write, rejected it with “disk is full”, and confirmed the draft remained but `.files-info` contained no Retry button. The issue requires failures (not only content conflicts) to remain actionable with Retry.

8. **[Scope adherence] Derived-update failures are swallowed and presented as fully saved.** `server/src/index.js:1336-1345` catches every `generateEnvSkill` failure without retaining or reporting it, while the settings PUT has already answered success. This does coalesce successful regeneration, but it does not implement #123's required distinction between “settings persisted” and “derived update failed/pending”; the new server test only exercises the successful convergence case.

Verification: checked out `878b8b2`; focused browser race suite, focused server settings suite, typecheck, and all 24 normal web suites pass. The full server command reaches only the documented known-red `test/crons.test.mjs` #4. Mutation-checking away the mounted stale-buffer comparison makes the existing focused suite fail its two draft/status assertions, so that test protects the in-pane fix; it simply does not cover the unmount/resource-identity race above. The required concurrency, lost-response/timeout, corrupt settings, late-failure, numeric-input, and derived-failure cases are absent from the committed suites.

lvwerra:
All eight findings were real and all eight are fixed in `b84cdce`. Nothing pushed back on.

You were also right about the bigger thing behind them: I had treated the "Agreed behavior and PR boundary" summary as the whole binding scope and read the numbered Requirements as elaboration. They are not — §2 asks for the save lifecycle outliving the panel, the app-level late failure and the numeric-input rule; §3 asks for the revision precondition, the malformed-settings refusal, the finite timeout with read-back reconciliation and the derived-vs-persistence distinction; §4 asks for cross-file/remount isolation and Retry on any failure. The four things I listed as "deliberately left out" were four requirements I had misfiled. The boundary excludes backups/trash/history/restore, a diff or recovery UI, a global state rewrite, and #121/#122's territory — none of which this adds.

Evidence page, rebuilt with the new states and the mutation table: https://lvwerra-agent-artifacts.static.hf.space/save-ordering-123.html

## 1 — a file success erasing a newer draft after navigation/unmount (blocking)

Reproduced exactly as you described. The write now carries the resource it belongs to (`{ sessionId, path }`), captured when it goes on the wire, and the answer is judged against that rather than against whatever the viewer shows when it lands (`FilesPane.tsx`, `write`/`onCommit`):

- the remembered draft is released **only if that draft is the text that was committed** — same session, same path, same bytes. A newer draft for the same file is kept and rebased onto the version that just landed; a draft belonging to another file is not touched at all.
- viewer state (`meta`, `baseRef`, `committed`, status) is only updated while the viewer is still on that file.
- a save asked for before navigating still completes: with the viewer gone, `write` takes the text from the remembered draft rather than dropping it.

Tests, in `web/test/saveRaces.test.mjs`:

- *a file save that lands after the view is gone* — save H, type I, `setShowFile(false)`, resolve H: draft is still I, rebased to `tag-H`; reopening shows I, still dirty; saving it writes I against `tag-H`.
- *a file save that lands while another file is open* — save J, switch to `/other.txt`, type there, resolve J: the other file's draft, text and status are untouched, **and its next save carries its own base (`tag-0`), not `tag-J`**.

Mutation: unconditional `remember(sid, {draft:null})` → 8 checks fail. Dropping the `if (!here) return` guard → the base check fails (that one initially passed with only the draft assertions, which is why the base assertion is there — thank you for the specific repro, it made the weaker version of the test obvious).

On the two secondary points: `saving` is now part of `unsaved`, so closing during an in-flight write goes through the confirm dialog. I left the pane's ✕ (line 1161) calling `onClose` directly — the buffer survives it through the existing draft store, and routing it through `leaveView` would change pane-level behaviour beyond this issue.

## 2 — no revision precondition (blocking)

`GET /api/config` and `/api/secrets` now return `rev`, the hash of the committed bytes (not mtime — the FUSE mount rewrites those). `PUT` takes `?base=…`, the same shape the file `/write` route already uses:

- no `base` for a resource that exists → `409 {code:'base-required'}`
- stale `base` → `409 {code:'stale', rev, value}`, carrying what is actually stored
- read-check-write runs synchronously inside one handler, so nothing lands between the check and the rename

Your two-tab repro is now a server test: A commits, B (holding the old rev) is refused, a `GET` still shows A's artifacts Space, and B succeeds once it replaces the revision it was told about. The client keeps the rev per resource, sends it with every save, and shows a conflict as *changed elsewhere* with **Keep mine** (re-send against the reported revision — a click, never automatic) and **Use theirs** (adopt the stored value into the panel). `SettingsView` no longer discards the committed response: it takes `rev` and `derived` from it. It deliberately does *not* adopt the normalized committed *value* into the editing buffer — that would put a response into a field the user may be typing in, which is the same class of bug as the rest of this PR. Semantics are documented in `docs/saving.md`: PUT is whole-resource replacement, and the precondition is what preserves unrelated fields.

## 3 — malformed settings overwritten by defaults (blocking)

`readSettingsFile` now separates *missing* (defaults, writable) from *unreadable* (damaged JSON, bad permissions). A read reports `readError` and the panel says so; a write is refused with `409 {code:'unreadable'}` and the bytes are left exactly as they were. Internal consumers still take defaults so nothing throws. Server test: a hand-written config with a syntax error stays byte-identical after an ordinary edit is attempted. The derived pass refuses to publish from a settings file it could not read, for the same reason.

## 4 — save state dying with the panel (blocking)

The savers moved to `web/src/lib/settingsSaves.ts` and live for the page, not the component. Closing Settings cannot cancel work already asked for; a failure that arrives afterwards is reported by `SettingsSaveAlert` in the app chrome, with *Open settings* and *Retry* (or *Keep mine* for a conflict); reopening shows the value still owed rather than fetching over it. `beforeunload` warns while anything is outstanding, and `docs/saving.md` states plainly that this is a prompt and not a guarantee. Tested end to end: change → close → reject → alert with both buttons → retry from the alert → reopen shows `23`, not the server's older value.

## 5 — no timeout or lost-response reconciliation (blocking)

`saveQueue` takes `timeoutMs` (15s settings, 30s files) and a `reconcile` hook. On timeout the slot is released — so later edits are not trapped — and the state is *not confirmed*, not *failed*, because a lost answer may mean the server committed. Before anything is sent again, `reconcile` reads the resource back: if what is stored is what was sent, it is recorded as committed and **not** rewritten; if not, the value is sent again through the ordinary precondition path. Never a forced overwrite. Both branches are tested, plus a check that the resource actually carries a finite window (the behavioural test overrides the timeout to 250 ms, so without that check a `TIMEOUT_MS = 0` mutation slipped through — it does not now).

## 6 — an incomplete number saved as 0

The field keeps a text draft; the setting keeps its last valid value. An empty or invalid box is shown as typed, marked, and explains itself ("Not a number yet — the saved threshold is still $17"); nothing is dispatched until it parses again, and blur snaps back to the committed value. Tested including the assertion that the stored value is still 17 while the box is empty.

## 7 — no Retry on a non-conflict file failure

The error branch of the info strip now renders Discard + Retry beside the message. Testing it needed the pane, not just `FileView`, so the harness mounts a real `FilesPane` on its own session: park a write, reject it with "disk is full", assert `.files-info` contains a Retry, click it, and assert the resend carries the same text **and still carries its precondition** — Retry is not a quiet overwrite.

## 8 — derived failures swallowed

`refreshEnvSkill` keeps `{pending, error, at}`, `generateEnvSkill` no longer swallows its own write failure, and both settings reads report `derived`. The panel distinguishes the two states: "Saved — but the `environment` skill agents read could not be updated: …". Server test makes the skill file unwritable, confirms the settings save still succeeds and really persisted, that the derived failure is reported rather than hidden, and that it clears once a later save gets through.

## Runs

- `web/test/saveRaces.test.mjs` — 95 checks, was 43. Ran repeatedly, stable.
- `server/test/settings-write.test.mjs` — 30 checks, was 13.
- Full web suites (24), `npm run test:render`, every server suite individually, `tsc --noEmit`, `vite build`.
- `server/test/crons.test.mjs` #4 still fails here and on stashed clean `main` at ef08e84. Not mine, not touched.
- New: `docs/saving.md` — immediate vs explicit policy, revision/error semantics, pending-operation lifetime, browser-termination limits.

One thing worth a fresh reviewer's attention: the file saver's timeout and reconciliation share the queue with settings, but only the settings path is covered by a behavioural test (the file `reconcile` reads the preview back and treats an `html` file — whose source is not in the preview payload — as "lost", which then goes through the base precondition and surfaces a conflict rather than overwriting). That is deliberate, and it is the thinnest-covered part of this change.

lvwerra (commented):
Second independent pass on `b84cdce17a686fce4e15b694cc70b465f40a5fa5`. **Changes are still needed.** These are additional failure paths through the revised implementation, not objections to the eight fixes accepted in round one.

1. **P1 — Timeout reconciliation silently overwrites another client's settings.** [settingsSaves.ts:79–85](https://github.com/huggingface/agent-manager/blob/b84cdce17a686fce4e15b694cc70b465f40a5fa5/web/src/lib/settingsSaves.ts#L79) installs the read-back revision *before* checking whether its value matches the missing write. A mismatch is then classified as `lost`, and the old whole-resource value is automatically retried against that newly acquired revision. Reproduced with the actual browser API client and isolated server: let my threshold save commit but withhold its response; another client saves `artifacts.space = 'other-client/valuable'`; time out and click ordinary **Retry**. The second PUT succeeds, erases that Space, and the UI reports `saved ✓`. No Keep mine confirmation occurs. The secrets path has the same unconditional revision adoption at line 96. A different read-back value is not evidence that the first write never landed, and cannot authorize replacing a later writer's version.

2. **P1 — Reopening Settings during an active write loses the active value and permits an older field value to overwrite it.** [saveQueue.ts:231](https://github.com/huggingface/agent-manager/blob/b84cdce17a686fce4e15b694cc70b465f40a5fa5/web/src/lib/saveQueue.ts#L231) exposes only the waiting slot/unresolved value through `pending()`; the active job was removed from the slot at line 176. Consequently, `pendingSettings()` returns null while the first save is in flight, and [SettingsView's mount read](https://github.com/huggingface/agent-manager/blob/b84cdce17a686fce4e15b694cc70b465f40a5fa5/web/src/components/SettingsView.tsx#L272) fetches over it. Reproduced against the real endpoints by holding the outgoing PUT: change the threshold from 0 to 11, close/reopen Settings before the request reaches the server (the input now shows 0), edit only Artifacts Space, then release the first write. The two successful PUTs store threshold 11 and then threshold **0**; the latter uses the first response's new revision, so there is no conflict. The same hole exists for descriptions. The page-owned state must retain the active desired value as well as queued/failed work.

3. **P1 — A queued normal file Save persists text typed after that Save, then clears its draft and calls it saved.** [FilesPane.ts:658–668](https://github.com/huggingface/agent-manager/blob/b84cdce17a686fce4e15b694cc70b465f40a5fa5/web/src/components/FilesPane.tsx#L658) queues only resource/force information and reads the live buffer when dispatching. In the actual FileView: type A and Save; while A is parked, type B and Save; type C **without another Save**; resolve A. The second request contains C, not B. Resolving it removes C's remembered draft and reports `saved`. §1 explicitly distinguishes normal Save from Save and close: normal Save must commit the requested revision, with subsequently typed text remaining dirty. The queued request needs to retain that requested version.

4. **P2 — Save and close still closes over newer unsaved text after the operator chooses Keep editing.** [FilesPane.ts:1059–1070](https://github.com/huggingface/agent-manager/blob/b84cdce17a686fce4e15b694cc70b465f40a5fa5/web/src/components/FilesPane.tsx#L1059) does not cancel the outstanding close intent or check the current buffer after awaiting `saveNow()`. Reproduced through the real FilesPane buttons: type A, Back, Save and close; hold A's write; click **Keep editing** and type B; resolve A. The viewer closes anyway. B remains only in the remembered draft. Separately, `saveNow()` resolves `true` while FileView's status is `dirty` and the draft is B. An old successful request is neither permission to close a newer buffer nor permission to reverse Keep editing.

5. **P2 — A hung reconciliation read wedges the resource permanently after the write timeout.** [saveQueue.ts:137–145](https://github.com/huggingface/agent-manager/blob/b84cdce17a686fce4e15b694cc70b465f40a5fa5/web/src/lib/saveQueue.ts#L137) sets `active = true` and awaits `options.reconcile()` without the finite timeout used for `send()`. Time out a write, click Retry, and leave the reconciliation read pending. Subsequent valid edits remain queued forever, status stays `saving`, and the Retry control disappears. A deferred-read probe remains stuck well beyond the configured timeout with the newer value waiting and no next write. The finite failure path must include the read-back step, for files and settings alike.

6. **P2 — Late derived-update failures are recorded by the server but never shown by the open Settings panel.** [DerivedNote](https://github.com/huggingface/agent-manager/blob/b84cdce17a686fce4e15b694cc70b465f40a5fa5/web/src/components/SettingsView.tsx#L60) ignores `pending`, and the client does not fetch a settled derived outcome after a PUT. In a real server/browser run, make the disposable environment source unwritable and edit the threshold. The successful PUT reports `{pending:true,error:null}`; a subsequent independent GET reports the EACCES failure, but the untouched panel still says `saved ✓` and has no derived warning. The existing server test polls the GET itself, which hides the missing production caller. This remaining part of §3 needs a way to deliver the eventual pending/failure outcome to the UI without resubmitting an already committed settings write.

7. **P2 — The descriptions endpoint accepts a payload that its own reader refuses, replacing all existing descriptions and stranding ordinary saves.** [index.js:1822–1834](https://github.com/huggingface/agent-manager/blob/b84cdce17a686fce4e15b694cc70b465f40a5fa5/server/src/index.js#L1822) accepts arrays because `typeof [] === 'object'`. With valid existing notes and their current revision, `PUT /api/secrets?base=<rev>` with `{"notes":[]}` returns 200 / `ok:true` and writes `[]` over the notes file. The next GET reports `secret-notes.json does not contain a settings object`; subsequent ordinary saves are refused as unreadable. This is a new interaction between the permissive writer and the added damaged-file protection. Validate the notes object's shape and value types before replacing committed bytes; the required request-validation coverage currently misses this case.

Verification and scope:

- Checked out and ran the exact head. The issue's cited locations still identify the intended code on base `ef08e84`; main has not moved. No new backups, trash, history, restore, global store migration, or automatic one-shot actions found. The scope problems above are incomplete required save behavior.
- The committed browser race suite passes, all **24 default web suites pass**, and `npm run build --prefix web` passes. No readerRedesign failure in this run.
- The settings server suite passes **30/30** in isolation with a dummy `HF_TOKEN` supplying the injected name its derived-description assertions assume. With all inherited credentials removed and no dummy, those two assertions fail; this is a test-fixture dependency, not a settings regression. All test roots were disposable and distribution was disabled.
- `npm test --prefix server` reaches the documented baseline `crons.test.mjs` #4 failure after six passing suites; I do not attribute that failure to this PR.
- Mutation-checking the mounted stale-buffer guard in a disposable copy produces a real failed saved-status assertion, then the expected missing-follow-up-write failure. That protection is real; the passing committed suite does not cover the seven sequences above. The additional browser races, actual-endpoint timeout loss, and pane close scenario were repeated in isolation.

No author-branch edits or pushes.

lvwerra:
All seven fixed in `6303a3c`. Nothing pushed back on — each one is a real path to lost work through the round-one code, and two of them (1 and 3) are the core invariant leaking out of the places I had guarded it.

Evidence page rebuilt, including a mutation table for both rounds: https://lvwerra-agent-artifacts.static.hf.space/save-ordering-123.html

## 1 — reconciliation overwriting another client (P1)

The read-back adopted the revision before deciding whose version it belonged to, which turned "somebody else wrote while we waited" into a fresh base for our old value. `reconcileWith` in `settingsSaves.ts` now decides first, and has three answers instead of two:

- read-back **equals what we sent** → it was ours, it landed: adopt the revision, record it saved, do not repeat it.
- read-back **differs, revision unchanged** → nothing was committed: send it again against the base we already hold.
- read-back **differs, revision moved** → somebody else is there. `Reconciled` gained a `conflict` outcome; the saver reports it through `onFail` exactly like a 409, the value stays put, and *Keep mine* / *Use theirs* is the operator's call.

Both resources share that function, so the secrets path is fixed with it. Test — *a lost answer while somebody else was writing*: park the threshold write, have another client commit `artifacts.space = 'other-client/valuable'` at a new revision, time out, click ordinary **Retry**. No second PUT goes out, their Space is still on the server, and the flag reads *changed elsewhere*; only then does *Keep mine* send, against the revision the check found. Mutation (adopt the revision first, always classify `lost`) fails two checks.

## 2 — reopening during an active write (P1)

`pending()` reported the queued slot and the unresolved value but not the job on the wire, so `pendingSettings()` was null exactly while a write was out. The saver now tracks `inflight` and `pending()` returns the newest value it is trying to get to the server, whichever stage that is. Test — *reopening settings while a write is still out*: threshold 0 → 11, park it, close and reopen (the field shows 11, not 0), edit the Space, release the first write; the follow-up PUT carries **both** fields, and what ends up stored is 11. Mutation (drop `inflight` from `pending()`) fails three checks, including the stored value.

## 3 — a queued Save persisting text typed after it (P1)

You are right that this is §1's normal-Save/Save-and-close distinction, and reading the live buffer at dispatch collapsed the two. The request now carries the text captured when the operator asked; only the base is re-read as the write goes out (an earlier save of the same file may have committed meanwhile). Test — *a Save saves what was asked for, not what is typed after it*: type A → Save; type B → Save; type C with no Save; resolve A. The queued request contains **B**, and after it lands C is still dirty, still drafted, still on screen. Mutation (read `pending.current` again) fails three checks.

## 4 — Save and close after Keep editing (P2)

Two independent holes, so two guards and two tests:

- `saveNow` now answers the stricter question — `ok && !pending.current` — so a write that succeeded for older text comes back **false**. Tested directly on the `SaveState`: type X, `saveNow()`, type Y, resolve X → `resolved:false`, status dirty, draft Y.
- `saveAndClose` records the close intent and `backToEditing` withdraws it, so an answer arriving after **Keep editing** cannot close the view. Tested through the real pane buttons with nothing typed afterwards, which is the case the first guard does not cover.

Reverting either one alone fails its own check.

## 5 — a hung reconciliation read (P2)

`withTimeout` is generic now and wraps the read-back too; on timeout the state stays *not confirmed* and retryable with the slot released. Test — *a check that is itself unanswered*: hold the read, Retry, and after the window the saver is out of `saving`, still outstanding, and a later edit dispatches with the newer value. Mutation (unwrap the reconcile call) fails four checks.

## 6 — late derived failures never reaching the panel (P2)

You are right that the server test was checking the server, not the caller. There is now `GET /api/settings/derived`, and the store follows up on any commit whose response says `pending` (150 ms, then 500 ms, bounded at 12 tries) — a status check, not a resubmitted write. `DerivedNote` renders the settled error, and says "still updating…" if a pass stays pending past a couple of looks. Browser test: settle a save with `derived:{pending:true}`, let the pass fail afterwards, and the open panel shows *could not be updated* with **no** second PUT. The server test now reads the dedicated endpoint too.

## 7 — the descriptions endpoint accepting an array (P2)

`readNotesPayload` validates before anything is replaced: object (not array), string values, bounded key and value lengths → `400 {code:'invalid'}`. `/api/config` rejects a non-object body the same way rather than normalizing it into a file of defaults. Test: `{"notes":[]}` at the current revision is refused, the file is byte-identical, the next GET still reads, and an ordinary save still works. The mutation reproduces your chain exactly — 200, then every later read and save refused as damaged.

## On your fixture note

Confirmed and fixed: those two assertions were riding on an inherited `HF_TOKEN`. The suite now injects its own `SETTINGS_WRITE_PROBE_KEY` into the spawned server and describes that, plus a check that it is detected so a future failure reads as "no key detected" rather than as a settings regression. `env -i` with nothing inherited: 38/38.

## Runs

- `web/test/saveRaces.test.mjs` — 129 checks (was 95).
- `server/test/settings-write.test.mjs` — 38 checks (was 30), green with a stripped environment.
- Full web suites (24), `npm run test:render`, every server suite individually, `tsc --noEmit`, `vite build`.
- One flake worth naming: `readerRedesign.test.mjs` failed once in a full-suite run while the box was at load ~10, then passed in isolation on this branch and on stashed clean `main`, and passed in the next full run. Not attributable to this PR.
- `crons.test.mjs` #4 unchanged and untouched.
- `docs/saving.md` updated: the three-way reconciliation, the finite window covering the read-back, normal-Save versus Save-and-close, the derived status endpoint, and the payload rules.

lvwerra:
Contract-collision follow-up pushed at `ca9cde2`.

This branch now owns the settings consumer half and is self-contained against `main`:

- both settings routes emit generic 5xx prose plus the failed `DATA_DIR`-relative filename in `details: [{ field: "path", message }]`;
- the settings client reads that structured field and turns it into `am-config.json could not be saved.` (or `secret-notes.json …`), while a response without the field keeps the generic fallback;
- the failed filename is visible beside Retry both inside Settings and in the app-level alert when Settings has already been closed;
- the old assertion expecting a filesystem/path-bearing free-text message is replaced by assertions for generic prose, the structured relative path, no absolute path/exception text, and unchanged on-disk bytes. Both settings files are covered.

I chose the data-root-relative filename rather than an absolute container path: it carries the useful distinction between the config and descriptions files without exposing an unstable internal mount path.

Standalone verification on this PR head: `settings-write.test.mjs` 41/41, all 25 web suites, TypeScript, and the production build pass. Mutation checks separately catch removing the producer field, the client extraction, and the visible in-panel diagnostic.

PR #144, opened 2026-09-09
Reader: load a useful amount of recent history on a cold open

### LINKED ISSUE #129: Reader: search the full conversation and load more history by default
# Reader: search the full conversation and load more history by default

## Goal and agreed scope

The operator approved whole-conversation Reader search and added:

> “in any case i want the loading of history a bit more extensive, at the moment it feels like to often i only see 1-2 messages by default”

Deliver these together in **one focused PR**: a useful amount of recent conversation loaded automatically, and an explicit way to search older messages without manually paging through the transcript. Keep Reader responsive and independent of the terminal. This is an issue specification, not a claim that either change has been implemented.

The planning default is **about 20 recent prompt/reply exchanges**, progressively loaded. This is an implementation starting target, not a count specified by the operator or permission to read unlimited bytes to reach it. Validate and tune against representative fixtures. An exchange means an actual user prompt and its associated work/reply, not two raw trace records. The history should be available in scrollback; do not shrink text or change disclosure defaults to fit 20 exchanges on screen.

**Additional operator requirement:** the first rendered conversation should cover at least **one Reader viewport** when enough history is available, so already-visible text does not move downward while older turns are loaded or measured. Viewport coverage and the approximately 20-exchange history target are separate criteria: a tall answer can fill a page without providing enough history, and many tiny exchanges might still not fill a tall viewport.

Search is limited to the **selected conversation's available transcript**. Keep existing loaded-history search. No fleet-wide search, recursive search of separate child conversations, new settings panel, background indexing of every session, or broad Reader rewrite.

## Current evidence and integration points

Planning checked `main` at `ef08e843d1c67fcc4e1c463415819ea351734d6d` on 2026-09-09. Recheck current main and related work before implementation.

- [ReaderStore](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/lib/readerStore.ts#L4) currently requests an initial **128 KiB** tail with `INITIAL_WINDOW_TURNS = 2`; ordinary pages request 384 KiB. Retaining an already-loaded store refreshes forward rather than automatically filling older context.
- [Server byte windows and index windows](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/server/src/traces.js#L1902) interpret that minimum differently. Byte windows grow until the parsed-record floor, source start or 8 MiB per-window ceiling; `windowIndex` takes exactly the last `min` message positions on a tail request. With `min=2`, an indexed source initially returns at most two transport messages. Raw records can also be tool/result/lifecycle fragments rather than readable exchanges. These are source findings, not a fresh reproduction of the operator's particular session.
- [ConversationView](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/conversation/ConversationView.tsx#L70) searches only loaded exchanges through [readerSearch](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/conversation/readerSearch.ts). It **already** says “Search loaded conversation” and discloses that earlier history has not been searched. Preserve that truthful behavior; missing scope disclosure is not the reported defect.
- Reuse `readerModel.ts` reconciliation, `exchanges.ts` grouping, `useVirtualRows.ts` and `readingPosition.ts`. The source adapters are in `traceWindows.ts`, `api.ts`, `index.js` trace routes, `traces.js` and `trace-revision.js`. Current full parsers have message/text display caps; searching their clipped output alone cannot establish a complete no-match result.
- [Reader architecture](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/docs/reader-architecture.md) and [review follow-up](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/docs/reader-review-followup.md) document the contracts merged in [#119](https://github.com/huggingface/agent-manager/pull/119). Inactive retained stores currently have an eight-entry / approximately 32 MiB budget; database/full-summary parsing remains a known cost. Do not assume merely enlarging a global transport constant solves the user-facing history problem.

### 1. Fuller initial history without a slower usable Reader

- On an ordinary cold visit, automatically make roughly 20 recent exchanges available, or all available exchanges when the conversation is shorter. Count reconciled, visible conversation exchanges, not tool records, lifecycle events, raw API turns or DOM nodes. An active last exchange need not be complete; never wait for the agent to finish before showing history.
- Render the usable composer promptly and prepare a stable first transcript viewport as specified below; once that viewport is ready, fill the remaining recent-history target in bounded background steps. Do not wait for all 20 exchanges or a whole-transcript summary before presenting a usable first page. Initial fill must not depend on a wheel event, clicking Earlier or opening Search. A summary failure/hang must not prevent a usable recent history.
- Choose and document byte, retained-data, request-count and work/time limits for automatic fill. On tool-heavy or huge-message traces, reaching a limit may yield fewer exchanges; disclose remaining history and keep Earlier usable. Do not loop forever trying to reach a count when the cursor cannot advance or an oversized record blocks progress. Preserve existing per-record protections or justify any scoped adjustment with tests.
- Prefer complete prompt/work/reply boundaries within the budget. Prepending must reconcile partial messages/results without duplicates, missing answers, fabricated completion or reordered tools. Do not discard relevant work records simply to make the exchange count look larger.
- Apply the selected-session policy consistently across supported byte-oriented and indexed/database traces. Inventory shared hook consumers: do not blindly change child readers' explicit top-of-context behavior or multiply automatic loading across collapsed Overview previews. Only active visible readers do automatic work; shared consumers must not duplicate it.
- Retain already-loaded history on warm return. If the desired recent context is still missing, fill only the missing part; do not throw away a warm store or repeat the entire preload on every mount, poll, visibility event or mode switch.
- Preserve newest user intent during prepends: Latest stays at the current bottom through late measurements, while a deliberate older reading position stays on the same message/offset. Coordinate with saved-position restoration so automatic fill neither consumes its retry budget incorrectly nor moves the user back to a stale anchor. Preserve draft, attachments, selections and disclosure state.
- User navigation/search-hit loading and normal live updates must not be starved by speculative history fill. Cancel or pause unnecessary work on source changes, hidden/inactive readers or disposal, respecting other active subscribers. Failure must leave existing text and composer available and release busy state.

#### First-viewport coverage and visual stability

- Measure the actual conversation scroller's usable height, excluding the composer, headers and other fixed controls, at the current width, zoom and normal disclosure state. On an ordinary cold load with sufficient readable history, make at least that much real conversation content available for the first presented transcript viewport. Continue beyond 20 exchanges if unusually short messages require it, within the documented safety bounds. A count of transport records, estimated virtual heights, blank spacers or a skeleton does not prove that a page of conversation is available. Keep virtualization; do not mount the entire loaded history just to measure it.
- Prioritize obtaining and measuring that first page, then progressively fill the larger history target. A brief stable loading surface is acceptable during bounded first-page preparation; do not flash a top-aligned one-message transcript and then push it down with subsequent prepends. Do not add an arbitrary sleep, wait for the entire preload, hide already-readable warm content, or block the composer to achieve apparent stability. An actually short/empty conversation must render its available content/empty state promptly. A read failure, no-progress cursor or safety limit must end preparation with a stable, honestly partial view rather than an indefinite spinner or a false claim of full coverage.
- Establish the initial semantic anchor/follow intent and scroll geometry before exposing the first transcript paint. Subsequent older-history prepends and virtual-row measurements must preserve the on-screen position of already-visible text, including the transition from content shorter than the viewport to scrollable content. Apply necessary scroll compensation before paint, not a visible jump followed by correction on a later frame; do not smooth-scroll preload corrections. Reserve known media dimensions and handle font/image/layout completion without resetting the anchor. Preserve a remembered manual position when one exists, and let new user scrolling override pending automatic placement.
- Re-evaluate coverage if the viewport changes during initial loading, with coalesced and bounded work rather than a resize-fetch loop. Preserve stable anchors after initialization as well. Distinguish older-history layout corrections from genuinely new live output: normal Latest-follow behavior for a new reply remains valid, but delayed rendering of already-existing older turns must not move the current reply down the screen.

### 2. Explicit whole-conversation search

- Keep the fast loaded-history search and add a clearly named whole-conversation scope/action. The user must deliberately request whole-history work; no full scan on each normal keypress or Reader mount. Define when an edited query is submitted, cancellation, and how Enter/Shift+Enter navigate results without accidentally starting duplicate scans.
- Use literal, case-insensitive text matching initially, consistent with the existing Reader's searchable text. Search readable prompts, assistant content and tool text/names already exposed by Reader search; exclude encoded image data and unrelated transport metadata. Do not substitute searching raw JSON syntax for searching the displayed conversation. Document the exact content scope and any format-specific limitations.
- Find matches outside the currently loaded window, including earlier than summary/parser display limits where the source still contains readable text. Do not silently skip text after a display truncation and then report a complete search. If some portion genuinely cannot be searched safely, report incomplete coverage and its reason.
- Support the local session transcript formats already supported by Reader, including JSONL and indexed/SQLite sources. Reuse the existing selected-source resolver, pins, source identity and conversation boundaries; never accidentally search another conversation in the same database or a guardian/fork trace. Existing unsupported sources keep an explicit unsupported state. Shared file/bundle/child views must remain compatible; adding new search entry points to those separate surfaces is not required here.
- Return bounded, paginated results with useful text snippets and stable message/context locators. State whether counts are matching messages/exchanges or occurrences; do not present a capped page count as a full total. Empty results, still searching, canceled, failed, unsupported and partially searched are distinct outcomes. Only a completed search of its stated scope may say no matches were found in that scope.
- Define a finite scan boundary for a live conversation so continuing output cannot make a search run forever. Make freshness/coverage explicit if new output arrives. Bind requests/results to query, session/source generation and the relevant source revision or snapshot; old completions must not overwrite a newer query. Append-only growth need not invalidate valid old hits, but replacements or incompatible mutable-row changes must not silently retarget them.
- Validate query and continuation inputs and resolve sources on the server. Scope any search cache to the exact source/query/version, bound its memory and lifetime, and invalidate it correctly. Do not put transcripts, queries or snippets into diagnostic/audit logs, URLs, analytics or new persistent browser storage. No external search service or new database infrastructure is needed by default.

### 3. Navigate to old results without breaking live reading

- Selecting a hit loads a bounded window around that message and brings the match into view with surrounding conversation. The user must not need to download every intervening page from the live tail first. Do not splice separated history windows together as if the missing interval were continuous.
- Keep result identity stable across pagination and deduplicate cross-window fragments. Preserve existing message/tool reconciliation and sanitized Markdown/highlighting. If a match is in text beyond the ordinary display cap, show a bounded matching excerpt/context or clearly explain the display limit; do not navigate to a row where the match cannot be found without explanation.
- Result loading must not corrupt live-tail cursors, mix source generations, mark a partial context window as the full conversation, or overwrite the retained pre-search view. Use an explicit historical window/navigation state or another small design with equivalent guarantees rather than overloading the contiguous live-history store.
- Clearing/leaving search restores the original semantic reading position or current Latest-follow intent; an explicit Latest action wins over old search restoration. New direct scrolling cancels stale programmatic targets. Handle variable-height rows, images, resize, narrow viewports and virtualization without moving the terminal host or page ancestors.
- Keep search/results keyboard accessible and reuse existing Reader styling/controls. Preserve the composer and attachments throughout searches and failures. Loading an old hit is read-only: it must not start, attach, claim, resize or type into a PTY.
- Coordinate [#128](https://github.com/huggingface/agent-manager/issues/128): searching or viewing an old reply must not mark a newer unseen reply as read. Automatic history preload is not evidence of reading either.

### 4. Resource and lifecycle design

- Keep first paint, recent-history fill, whole-history search and match-context loading as distinct operations with clear ownership and priority. Reuse the established store/request/generation contracts; do not start a second uncontrolled poll loop or make every source adapter implement inconsistent cancellation rules.
- Bound work on both client and server. A browser AbortSignal alone is not enough if a synchronous server parse keeps blocking the manager. Use incremental/asynchronous scanning, query-level paging or narrowly scoped worker offload where needed for the supported source. Optimize only the parsing/search paths necessary for this feature; a general parser rewrite is not required.
- Bound concurrent scans/results/retained contexts across several open panes. Superseded work must stop consuming resources, not merely have its result ignored forever. Repeated queries may reuse a valid bounded cache; opening every session must not create a persistent index or trigger a full scan.
- Reuse current error/empty/reconnect handling and maintain one scroll owner and the virtualized DOM bound. Keep inactive-reader eviction effective with the increased preload; do not silently remove cache budgets or retain unlimited search histories.

## Required tests and evidence

Test real source/store/view integration as well as helpers. Existing browser fixtures sometimes return all synthetic turns regardless of `bytes`/`min`; add realistic fixtures that honor window requests so a passing test cannot hide the current two-message default.

| --- | --- |
| Default context | Cold-open JSONL and indexed/SQLite conversations with 0, 1, 2, 5, 20 and many exchanges. Ordinary fixtures with sufficient history reach the documented recent-context target automatically, without scrolling/search or waiting on a summary. Shorter conversations load all available content. Assert logical exchanges, not raw record count. |
| First viewport | Cold-open with very short turns, one tall answer, collapsed tool-heavy exchanges and varied-height content at desktop, tall/narrow mobile and zoomed sizes. When sufficient readable history exists, the first presented transcript covers the actual usable Reader viewport; test a case needing more than 20 tiny exchanges and another where one answer fills a page but the broader history target must still load. Short/empty sources, blocked records, slow reads and limits settle honestly without indefinite waiting. Blank padding or estimated spacer height must not satisfy the assertion. |
| No initial text drift | Hold the source fixed and delay older pages, virtual-row measurement, fonts and images separately. Track the screen coordinates of a stable visible text anchor from the first transcript paint through successive prepends, including an underfilled-to-scrollable fallback transition. Assert no history-induced displacement beyond a small documented tolerance (target 2 px) on intermediate visible frames, not only the final position. Cover a user scroll or resize during preparation, remembered manual anchors and a warm return; stale corrections must not override newer intent. Test genuine live appends separately so expected Latest-follow movement is not confused with the regression. |
| Tool-heavy and large records | Many tool/lifecycle records per prompt, split prompt/answer fragments, multiline messages, large tool output, empty trailing records and a single oversized record. Useful context increases within documented budgets; no duplicate/missing/reordered content, infinite paging, false beginning or fake completion. |
| Progressive responsiveness | Delay initial response, summary, later preload pages and one entire reader independently. First readable content and composer remain usable; one pane cannot block another. Cancel/fail preload and verify the busy slot releases, retained text survives and Earlier/search/live refresh still work. |
| Warm history and scroll | A/B/A navigation, reload with a saved older anchor, actual Reader/Terminal round trips, foreground return, new replies during fill and late layout changes. No redundant cold refill, lost draft or jump away from manual position; Latest stays at the geometric bottom. |
| Search correctness | Unique match only near the start of a transcript larger than the initial/history budgets; matches crossing scan chunks or message fragments; quoted/Unicode/literal punctuation; prompt, assistant and tool text; duplicate content/timestamps; matches beyond ordinary display caps. Compare results with an independent fixture oracle, with consistent counts and no duplicated page-boundary hits. |
| Coverage and pagination | No matches, many matches exceeding a result page, bounded-scan continuation, unsupported format, unreadable/truncated/oversized source, canceled scan and server failure. Never label partial coverage as a complete empty result; page through deterministic results without omissions. |
| Source isolation and races | Switch session or query mid-scan; resolve canceled requests late; replace/shrink a trace; append output during a search; mutate a SQLite streaming row and WAL. Results/cursors stay tied to the right source and query, and old hits are verified rather than silently opening different content. |
| Result navigation | Jump directly to a far-old hit without loading the intervening transcript into the browser; verify surrounding prompt/answer context, sanitized highlight and bounded DOM/data. Clear search, choose Latest, scroll during pending navigation and revisit a hit. Correct anchors and live-tail continuity survive. |
| Shared surfaces and Unread | Multiple active subscribers, hidden/released stores, children with their existing 2 MiB/top-of-context policy, Overview previews, file/trace consumers, attachments and input-required notices. No duplicate speculative readers, new PTY/socket/input/resize, terminal bleed or read acknowledgements from background preload/old search results. |
| Resource limits | Large synthetic byte and database histories, repeated queries, many open panes, slow I/O, rapid cancellation and cache eviction. Assert bounded request counts/bytes/results/DOM and cleanup; measure browser responsiveness and server event-loop impact, not just response size. |

- Extend/add normally discovered suites alongside `web/test/readerModel.test.mjs`, `readerRedesign.test.mjs`, `traceWindows.test.mjs` and `exchanges.test.mjs`; server coverage should include `reader-protocol.test.mjs`, `trace-window.test.mjs`, `trace-tail.test.mjs` and relevant source adapters. Add actual endpoint/browser wiring coverage, not only unused helpers.
- Record before/after under the same synthetic fixtures: time to usable composer and first stable viewport, actual first-viewport coverage and maximum history-induced anchor displacement, time and bytes to the recent-context target, number of available exchanges, time to first/complete search result, hit navigation latency, mounted rows and server responsiveness. Include cold/warm, tool-heavy, large JSONL and SQLite cases. Set reproducible regression budgets and explain any tuning away from the approximately 20-exchange starting target. Do not trade away first-page responsiveness by waiting for all background history merely to eliminate visible drift.
- Run focused new suites, normal web/server suites, `npm run test:render` and `npm run build` in `web`, plus relevant full-app Reader checks. Report exact commands/results. If a baseline suite fails (including the previously documented date-sensitive cron test), verify/report it separately and run otherwise-skipped relevant suites; do not hide unrelated failures or fix unrelated code in this PR.
- Use disposable sources, databases, state roots and isolated test servers with fake transports. Never send paid prompts, search private operator transcripts for fixtures, write installed skills, restart production or deploy. Performance diagnostics should contain counts/timings only, not user conversation contents.

## Coordination and handoff

- Coordinate [#127](https://github.com/huggingface/agent-manager/issues/127) for mode-switch scroll preservation, [#124](https://github.com/huggingface/agent-manager/issues/124) for app return/reconnect refresh, [#122](https://github.com/huggingface/agent-manager/issues/122) for attachments and [#120](https://github.com/huggingface/agent-manager/pull/120) for file-preview navigation. Do not duplicate their controllers or broaden their scopes. [#80](https://github.com/huggingface/agent-manager/pull/80) is older overlapping reader work; preserve the merged #119 protocol rather than restoring an older path.
- Update Reader architecture/validation docs with the new initial-history policy, search content/coverage contract, live-source behavior and actual resource bounds. Include a short demonstration of cold-opening a long conversation and finding/navigating an unloaded old message.
- Open one focused PR referencing this issue with design rationale, tests, measurements and known limitations. No merge or deployment without separate authorization. Follow `SECURITY.md` for any security-sensitive discovery; keep private data and exploit narratives out of public issue/PR content.

lvwerra:
Closes #129

## Scope

All three parts of the agreed scope: fuller initial history, whole-conversation
search, and direct navigation to unloaded old hits. The first commit did the
history half and deliberately left search; the review was right that they were
agreed as inseparable, so the second commit adds search, hit navigation, and
four review fixes. See the [review reply](https://github.com/huggingface/agent-manager/pull/144#issuecomment-5606237664)
for the fix-by-fix account.

## The defect, reproduced first

Reproduced against `main` at `ef08e843` before any change, with the real server window code and the real exchange grouping, on synthetic fixtures:

| Fixture | Exchanges the cold reader showed | Exchanges available |
| --- | --- | --- |
| OpenCode SQLite session | **1** | 40 |
| Claude JSONL, 60 KiB tool output per exchange | **3** | 60 |
| Claude JSONL, small turns | 137 | 200 |

Two distinct causes:

- **Indexed sources take `min` literally.** `windowIndex` slices exactly `min` messages off the tail, and the reader's first-paint floor is `2`. Unlike a byte window there is no span to grow — the whole conversation is parsed before `windowIndex` is reached — so asking for fewer rows saved nothing and cost the reader its history.
- **128 KiB is one tool-heavy exchange**, and nothing loaded more until the user scrolled, clicked Earlier or opened Search.

The third row is why this is conditional rather than universal, and why the fix is a fill rather than a bigger first window.

## What changed

**The first window is untouched** — same bytes, same floor, same cost to first paint. After it lands, `ReaderStore` pages backward on its own toward ~20 exchanges.

- Counted as **reconciled exchanges**, not records, tool results or lifecycle events. `countExchanges` shares `isOperatorPrompt` with `splitExchanges` (moved into `readerModel.ts` so there is one definition), and nine fixture shapes assert the two agree.
- **Bounded**: 6 backward requests, ~3 MiB, 20 s per continuous run, 50 ms between steps. Stops at the start of the source, on a blocked record, on a cursor that answers without advancing, and on any read failure. Requests and bytes are cumulative per transcript; the wall clock is per run.
- **Opt-in per consumer** (`useTraceWindows({ history })`). Child transcripts keep their 2 MiB top-of-context policy and previews page nothing. One store shared by several subscribers still fills once.
- **Warm return finishes an unfinished fill and never repeats a finished one.**
- `windowIndex` treats `min` as a floor **on a tail request only**; backward paging still honours the requested page size exactly, since that is a caller walking the conversation deliberately.

**First-viewport coverage** is a separate criterion from the exchange target, because they answer different questions. The view keeps raising the target, to a cap of 60, while the sum of *measured* row heights has not covered the scroller's content box. Estimated heights, spacers and record counts cannot satisfy it.

**No initial text drift** needed three fixes in `useVirtualRows`, each found by measuring the anchor every frame rather than only at the end:

1. A transcript shorter than the window now grows **upward from the bottom** (`flex: 1 0 auto` + `margin-top: auto`, applied only when there are exchanges so the empty state is unchanged). Below the window height there is no scroll offset to compensate with, so a top-aligned transcript was pushed down by every page that arrived. `flex` rather than `min-height: 100%` because a percentage resolves against the content box and the scroller's own padding then becomes 16 px of phantom scroll — which is itself worth 12 px of drift.
2. The reading position is re-asserted whenever the scroller's height differs from the height it was last asserted against. A prepend commits with a row slice derived from the *previous* scroll position; correcting the viewport re-renders with the right slice and a different real height, and that render changes neither `keys` nor `offsets`, so no effect keyed on those ran for it. The frame it painted was a 217–325 px jump, undone a frame later.
3. That correction runs **inside the ResizeObserver callback**, before paint, instead of waiting for React to re-render from the new heights — one frame too late, and the frame is visible.

Latest is preserved through all of it: the position is corrected before the scroll event that would otherwise report our own correction as the user scrolling away.

## Measurements

Identical fixtures, Chromium 1000×760, "before" is `main`'s source with the same tests.

| Fixture | Exchanges | First viewport filled | Max anchor drift | Backward reads | Bytes | Composer ready |
| --- | --- | --- | --- | --- | --- | --- |
| Tool-heavy, 90 KiB/exchange | 2 → **22** | 28% → **100%** | 0 → 0.69 px | 0 → 5 | 128 KiB → 2.0 MiB | 68 → 61 ms |
| Large JSONL, 20 KiB answers | 50 → **200** | 100% → 100% | 0 → 0.13 px | 0 → 1 | 128 KiB → 509 KiB | 28 → 19 ms |
| Ordinary small turns | 200 → 200 | 100% → 100% | 0 → 0.31 px | 0 → 0 | 117 KiB → 117 KiB | 26 → 20 ms |
| OpenCode SQLite, 40 exchanges | 1 → **20** (first window) | — | — | 0 → 0 | — | — |

Time to a usable composer and to first paint are unchanged. **"Before" drift is 0 because nothing arrives to displace anything**, not because the old path was stable — automatic prepending is new here, so there is no honest before-number for it. Regression budgets for the tool-heavy fixture: ≥20 exchanges, viewport covered, ≤2 px drift, ≤6 backward reads, ≤3 MiB.

## Tests

New: `web/test/readerHistory.test.mjs` (store, against a source that honours `bytes`/`min` the way `traces.js` does), `web/test/readerFirstViewport.test.mjs` (real reader in Chromium: coverage, per-frame drift, live append, hung summary, short conversations, request bounds), `server/test/trace-index-history.test.mjs` (indexed first window, exact backward paging, two conversations in one database). Extended `web/test/exchanges.test.mjs` to pin `countExchanges` against `splitExchanges`.

The issue warns that existing browser fixtures return all synthetic turns whatever was asked for. Both new fixtures serve real byte windows and grow the span until the floor is met, so a passing test cannot hide the two-message default — and the tool-heavy case asserts the first window is *short* of the target before asserting the fill reaches it.

Mutation-checked; each reverted mechanism fails a named assertion:

| Reverted | Fails with |
| --- | --- |
| Automatic fill | `only 2 exchanges became available on their own; expected at least 20` |
| Same-frame re-placement | `and Latest is still the bottom` |
| Bottom growth | `already-visible text does not move while history loads (max 454.63px…)` |
| `min-height: 100%` instead of flex | `…(max 12.09px…)` |
| Coverage request | `so it keeps going until they do (24 exchanges, …covered: false)` |
| Indexed tail floor | `a cold open sees a readable stretch of conversation, not two records (got 1 prompts in 2 messages)` |
| Leading-exchange counting rule | `answer first: counted without grouping matches grouping` |
| Fill budgets | `a budget ends the fill` |
| No-progress guard | `one backward read proves it, a second is the retry — not 6 of them` |

## Runs

- `web`: `npx tsc --noEmit`, `node ../scripts/run-suites.mjs` (**25 suites passed**), `npm run test:render` (all checks passed), `npm run build` (succeeds; retains the pre-existing large-chunk warning).
- `server`: the suite runner still stops at `test/crons.test.mjs`, so **every suite was run individually: 30 passed, `test/crons.test.mjs` failed.** That is the documented date-sensitive failure (#4, "run on restart fires once for enabled running jobs, not stopped ones"), confirmed on unchanged `main` before any edit; no cron code is touched here.
- `test/cron-api.test.mjs` failed **once** inside a 30-suite back-to-back local loop and has passed 6/6 since, including three times immediately after `crons.test.mjs`. It starts a server on a fixed port; I am reporting it rather than hiding it, but I could not reproduce it and nothing here touches cron or that port.
- `git diff --check` clean. All fixtures are synthetic and disposable; no production session was prompted or restarted, nothing was deployed, and no transcript content appears in any diagnostic.

## Whole-conversation search

Typing still filters the loaded stretch instantly. Searching everything older is
a separate explicit action (button, or ⌘/Ctrl+Enter) because it reads the
transcript: nothing scans on a keystroke or on mount.

`GET /api/trace/:id/search` scans backward from the tail in bounded pages —
256 KiB per page, ≤40 pages, ≤12 MiB, ≤4 s, ≤50 matching messages per request,
≤200-character query — continued with a cursor that is a page boundary, so
continuing can neither repeat a match nor step over one. The scan boundary for a
live conversation is the size read when the request started. Matching is literal
and case-insensitive over exactly the fields the Reader's own loaded search uses;
where a parser clipped a long block, the response says so rather than implying it
read what was cut. Empty, searching, partial, failed, cancelled and no-matches
are distinct states, and only a scan that reached the beginning says no message
contains the term.

Opening a result is **one read** of the window that result names, rendered
beside the live history rather than inside it: the live cursor never advances,
nothing is spliced, ≤30 exchanges of context ending on the match, no socket and
no input. The reader's place is captured before the window opens and restored
when it closes — verified to the pixel through 1152 px of history arriving while
the old window was up.

Reproduced first, as a test: a term in exchange 0 of a 120-exchange transcript
with 10 exchanges loaded. One click finds it; one read opens it.

## Things a reviewer should push back on if they disagree

- **Four test files switched from `bundle: false` to `bundle: true`.** Moving `isOperatorPrompt` into `readerModel.ts` gives `exchanges.ts` its first import, which an unbundled transpile cannot resolve. Bundling is strictly more faithful, but it is churn in files this PR is otherwise not about. The alternative was a second copy of the prompt-envelope rule, which would drift.
- **`ConversationView` gained a layout effect that reads `measuredHeight()` on every measurement pass.** It is a sum over loaded keys and it stops early once covered, but it is work in a hot path.
- **The post-commit height check reads `scrollHeight` on every commit**, including ones caused by typing in the composer. That is one forced reflow per commit, in line with the `origin()` reads the hook already does, but it is a real cost.
- **20 / 60 / 6 / 3 MiB / 20 s are starting points**, validated against the fixtures above and nothing else. The issue asks for them to be tuned against representative traces; real ones would be better evidence than mine.
- I removed a guard I had added to `onScroll` (distinguishing our own scroll corrections from the user's) once fixing the root cause made it redundant — the mutation showed the suite passing without it. If you would rather keep belt-and-braces there, say so.
- **Search results render above the transcript, inside the same scroller.** One scroll owner, no second pane — but on a narrow window a long result list pushes the conversation down until you clear it.
- **A hit's context is the window it was found in**, so it ends on the match with preceding context and none after it. Showing text on both sides would need a second read; I did not add one.
- **The whole-history scan is server-side and synchronous per page.** It reuses the existing window reader rather than a new index, which keeps it bounded and cache-friendly but means a very large transcript is several requests rather than one fast one. No new database or search infrastructure.
- **Bundle and child surfaces have no search entry point.** The issue said adding them was not required; the endpoint refuses a bundle explicitly rather than searching the wrong thing.

### DIFF
--- a/docs/reader-architecture.md
+++ b/docs/reader-architecture.md
@@ -38,13 +38,29 @@ available before a transcript exists and survives transcript failures.
    ancestor. Terminal and reader are mutually exclusive rendering surfaces.
 8. A redraw is not model work. The reader's activity indicator uses transcript
    evidence; transport state and terminal-derived session state stay distinct.
+9. A cold reader loads a useful amount of recent conversation on its own, in
+   bounded background steps after the first paint, counted in exchanges rather
+   than transport records. It is opt-in per consumer: children and previews keep
+   their own policy. A count is a target, not a promise — budgets, the start of
+   the conversation, or a cursor that cannot advance end the fill, and the
+   remaining history stays disclosed behind `Load earlier turns`.
+10. History arriving must not move text that is already on screen. The reading
+   position is re-asserted in the same frame as the size change that would
+   otherwise displace it, and a transcript shorter than the window grows upward
+   from the bottom rather than downward from the top.
 
 ## Verification
 
 Regression fixtures cover empty first send, hanging and failed requests, source
 switches and replacement, warm remount, cross-window tool/message/queue/final
 events, mutable SQLite records, bounded forward catch-up, child refresh, long
-history rendering/search, and outer-scroll containment. Existing trace, Markdown,
+history rendering/search, and outer-scroll containment. Initial history has its
+own two suites: `web/test/readerHistory.test.mjs` drives the store against a
+source that honours `bytes`/`min` (budgets, blocked and stalled cursors, warm
+return, release), and `web/test/readerFirstViewport.test.mjs` drives the real
+reader in a browser and measures viewport coverage and per-frame anchor
+displacement. `server/test/trace-index-history.test.mjs` covers the indexed
+first window and conversation isolation inside one database. Existing trace, Markdown,
 attachment, input-required, and terminal tests remain part of the contract.
 
 No transcript contents are emitted as diagnostics. No new runtime framework is
@@ -52,13 +68,27 @@ needed: the store uses immutable snapshots and React's external-store adapter.
 
 ## Bounds and tradeoffs
 
+- Initial history: about 20 recent exchanges, or all of a shorter conversation,
+  paged in after the first paint at one backward page per 50 ms. Bounded by six
+  backward requests, ~3 MiB (indexed rows charged at an assumed 4 KiB each) and
+  20 seconds per fill, whichever comes first. The view may raise the target, up
+  to 60 exchanges, while a page of MEASURED conversation has not yet covered the
+  scroller — twenty one-line exchanges do not fill a tall window, and one long
+  answer fills it without providing any history, so the two criteria run
+  separately. Estimated row heights and spacer divs do not count as coverage.
+  The budget is spent once per transcript: a warm remount finishes an unfinished
+  fill and does not restart a completed or exhausted one.
 - Initial read: 128 KiB / two transport messages. Ordinary pages: 384 KiB;
   the server can extend a page to finish a record, up to its existing 8 MiB
   ceiling. Forward catch-up preserves every intervening page, including backlogs
   larger than that ceiling. Oversized records produce an explicit recovery state.
   Child readers use a bounded 2 MiB initial window, open at its top, and disclose
   when earlier context (including the task) still needs paging. Latest opts into
-  following the child tail.
+  following the child tail. They do not take part in automatic history fill.
+  An INDEXED source treats the first-paint floor as a floor rather than an exact
+  count: it has no span to grow and has already parsed the whole conversation, so
+  returning two rows of it saved nothing and showed one exchange. Backward paging
+  from an indexed source still honours the requested page size exactly.
 - Window and summary requests time out after 12 seconds. Failed reads back off
   to 30 seconds. Visible readers poll every 3–10 seconds; SQLite uses 10 seconds.
   Hidden/inactive readers cancel outstanding requests and do not poll. A browser
--- a/docs/reader-architecture.md
+++ b/docs/reader-architecture.md
@@ -47,7 +47,21 @@ available before a transcript exists and survives transcript failures.
 10. History arriving must not move text that is already on screen. The reading
    position is re-asserted in the same frame as the size change that would
    otherwise displace it, and a transcript shorter than the window grows upward
-   from the bottom rather than downward from the top.
+   from the bottom rather than downward from the top. The first transcript a
+   cold reader PRESENTS is held back — laid out and measured, not painted —
+   until it covers the reader or until one of four bounds ends the wait.
+11. Search has two scopes. Typing filters the loaded stretch, instantly and
+   locally, as it always has. Searching the whole conversation is a separate,
+   explicit action, because it reads the transcript: nothing scans on a
+   keystroke or on mount. Both search the same content, so they cannot disagree
+   about what counts as conversation.
+12. Opening an old result is a read, not a move. It fetches one bounded window
+   named by the result itself and shows it beside the live history rather than
+   inside it: the live cursor does not advance, nothing is spliced, the reader's
+   place is remembered before the window opens and restored when it closes, and
+   nothing about it starts, claims, resizes or types into a terminal.
+13. Speculative work yields. A forward read — Latest, refresh, the live poll —
+   never waits behind an automatic backward page.
 
 ## Verification
 
@@ -69,9 +83,31 @@ needed: the store uses immutable snapshots and React's external-store adapter.
 ## Bounds and tradeoffs
 
 - Initial history: about 20 recent exchanges, or all of a shorter conversation,
-  paged in after the first paint at one backward page per 50 ms. Bounded by six
-  backward requests, ~3 MiB (indexed rows charged at an assumed 4 KiB each) and
-  20 seconds per fill, whichever comes first. The view may raise the target, up
+  paged in after the first response at one backward page per 50 ms. Bounded by
+  six backward requests, ~3 MiB RETAINED (indexed rows charged at an assumed
+  4 KiB each) and 20 seconds per continuous run, whichever comes first. Retained
+  rather than read: a page's size is not knowable before fetching it, so the
+  request that crosses the budget is the last one. What bounds a single read is
+  the message floor a speculative page asks for — one, the smallest that still
+  guarantees progress — which stops the server growing a response beyond a
+  single oversized record. Left to the server's own default of twelve, one page
+  of 900 KiB answers grew to six megabytes.
+- First paint holds until the transcript is worth showing, for at most two
+  seconds, and ends earlier on coverage, the start of the conversation, an
+  unusable source or an exhausted fill. The composer is never part of the wait
+  and a warm store is never hidden.
+- Whole-conversation search: 256 KiB scan pages, at most 40 pages, 12 MiB or
+  four seconds per request, 50 matching messages per page, and a 200-character
+  query. Longer conversations are more bounded requests, continued with a cursor
+  that is a page boundary, so continuing can neither repeat nor skip a match. The
+  scan runs newest-first from the size read when the request started, so output
+  appended while it runs is outside that request rather than something it chases.
+  Results carry the window that opens them; opening one is a single read.
+  Matching is literal and case-insensitive over the text the reader displays —
+  prompts, assistant text, thinking, tool names, tool input and output, shell
+  commands and their streams — and never image data or transport metadata. Where
+  a parser clipped a long block, the scan says so instead of implying it read
+  what was cut. The view may raise the target, up
   to 60 exchanges, while a page of MEASURED conversation has not yet covered the
   scroller — twenty one-line exchanges do not fill a tall window, and one long
   answer fills it without providing any history, so the two criteria run
@@ -114,10 +150,14 @@ needed: the store uses immutable snapshots and React's external-store adapter.
   undetectable from stats alone. An unobserved
   truncate-and-regrow with the same header/inode can still require a full browser
   reload; this is not a content-addressed log protocol.
-- Search means loaded content, not a server-side search of the entire trace. It
-  navigates matching turns, not every individual highlighted word. Virtualization
-  also means native browser Find, selection and accessibility traversal see the
-  mounted stretch; use reader search or the raw download for full-history work.
+- Search has the two scopes above. The instant filter still means loaded content,
+  and still says so. The explicit scope reads the whole transcript on the server
+  and returns places to jump to, not a second transcript: it navigates matching
+  messages, not every highlighted word, and a result opens roughly thirty
+  exchanges of context ending on the match. Bundle and child surfaces keep their
+  existing behaviour; no new search entry point was added to them. Virtualization
+  still means native browser Find, selection and accessibility traversal see the
+  mounted stretch.
 - Working is based on Codex task events and Claude message/stop-reason evidence,
   gated by the session process still being alive. Unknown activity is not promoted
   to working from a terminal repaint. Other formats can add explicit activity
--- a/docs/reader-validation.md
+++ b/docs/reader-validation.md
@@ -29,12 +29,56 @@ Validated on 2026-09-04, based on main `705e32dc`.
 | Operator HTML/XML disappears as harness text | Named harness envelopes in parser and exchange grouping instead of every leading `<` | `<div>Please review this markup</div>` remains a user prompt and opens a prompt band |
 | Existing UI contracts | Retained shared composer, attachment flow and info panel | Full-app reader-info, attachment-input and terminal UI integration suites pass |
 
+## Initial history (issue #129)
+
+Added 2026-09-09, on main `ef08e843`.
+
+| Problem / contract | Implementation | Evidence |
+| --- | --- | --- |
+| A cold reader shows one or two exchanges | Bounded backward fill to ~20 exchanges after the first paint, counted in exchanges; indexed sources get a first-window floor | Tool-heavy JSONL fixture goes from 2 exchanges to 22 with no scrolling, search or summary; a 40-exchange OpenCode database goes from 1 to 20 in its first window |
+| Counting records instead of conversation | `countExchanges` shares `isOperatorPrompt` with the grouping | Nine fixture shapes assert `countExchanges === splitExchanges().length`, including a transcript starting mid-exchange and harness envelopes that are not prompts |
+| A fill that cannot finish keeps trying | Request/byte/time budgets, plus start-of-source, blocked and no-progress stops | 900 KiB answers stop on the byte budget at ≤6 backward reads; a blocked cursor and a cursor that answers without advancing each stop after one retry, and neither claims the beginning of the conversation |
+| Preload competes with the reader | One request slot, 50 ms between steps, opt-in per consumer, cancelled on release | Composer and first text stay available under a 60 s hung summary; releasing a reader mid-fill stops it within one step; no child reader is created by a parent preload |
+| A warm return re-runs the preload | Budget and target are per transcript, not per mount | Remount after a completed fill issues zero further backward reads and keeps its history |
+| The first page does not fill the reader | The view raises the target while MEASURED rows have not covered the scroller's content box | 20 one-line exchanges leave a 2600 px reader uncovered; it continues to 94 and covers it, in one extra backward read |
+| Covering the page is mistaken for having history | Coverage and the exchange target are separate criteria | One 140-line answer covers the reader at 2 exchanges; the fill still reaches 22 |
+| Older history pushes visible text down | Position re-asserted in the same frame as the size change; transcript shorter than the window grows from the bottom | 0.69 px maximum anchor displacement over 110 frames while filling; 0.51 px across the underfilled→scrollable transition over 72 frames; a genuine live append still moves the view 77 px, which is Latest working |
+| Latest quietly stops following mid-preload | Corrections land before the scroll event that reports them | Latest is claimed on every one of 110 frames during the fill |
+
+Removing each mechanism fails a named assertion: no fill → “only 2 exchanges
+became available on their own”; no same-frame re-placement → “Latest is still
+the bottom”; no bottom-growth → 454 px of displacement; no coverage request →
+uncovered 2600 px reader; no indexed floor → “1 prompts in 2 messages”; a
+counter that ignores a leading exchange → the grouping comparison.
+
+### Measurements
+
+Same fixtures, before and after, Chromium at 1000×760. “Before” is main's source
+with the same tests.
+
+| Fixture | Exchanges after settling | First viewport filled | Max anchor drift | Backward reads | Bytes | Composer ready |
+| --- | --- | --- | --- | --- | --- | --- |
+| Tool-heavy, 90 KiB/exchange | 2 → **22** | 28% → **100%** | 0 → 0.69 px | 0 → 5 | 128 KiB → 2.0 MiB | 68 → 61 ms |
+| Large JSONL, 20 KiB answers | 50 → **200** | 100% → 100% | 0 → 0.13 px | 0 → 1 | 128 KiB → 509 KiB | 28 → 19 ms |
+| Ordinary small turns | 200 → 200 | 100% → 100% | 0 → 0.31 px | 0 → 0 | 117 KiB → 117 KiB | 26 → 20 ms |
+| OpenCode database, 40 exchanges | 1 → **20** (first window) | — | — | 0 → 0 | — | — |
+
+Time to a usable composer and to the first transcript paint are unchanged: the
+first window is the same request it always was. “Before” drift is zero because
+nothing arrives to displace anything, not because the old path was stable — the
+displacement this fixes is a property of prepending, which nothing did
+automatically before. Regression budgets for the tool-heavy fixture: ≥20
+exchanges, viewport covered, ≤2 px drift, ≤6 backward reads, ≤3 MiB.
+
 ## Runs
 
 - Web: `npm run build`, `npm test` (23 suites), `npm run test:render`.
+  For #129, re-run: `npm test` (25 suites), `npm run test:render`, `npm run build`.
 - Server: all 27 non-cron default suites passed. This includes trace windows,
   queued prompts, child traces, protocol regressions, attachments, permissions,
   migration, resize and state checkpoint tests.
+  For #129, re-run: every suite individually (the runner still stops at the cron
+  failure below); 30 passed, `test/crons.test.mjs` failed, unchanged from main.
 - Full-app browser integration: `reader-info.test.mjs`,
   `screenshot-input.test.mjs`, `terminal-ui.test.mjs`, using the production build.
 - Final targeted rerun: reader model/browser/protocol, legacy trace windows and
--- a/docs/reader-validation.md
+++ b/docs/reader-validation.md
@@ -45,11 +45,33 @@ Added 2026-09-09, on main `ef08e843`.
 | Older history pushes visible text down | Position re-asserted in the same frame as the size change; transcript shorter than the window grows from the bottom | 0.69 px maximum anchor displacement over 110 frames while filling; 0.51 px across the underfilled→scrollable transition over 72 frames; a genuine live append still moves the view 77 px, which is Latest working |
 | Latest quietly stops following mid-preload | Corrections land before the scroll event that reports them | Latest is claimed on every one of 110 frames during the fill |
 
+### Review round (2026-09-09)
+
+| Finding | Fix | Evidence |
+| --- | --- | --- |
+| The whole-conversation half of #129 was absent | Server scan endpoint, an explicit Reader action, and an old-hit window separate from the live store | A term in exchange 0 of a 120-exchange transcript, with 10 exchanges loaded: one click finds it, one read opens it, and the reader's place is restored on the way back |
+| The first PRESENTED viewport was still the first response | Rows are laid out and measured but not painted until they cover the reader, bounded four ways | The frame the transcript first becomes visible on is asserted covered, on a fixture whose first window is two exchanges and whose older pages are slow |
+| A forward read waited behind speculative history | A forward read cancels an in-flight speculative page instead of inheriting its promise; the abandoned step is uncharged and restartable | With a 300 ms backward page in flight, `loadNewer()` issues an `after` request rather than resolving with the backward one |
+| The documented byte limit was not a limit | A speculative page asks for a floor of one message, so the server cannot grow it past one record; the step is charged before the next is scheduled | On 900 KiB answers: largest page 1.5 MiB (was 6), retained 4.4 MiB against a 3 MiB budget plus the page that crossed it, in four requests |
+
+Two further defects surfaced while testing the fix, both fixed here: the virtual
+list kept correcting the scroller while its own rows were hidden behind an old
+window (every box measures zero, so it scrolled the borrower to the top), and
+scrolling inside a borrowed window could make the live transcript page backward.
+
 Removing each mechanism fails a named assertion: no fill → “only 2 exchanges
 became available on their own”; no same-frame re-placement → “Latest is still
 the bottom”; no bottom-growth → 454 px of displacement; no coverage request →
 uncovered 2600 px reader; no indexed floor → “1 prompts in 2 messages”; a
-counter that ignores a leading exchange → the grouping comparison.
+counter that ignores a leading exchange → the grouping comparison; no
+whole-history scan → the results never appear; a hit opened as a tail → “opened
+with the locator the search handed back”; no run token → “the newest query owns
+the results”; always claiming a complete scan → the partial-coverage line never
+appears; no clipped disclosure → its line never appears; a server scan limited
+to one page → “found the needle far outside the tail (0 hits)”; a database scan
+that ignores the conversation id → the neighbouring conversation leaks in; the
+virtual list not standing down while hidden → “the old window opens on its
+match”; no remembered position → the row returns 704px away instead of 0.
 
 ### Measurements
 
@@ -73,7 +95,7 @@ exchanges, viewport covered, ≤2 px drift, ≤6 backward reads, ≤3 MiB.
 ## Runs
 
 - Web: `npm run build`, `npm test` (23 suites), `npm run test:render`.
-  For #129, re-run: `npm test` (25 suites), `npm run test:render`, `npm run build`.
+  For #129, re-run: `npm test` (26 suites), `npm run test:render`, `npm run build`.
 - Server: all 27 non-cron default suites passed. This includes trace windows,
   queued prompts, child traces, protocol regressions, attachments, permissions,
   migration, resize and state checkpoint tests.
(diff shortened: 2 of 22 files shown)


lvwerra (commented):
1. **Blocking — the explicitly inseparable search half of #129 is absent.** The agreed scope says recent-history loading, whole-conversation search, and direct navigation to unloaded old hits must ship together in one PR. This diff leaves `ConversationView` filtering only the already-loaded `exchanges` (`web/src/components/conversation/ConversationView.tsx:75-76,145-170`) and adds neither a whole-conversation scan nor a separate old-hit window. To trigger it, put a unique term before the retained tail in a long transcript and search for it: the Reader still reports that earlier history was not searched, and the user must manually page backward. The PR body acknowledges requirements 2 and 3 were deliberately omitted while still using `Closes #129`; that would close an issue whose main user path remains unavailable.

2. **Blocking — the first *presented* transcript viewport is still underfilled.** The initial 128 KiB/two-turn response is rendered immediately, and viewport coverage is requested only afterward from a layout effect (`ConversationView.tsx:207-212`). The new browser test itself observes this: it waits until `Question 199` is visible, records only two exchanges (`readerFirstViewport.test.mjs:214-220`), and in the delayed tool-heavy case explicitly confirms that this first visible window is shorter than the Reader (`:251-257`). The later bottom-anchored prepends avoid moving that text, but they do not satisfy the requirement that the first presented transcript cover one real Reader viewport when enough history exists. On a long tool-heavy trace with a delayed `before` response, the user first sees two messages and empty space, then watches older content fill upward; the issue allowed a brief stable preparation surface specifically to avoid exposing this partial first transcript paint.

3. **High — a forward refresh is swallowed whenever speculative history has the single request slot.** `loadNewer()` calls `read('after')`, but `read()` returns any existing request's promise without regard to direction (`readerStore.ts:197,282-286`). With a slow automatic `before` request in flight, clicking Latest or a live poll therefore waits for that backward request and never issues an `after` request. I reproduced this by delaying a fixture's `before` response by 300 ms, waiting for `loading === 'before'`, and calling `loadNewer()`; the promise completed with no `after` call. New output remains missing until a later poll, which each completed backward page reschedules for another 3–10 seconds. This is the concrete starvation case the issue requires speculative fill not to cause.

4. **High — the documented 3 MiB automatic-fill byte limit is not a limit.** Backward reads send 384 KiB with `min` omitted (`readerStore.ts:292-294`), so the byte source applies its default floor of 12 messages and grows one response as far as 8 MiB (`server/src/traces.js:1902-1904,1942-1961`). The client only charges bytes after that response (`readerStore.ts:260-268`), and `read()`'s `finally` schedules the next fill step before that downstream accounting runs (`:344-352`). Instrumenting the PR's existing 900 KiB-answer budget fixture produced a 1 MiB tail followed by two 6 MiB backward ranges (about 10.5 MiB of retained cursor advance) while the test still passed and reported `limited`. Thus one request already exceeds 3 MiB and the ordering permits an extra over-budget request. `readerHistory.test.mjs:171-184` checks only the six-request cap, so the claimed byte regression budget is not tested.

Verification: the full 25-suite web runner and the focused indexed-history server test pass on `192e791`. Replacing `wantHistory()` with a no-op makes `readerHistory.test.mjs` fail at the tool-heavy target assertion, so the basic automatic-fill test is mutation-sensitive. The two targeted probes above expose the untested forward-priority and byte-limit failures.

lvwerra:
All four findings were real and all four are fixed in `e2e1228`. Nothing pushed back on. Two more defects surfaced while testing the fixes and are fixed here too.

## 1. The search half — implemented

No argument: it is in. Reproduced the reviewer's trigger first, as a test — a term in exchange **0** of a 120-exchange transcript with **10** exchanges loaded — and the premise is asserted before anything else, so a test that could be satisfied by filtering loaded turns fails immediately.

**Server** — `GET /api/trace/:id/search`. Scans backward from the tail in bounded pages: 256 KiB per page, ≤40 pages, ≤12 MiB, ≤4 s, ≤50 matching messages per request, ≤200-character query. Longer conversations are more bounded requests, continued with a cursor that is a **page boundary**, so continuing can neither repeat a match nor step over one. The scan boundary for a live conversation is the size read when the request started, so appended output is outside that request rather than something it chases. Matching is literal and case-insensitive over exactly the fields the Reader's own loaded search uses — prompts, assistant text, thinking, tool names, tool input and output, shell commands and their streams — never image data or transport metadata. Where a parser clipped a long block, the response says so instead of implying it read what was cut.

**Client** — typing still filters the loaded stretch instantly and unchanged. Whole-conversation search is a separate action (a button, or ⌘/Ctrl+Enter) because it reads the transcript: nothing scans on a keystroke or on mount, asserted. Enter still walks loaded matches, so an edited query cannot start a scan by accident or start two at once. Empty / searching / partial / failed / cancelled / no-matches are distinct states, and only a scan that reached the beginning says "No message in this conversation contains …".

**Opening an old hit** is one read of the window the result names — not a walk. It renders beside the live history, not inside it: the live cursor never advances, nothing is spliced, ≤30 exchanges of context ending on the match, and it starts no socket and sends no input (asserted). The reader's place is captured before the window opens and restored when it closes — verified to the pixel through **1152 px of history arriving while the old window was up**: the row returns to 67 px, exactly where it was.

`Closes #129` now stands without the caveat that was attached to it.

## 2. First presented viewport — fixed

You were right that coverage was requested only after the first response was painted. Rows are now **laid out and measured but not painted** until they cover the reader, bounded four ways: coverage, the start of the conversation, an unusable source (error/blocked/exhausted fill), and a 2 s deadline. The composer is never part of the wait, and a warm store is never hidden.

The test asserts on the **frame the transcript first becomes visible**, not on the settled state — on a fixture whose first window is two exchanges and whose older pages are deliberately slow. A short conversation is still shown at once (asserted separately).

The `readerFirstViewport` case you cited as evidence against the old behaviour has been repointed: its old precondition is now unreachable by construction, so it tests what still needs pinning — that a transcript shorter than the window sits against the composer.

## 3. Forward reads swallowed — fixed

Reproduced exactly as described, as a failing test first. A forward read now cancels an in-flight **speculative** page instead of inheriting its promise; a user's own backward page is never cancelled by a background poll. The abandoned step is uncharged and un-stalled — it simply did not happen — and the fill restarts it once the forward read settles.

## 4. The byte limit — fixed, and the constant renamed

Two separate defects, both real:

- Backward reads sent no `min`, so the server's own floor of twelve took over and grew one page to 6 MiB. Speculative pages now ask for a floor of **one** — the smallest that still guarantees progress — so the server cannot grow a response past a single oversized record. On your 900 KiB-answer fixture: **largest page 1.5 MiB, down from 6**.
- Charging happened in a `.then()` chained downstream of `read()`, so the budget still read zero when the next page was scheduled. It is now charged synchronously at the top of `planFill`, before anything decides.

I also renamed the constant to `FILL_MAX_RETAINED_BYTES`, because you were right that the old name promised something it could not deliver: a page's size is not knowable before fetching it, so the request that crosses the budget is the last one, and a server page reads slightly more than it returns. The three quantities are now bounded and asserted separately — largest page ≤2 MiB, retained ≤ budget + the page that crossed it (4.4 MiB observed), total reads ≤ requests × page ceiling — instead of one number that was none of them.

## Two more, found while testing the above

- **The virtual list kept correcting the scroller while its own rows were hidden** behind an old window. Every box in a `display:none` subtree measures zero, so "holding the position" scrolled the borrower to the top and lost the place to come back to. It now stands down when its list is not on screen.
- **Scrolling inside a borrowed window could make the live transcript page backward** — the live view's "near the top, load earlier" rule firing on someone else's scroll position.

## Tests

New: `server/test/trace-search.test.mjs`, `web/test/readerSearch.test.mjs`. Extended: `readerHistory` (forward priority, the three byte bounds), `readerFirstViewport` (first visible frame).

Every fix is mutation-checked — reverting it fails a named assertion:

| Reverted | Fails with |
| --- | --- |
| Whole-history scan | results never appear |
| Server scan limited to one page | `found the needle far outside the tail (0 hits)` |
| Database scan ignoring the conversation id | the neighbouring conversation leaks in |
| Hit opened as a tail | `a hit is opened with the locator the search handed back` |
| Query/run binding | `the newest query owns the results` |
| Always claiming a complete scan | the partial-coverage line never appears |
| Clipped disclosure | its line never appears |
| First-page preparation | `the first transcript the reader shows covers it (…covered:false, ex:2)` |
| Forward priority | `asking for the latest while history is loading actually asks the source for it` |
| Speculative page floor | `the server cannot grow one speculative page to megabytes (largest 6.0 MiB)` |
| Charging before scheduling | `retained history stays inside the budget plus the page that crossed it (5.3 MiB…)` |
| Virtual list standing down while hidden | `the old window opens on its match` |
| Remembered position | the row returns 704 px away instead of 0 |
| Borrowed-scroll isolation | `scrolling inside an old window does not page the live transcript` |

Three of those assertions had to be rewritten before they bit — my first versions were satisfied by the mutation, which makes them worse than nothing. In particular the stale-query race is only reachable through the keyboard shortcut (the button disables itself during a scan) and only observable when the abandoned request is the **slower** one and the transport answers it despite the abort; the test now sets all three up deliberately.

## One existing fixture changed

`readerRedesign`'s `paging-priority` mount opened with fewer exchanges than the reader now loads by itself, so it reached the beginning of the conversation before the Earlier click its assertion is about, and the button was legitimately gone. Its `from` moved so the fixture has more history than the automatic target — the assertion it makes is unchanged, and the comment says why the number matters.

## Runs

- `web`: `npx tsc --noEmit`, `node ../scripts/run-suites.mjs` — **26 suites passed**; `npm run test:render` — all checks passed; `npm run build` — succeeds (retains the pre-existing large-chunk warning).
- `server`: the runner still stops at `test/crons.test.mjs`, so every suite was run individually — **31 passed, `test/crons.test.mjs` failed**, the documented date-sensitive #4, confirmed red on unchanged main and untouched here.
- `readerSearch` run 3× back to back for flakiness after one assertion proved timing-sensitive; stable.
- Docs updated: architecture gains the search contract, the preparation phase and corrected bounds; validation gains this review round with its before/after numbers.

PR #145, opened 2026-09-09
Standardize API input validation and error handling

### LINKED ISSUE #134: Standardize API input validation and error handling
## Goal and agreed scope

Make HTTP API failures predictable for the browser and supported agent/script clients: validate inputs before acting, reliably catch asynchronous request failures, and preserve a safe, useful error message and machine-readable reason through to the caller.

The operator approved this as one focused PR. Prioritize the existing Express 4 request/error boundary and frontend response decoding, not a framework migration or a rewrite of domain services. Safe trace re-imports and stable scheduled-target identities were explicitly deferred; their transaction/targeting behavior is not part of this issue.

## Starting points

Planning inspected main at `ef08e843d1c67fcc4e1c463415819ea351734d6d`; recheck before implementing. The server declares Express 4. `web/src/api.ts` has a shared helper that reduces failures to their HTTP status, another that preserves the server message, and specialized handling for uploads, trace availability and sharing. Preserve those useful domain distinctions while consolidating the transport/error contract.

Express 4's router invokes handlers inside a synchronous try/catch but does not itself consume their returned promises; use an explicit request-scoped rejection-forwarding pattern on the current stack. Do not assume the automatic promise behavior described for Express 5 applies here. See the [Express 4.19.2 router source](https://github.com/expressjs/express/blob/4.19.2/lib/router/layer.js). A global unhandled-rejection logger is not a substitute for answering the affected request.

This is a desired correctness/compatibility specification, not a public vulnerability reproduction. Follow SECURITY.md for any sensitive findings encountered during implementation; do not publish operator payloads, credentials or private review evidence.

## Requirements

| Area | Required behavior |
| --- | --- |
| Inventory the boundary | Inventory HTTP handlers/middleware, existing validation and browser/native consumers before migration. Identify JSON, text/plain, raw upload, download, stream/long-poll and deliberately empty responses. Record existing domain statuses, codes, supplemental fields and meaningful success responses containing `ok: false`; do not classify every such result as a transport failure. Keep streaming/WebSocket protocol handling explicit rather than forcing every endpoint through a JSON wrapper. |
| Reliable async handling | Add a small shared async-handler/error boundary and cover applicable existing handlers and async middleware, including synchronous throws and rejected promises. Preserve local catches that intentionally isolate optional subsystems. Ensure each request receives one appropriate response or controlled stream termination, never hangs merely because a promise rejected, and never calls the error path twice. Event-emitter/callback failures need explicit forwarding/cleanup; a promise wrapper alone does not catch them. |
| Compatible error envelope | Retain a top-level `error` string for existing clients, with a stable machine-readable `code` and narrowly defined optional safe fields such as validation details, conflict/revision data or retry guidance. Document the shape/status mapping and preserve established fields such as trace reason, lock reason, redaction hits and conflict tags during migration. Prefer an additive contract; do not replace the error string with an object or require all copied agent scripts to update at once. |
| Safe, truthful errors | Expected domain failures get actionable messages and appropriate status codes. Unexpected internal errors get a generic safe message/code, not a stack trace, credential, file contents, raw upstream response or arbitrary exception text. Keep diagnostic correlation metadata bounded and safe. Do not downgrade a failed request to HTTP 200 or convert a legitimate successful empty response into a JSON parse failure. A completed partial operation must retain its documented partial-result semantics rather than falsely implying nothing happened. |
| Server-side input validation | Validate types, required fields, enums, numeric bounds and semantic constraints at the request boundary before domain side effects. Cover mutation payloads and applicable path/query parameters using small reusable validators or schemas; TypeScript annotations alone are not validation. Inventory supported forms and aliases, including text prompts and uploads. Distinguish omitted/null/empty/false values; avoid blanket coercions that change meaning. Define unknown-field behavior per existing endpoint compatibility rather than silently dropping meaningful fields or globally rejecting formerly supported extension data. |
| Preserve mutation semantics | Document whether an endpoint replaces an object, patches selected fields, executes a command or starts asynchronous work. Keep existing valid client behavior and supported PUT/PATCH routes; this is not permission to rename endpoints or change partial-update semantics wholesale. Coordinate settings validation with #123. Persisted-store schemas, durable commit guarantees, corrupt-state recovery and import transaction changes remain outside scope; if one is necessary, identify the separate blocker rather than expanding this PR. |
| Browser decoding | Consolidate JSON/status/error decoding into one tested transport contract with a typed error carrying HTTP status, stable code and safe optional domain data. Preserve compatibility with legacy `{ error: string }` responses. Handle empty, malformed and non-JSON failure responses from proxies or interrupted connections without hiding the status behind a JSON syntax error or rendering arbitrary HTML. Update applicable call sites so expected validation/conflict errors are useful rather than bare numbers. Keep uploads' progress/cancel behavior and their XMLHttpRequest path. |
| Preserve domain and cancellation behavior | Keep equivalent behavior for TraceUnavailable/no-trace, sharing redaction refusal, file revision conflicts, rate limits and privacy locks. A new session with no trace must still show its usable Reader composer. Distinguish deliberate abort/cancellation from an HTTP rejection, timeout, offline/network failure or unreadable successful response. Do not replace existing Reader deadlines or remote long-poll timing with a blanket short timeout. Error metadata is not authority to retry: do not automatically replay state-changing requests or imply uncertain delivery did not happen. |
| Streams and committed responses | Before headers are sent, return the documented structured failure where applicable. After streaming/headers have begun, do not append JSON into a file/terminal stream, send another status or trigger headers-already-sent errors. Close/cancel or use the protocol's existing failure mechanism, releasing timers/listeners/resources. Client disconnect and late async completion must not create a second response, retry an operation or affect unrelated sessions. Do not redesign WebSocket framing or remote protocols. |
| Routing, guards and diagnostics | Add deliberate API not-found/body-parser error handling so ordinary API mistakes do not fall through to an HTML app page. Preserve static asset and SPA routing. Keep origin/privacy/attribution guards and their ordering/semantics; coordinate #130/#131 rather than bypassing them to simplify error handling. Preserve one accurate audit outcome and the filtering requirements of #132; do not broaden rejected-body capture or add secret-bearing diagnostics. A request wrapper must not introduce a process restart/agent revival policy. |

## Required tests

Use real local HTTP/middleware integration plus focused decoder/validator and browser tests with isolated state, fake agents/dispatchers and mocked external services. Test actual responses and side effects, not only construction of an Error object.

| --- | --- |
| Async failures | Inject a synchronous throw, immediate and delayed rejection, non-Error rejection, callback/stream error and late failure after a response. Requests settle exactly once with the documented outcome, without escaping to an unhandled rejection, hanging or sending a second response. A subsequent unrelated request still succeeds. |
| Validation before action | Missing/wrong-type/null/empty values, arrays where scalar values are expected, duplicate/structured query values, invalid enum/numeric bounds, malformed JSON and unsupported content types under the documented policy. Assert zero domain calls, file writes, notifications or agent input for rejected inputs. Include valid boundary values, supported aliases/text payloads and existing replacement/patch semantics as positive cases. |
| Contract and client compatibility | Existing error string plus added code/details; old-style server errors consumed by the new browser; supported native/internal/remote clients; 400/403/404/409/413/429/5xx as applicable. Preserve existing domain fields and success payloads, including deliberately empty success and meaningful `ok: false` results. |
| Domain UI | Useful session/group validation messages, file conflict details, rate-limit feedback, public-share redaction refusal and trace-unavailable/no-conversation states. Test the actual affected controls, retaining user input/drafts and existing failure actions. No empty-state substitution for a failed fetch, blanket popup system or per-poll toast spam. |
| Network and parsing | Non-JSON HTML/text failure, empty failure body, malformed JSON failure, malformed expected-JSON success, abort, timeout and connection loss. Messages remain safe and useful, the HTTP status is retained where present, and no automatic mutation replay occurs. Error responses should be read once, with bounded diagnostic content. |
| Uploads and streams | Real XMLHttpRequest success/progress/cancel/error decoding; file/download and remote long-poll failures before/after headers; client disconnect and late callbacks. No JSON appended to stream bytes, duplicate work, leaked listeners/timers or changed legitimate long-poll behavior. |
| Routing and protection | API not-found and parser errors versus ordinary SPA/static requests; privacy/origin/attribution denial retains its code/reason and zero forbidden effects. Test with guards enabled and safe synthetic credentials. Audit recording reflects the actual final response once, without raw sensitive failure content. |
| Error safety and regressions | Inject credential-like strings, private paths, payload text and markup into internal/upstream exceptions. They must not leak through public error bodies or unsafe diagnostics. Verify existing normal API fixtures, Reader/terminal separation, save controllers and unaffected optional-subsystem isolation. |

- Add normally discovered tests. Run focused API/validator/decoder suites, relevant files/settings/attachments/trace/remote/operations integrations and browser cases, normal server/web suites, and frontend typecheck/production build. Report exact commands and unrelated baseline failures separately; still run the new focused suites if an aggregate runner stops early. Do not implement the deferred CI project to satisfy this requirement.
- Include a route/response compatibility inventory, error-code/status table, documented validator conventions and concise examples of adding a new handler/client. Keep helpers small; no backend TypeScript conversion, schema-generation platform or giant route extraction is required.

## Integration and exclusions

Start with `server/src/index.js`, applicable route-level validators, `web/src/api.ts`, shared TypeScript API types and error-consuming components/tests. Preserve the approved boundaries in [#121](https://github.com/huggingface/agent-manager/issues/121), [#122](https://github.com/huggingface/agent-manager/issues/122), [#123](https://github.com/huggingface/agent-manager/issues/123), [#124](https://github.com/huggingface/agent-manager/issues/124), [#129](https://github.com/huggingface/agent-manager/issues/129), [#130](https://github.com/huggingface/agent-manager/issues/130), [#131](https://github.com/huggingface/agent-manager/issues/131) and [#132](https://github.com/huggingface/agent-manager/issues/132). Share applicable helpers; do not duplicate or take over their feature work.

No safe-import replacement implementation, scheduled-target identity changes, core-store durability/recovery rewrite, automatic retries/delivery receipts, process-supervision/restart policy, new authentication, general dialog/UI redesign, Express major-version migration, mass dependency upgrades, build/CI changes or deployment. Any incompatible client/API behavior needs a separately agreed migration, not an undocumented breaking change. No real agent prompts/stops, notifications, live credential/log reads, hosting changes or paid jobs in tests. Open one focused PR; merging and deployment require separate authorization.

lvwerra:
Closes #134

## Implemented

- A shared Express 4 route wrapper catches synchronous and asynchronous request failures. API not-found/parser errors use JSON; committed streams terminate without a second response. Remote poll and file-upload callbacks have explicit cleanup.
- Runtime validation before route actions: JSON types, required fields, enums, queries, target references, settings, text prompts and raw uploads. Existing replacement/patch semantics and native text/query aliases remain supported; existing history-size clamps remain compatible.
- Additive `{ error: string, code }` failures, safe unexpected errors, preserved domain fields, and truthful incomplete audit outcomes. Request-body logging/filtering is unchanged.
- One typed browser decoder for fetch and XHR, retaining HTTP status and safe domain data. Legacy error strings, empty successes, trace availability, sharing refusals, file conflicts, rate limits, cancellation and upload progress remain supported. Existing controls show useful errors and retain drafts.
- [Route/response inventory, status table, validator conventions and examples](https://github.com/huggingface/agent-manager/blob/fix/api-errors-validation-134/docs/api-errors.md).

## Deliberately left alone

No Express/framework migration, domain-service rewrite, safe trace re-import transaction work, scheduled-target identity changes, store durability/recovery, automatic mutation retries, delivery receipts, process supervision, authentication, WebSocket framing, deployment or general UI redesign. Optional-subsystem fallbacks and meaningful successful partial/capability results remain intact. The file overwrite policy and settings durability work remain with their separate issues; origin protection and audit content filtering are not reimplemented here.

The audit's additive `incomplete` flag describes an unfinished HTTP response, not a rolled-back operation or permission to retry. The only direct persistence changes here stop request handlers from acknowledging failed writes; there is no new persistence protocol.

## Verification

All tests used isolated local state, fake agents and mocked external services. No live agent work, notifications or deployments.

- `cd server && npm test`: new tests pass; stops at the already-known main failure in `test/crons.test.mjs` #4. The 21 suites after that point were also run individually and pass (29 passing discovered server suites overall, excluding the known-red suite).
- `cd server && node ../scripts/run-suites.mjs api-boundary api-http attachments backup cron-api operations reader-protocol trace`: 12 suites pass.
- `cd server && node ../scripts/run-suites.mjs terminal-ui screenshot-input reader-info --manual`: all 3 desktop integration suites pass.
- `cd server && node ../scripts/run-suites.mjs api-boundary api-http operations`: all 3 pass after the final audit change.
- `cd web && npm test`: all 25 discovered suites pass, including real control/fetch/XHR tests.
- `cd web && npm run build`: typecheck and production build pass.
- Ten deliberate mutations are caught: promise forwarding, validation, safe failure bodies, committed-stream handling, poll cleanup, HTTP status retention, trace error classification, bounded failure reads, the production attachment route wrapper, and incomplete audit outcomes. Mutations were made only in a separate disposable checkout and restored.

<details>
<summary>Individual server commands run after the aggregate stopped</summary>

```sh
node test/first-run.test.mjs
node test/hidden.test.mjs
node test/input-required.test.mjs
node test/next-name.test.mjs
node test/opencode-resume.test.mjs
node test/operations.test.mjs
node test/queued-prompts.test.mjs
node test/reader-protocol.test.mjs
node test/repin.test.mjs
node test/revive.test.mjs
node test/slowfs.test.mjs
node test/spawn-group.test.mjs
node test/subagent-traces.test.mjs
node test/terminal-modes.test.mjs
node test/trace-download.test.mjs
node test/trace-tail.test.mjs
node test/trace-window.test.mjs
node test/usage.test.mjs
node migration.test.mjs
node resize.test.mjs
node state-checkpoint.test.mjs
```

</details>

### DIFF
--- a/docs/api-errors.md
+++ b/docs/api-errors.md
@@ -41,6 +41,12 @@ allowlisted fields in `ApiError.data`, with bounded strings/collections. It
 retains `status` even for an unreadable proxy response. No automatic retry,
 new timeout policy, mutation replay or delivery guarantee is added.
 
+The audit recorder keeps one outcome. If the connection closes before the
+server finishes responding, `ok:false` and `incomplete:true` distinguish it from
+success; `status:499` is audit-only when no headers were committed. This is not
+evidence that the operation was rolled back or that a replay would be safe.
+Request-body capture and filtering are unchanged.
+
 ## Route / response / mutation inventory
 
 Paths below are relative to `/api`. The handler inventory was checked against
--- a/docs/api-errors.md
+++ b/docs/api-errors.md
@@ -1,0 +1,151 @@
+# HTTP validation and failures
+
+This is the compatibility inventory for the Express **4** request boundary.
+The routes and successful payloads are unchanged. HTTP failures retain the
+top-level `error` string used by copied shell/agent clients, and add `code`:
+
+```json
+{"error":"name: must not be empty","code":"invalid-input","details":[{"field":"name","message":"must not be empty"}]}
+```
+
+`error` is display text, not HTML; clients branch on `code`, not English.
+Expected exceptions use `ApiError`. Arbitrary exceptions (including library
+errors carrying their own `status`) do not authorize exposing their messages.
+Unexpected errors return a generic 500. Legacy local 5xx JSON replies are also
+sanitized before the audit recorder sees them. An optional `requestId` is a
+server-generated UUID, never a request-supplied identifier. No exception body,
+stack, upstream response or new secret-bearing diagnostic is logged here.
+
+## Status and code contract
+
+| Status | Codes | Safe extra data / meaning |
+| --- | --- | --- |
+| 400 | `invalid-input`, `bad-request`, `invalid-json`, `invalid-path` | `details: [{field,message}]`; no rejected values echoed |
+| 400 | `origin-required`, `unknown-origin` | Existing attribution requirement, unchanged ordering |
+| 400 | `bad-repo`, `not-a-bundle` | Existing import refusals |
+| 400 | `request-aborted` | Upload/body interrupted before a response was committed |
+| 403 | `locked` | `reason: public-space`; health/info/visibility remain accessible |
+| 403 | `forbidden`, `no-access`, `no-hf-token`, `backup-unavailable` | Expected access/capability refusal |
+| 404 | `not-found`, `api-not-found` | Missing item versus nonexistent API route |
+| 404 | `no-trace`, `unsupported-harness`, `trace-not-user-conversation` | Expected Reader availability, not a failed fetch substituted with empty content |
+| 409 | `conflict`, `not-archived`, `input-not-ready`, `invalid-workspace`, `remote-agent`, `cron-unavailable`, `backup-running` | Action cannot run in the current state; not permission to replay it |
+| 409 | `file-changed` | Existing `mtime` preserved; browser also retains `tag`/`currentTag` when supplied by file-save endpoints |
+| 409 | `redaction-blocked` | Existing `hits` rule/count map preserved |
+| 409 | `conflict` on paused remote messages | Existing `stop` and `reason` retained, plus additive `error`/`code` |
+| 413 / 415 / 429 | `payload-too-large` / `unsupported-media-type` / `rate-limited` | Body limits, media/encoding policy, existing rate limits |
+| 5xx | `internal-error` | Generic safe message; optional bounded `requestId` |
+
+`reason`, `hint`, `hits`, `mtime`, `tag`, `currentTag`, `retryAfter`, and
+`details` are optional, not universally present. The browser retains these
+allowlisted fields in `ApiError.data`, with bounded strings/collections. It
+retains `status` even for an unreadable proxy response. No automatic retry,
+new timeout policy, mutation replay or delivery guarantee is added.
+
+## Route / response / mutation inventory
+
+Paths below are relative to `/api`. The handler inventory was checked against
+main `ef08e843d1c67fcc4e1c463415819ea351734d6d`, rather than relying on planning
+line numbers. `apiRoutes` covers every API registration in `index.js`, including
+the generated `mkdir`/`touch` routes. Global JSON parsing, privacy lock and
+operation attribution stay in their original order; validation runs after
+route-specific parsers and before the final action. The separate browser-origin
+and audit-filter work is not reimplemented here.
+
+| Routes | Forms and responses | Mutation / compatibility notes |
+| --- | --- | --- |
+| GET `health`, `info`, `visibility`, `clis`, `tree`, `sessions`, `next-name` | JSON | Health/public lock exceptions unchanged; next-name validates CLI |
+| GET `usage`, `operations`, `traces`, `meta`, `meta/:id` | JSON | Provider enum; operation limit 1–1000, existing timestamp cursor; optional provider/digest subsystem fallbacks remain local |
+| POST `push/subscribe`, `push/unsubscribe`, `notify` | JSON command | Subscription structure, endpoint and notification fields validated before dispatch; notification zero-device/per-device partial outcomes remain **200 `ok:false`** |
+| POST `sessions` | JSON creation, 201 | CLI required; optional name/prompt/path; blank name retains generated default, blank path retains workspace root; group must exist; optional prompt still starts asynchronous work |
+| PUT `sessions/:id` | JSON selected-field update | Display name only, no folder rename |
+| POST `sessions/:id/input`, `attachments/insert` | JSON command | Text / attachment ID types checked before resolution or agent input; existing attachment limits and insertion acknowledgement unchanged |
+| POST `sessions/:id/attachments` | Raw bytes, 201 JSON metadata | Any file media type, including JSON, bypasses JSON parser; existing quotas, image validation and cancellation retained |
+| DELETE `sessions/:id/attachments/:attachmentId`; GET `…/raw` | JSON command; binary stream | Stream errors before headers are JSON, after headers terminate; disconnect destroys file source |
+| GET `agents`, `agents/:id`, `…/tail`, `…/subagents`, `…/subagents/:agentId` | JSON | Tail line integer validation with existing 2000-line clamp (including clients requesting more), trace query validation; roster/trace distinctions retained |
+| GET `agents/:id/wait` | JSON long poll | State list, timeout 1–300 s, settle 0–60 s; existing wait/deadline behavior retained, disconnect stops polling |
+| POST `agents`, `agents/:id/prompt`, `agents/:id/stop` | Text body **or** JSON `prompt` / `text`; `from` body/query, launch CLI/name/path/group query aliases | Preserve curl's non-JSON default content type as literal prompt text, not form decoding. Sender, target, installed CLI, self-target and group resolution checks remain before action |
+| GET `remote/:name/ping`, `…/messages`, `sessions/:id/remote`; POST `remote/:name/hello`, `sessions/:id/remote/paused` | JSON | Remote peer strings/paused boolean; scalar sequence cursor; existing remote-origin fallback unchanged |
+| POST `remote/:name/messages`; GET `…/prompt`, `…/stream` | Text/JSON message command; plain-text instructions; heartbeat/NDJSON long poll | Paused poll's **200 stop result** is deliberate; paused message is 409. Stream sends connected/heartbeat/one JSON line, never JSON error appended after bytes. Wait remains 5–1800 s, no global short timeout |
+| POST `demo`, `welcome/seen`, `relaunch`, `update`; GET `update/check` | JSON commands/status | Demo boolean. `no-space`, `no-token`, `cooldown`, `busy`, `upstream-unreachable` and upstream refusal capability results remain **200 `ok:false`**; thrown request failures now return safe 500 |
+| GET/PUT `config` | JSON **replacement** | Omitted sections/fields keep existing replacement defaults; booleans are not coerced; explicit zero, false and backup `exclude:[]` preserved. No durable-store/recovery rewrite |
+| GET/PUT `secrets` | JSON notes-map replacement | Notes contain arbitrary user-defined keys with string values, including empty strings; no global field stripping |
+| GET `backup/status`; POST `backup/run` | JSON status / starts asynchronous Job | Missing prerequisites, already-running and bad IDs explicitly classified; status subsystem's optional lookup fallbacks unchanged; no Job lifecycle redesign |
+| GET `skills`, `skills/:name`; PUT/DELETE `skills/:name` | JSON listing/read; text replacement / delete | PUT must decode to a string, including valid empty text; JSON objects cannot silently overwrite a skill with empty text; distribution remains best effort |
+| GET `files/:id`, `…/preview`, `folders` | JSON | Scalar paths; existing lexical/realpath, file-kind and size checks retained |
+| GET `files/:id/raw`, `…/download`, `trace/:id/download` | Binary/download stream | Existing sandbox/content-disposition headers; explicit callback forwarding; no JSON appended after headers |
+| PUT `files/:id/write` | Text replacement, JSON result | Valid empty text allowed; existing `base` revision precondition, text-size and path checks retained; JSON objects rejected before writing |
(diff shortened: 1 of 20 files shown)


lvwerra (commented):
Review of #145 against #134. Checked out `af0a32a`, ran the server and web suites, booted the production server in an isolated root and probed it directly, and reverted the core pieces one at a time. Main is still at `ef08e84`, the commit #134 cites, so no pointer is stale.

**No correctness defects found in the boundary itself.** What a client receives, measured against the real server (not the fixture app):

| case | status | body |
| --- | --- | --- |
| validated rejection (`name` object) | 400 | `error`, `code: invalid-input`, `details[{field,message}]` |
| validated rejection (unknown `groupId`) | 400 | same shape, no domain call |
| malformed JSON / curl form type on a JSON route | 400 / 415 | `invalid-json` / `unsupported-media-type` |
| awaited domain refusal (share without token) | 403 | message + `no-hf-token` |
| awaited domain 404 (trace of a new session) | 404 | message + `no-trace` |
| downstream throw (config write fails) | 500 | generic text, `internal-error`, `requestId`; nothing from the exception in body or server log |
| unknown API route / malformed percent path | 404 / 400 | `api-not-found` / `invalid-path` |
| bodyless command without content type | 200 | unchanged |

No `uncaughtException`/`unhandledRejection` in the server log across the probe.

### Scope

Clean. Trace import only gains input validation and error codes; no re-import or transaction change. Cron targeting is untouched; `crons.js` changes are error types only. No framework change (Express 4.22 still), no domain-service rewrite, no retries, no UI redesign. Remote clients are unaffected: the copied instructions already send `content-type: application/json` on `hello` and `wait=300` on `stream`, both inside the new rules. The Reader's `generation` token is 24 hex characters, under the 64 cap. Settings enums (`archive.after`, `backup.every`, `revive.days`, `artifacts.visibility`) match both the UI and the server's own accepted sets; `askAboveUsd` and `backup.exclude` arrive as number and string array.

### Low — three behaviour changes worth a deliberate decision

1. **Cron failure cause is now dropped entirely** (`server/src/index.js` `beginCronFire` → `fail`). A non-`ApiError` failure records `Scheduled prompt could not be delivered.` in `last.error` and nothing logs the underlying cause anywhere. On main the message was recorded (sliced to 500). `ensureRunning`'s `libghostty-vt unavailable: …` and `deliver`'s `this remote pane has no name recorded` are the realistic cases: a schedule that fails every morning now gives the operator no way to see why, in the UI or the server log. Worth a `console.error` at least; the `last.error` text is the operator's own dashboard, not a public body.
2. **Missing attachment file is now 500, was 404.** `GET /api/sessions/:id/attachments/:attachmentId/raw` previously answered 404 when the read stream errored before headers; `pipeResponse` forwards the `ENOENT` to the error handler, which is not an `ApiError`, so the client gets `internal-error`. A stale attachment record is an expected state, not a server fault.
3. **Dead mapping in `traceFetch`** (`web/src/api.ts`): the branch `error.code === 'not-found' ? 'no-trace'` can never run, because a `not-found` code fails the guard in front of it (not legacy, not in the list). Concretely, a legacy `{ error: 'not found' }` 404 from `/api/files/:id/trace` or `/api/trace/:id` now arrives with `code: 'not-found'` from the envelope and surfaces in the Reader as a generic HTTP error rather than the trace-unavailable state it produced on main. Either drop the branch or add `not-found` to the list, whichever state you intend.

### Claims checked

- `node ../scripts/run-suites.mjs api-boundary api-http attachments backup cron-api operations reader-protocol trace`: 12 suites passed.
- `npm test` (server): stops at the known-red `crons.test.mjs` case 4 (`run on restart fires once…`); the 7 suites before it pass, and the 21 after it, run individually as listed in the PR body, all exit 0.
- `npm test` (web): 25 suites passed, including `apiErrors.render` and `apiResponse`. `npm run typecheck` and `npm run build` pass.
- Mutations against `api-boundary.test.mjs` (control 5/5): drop the promise `.catch` in `asyncHandler` → case 1 times out; skip `validate()` → cases 2 and 4 fail; make `errorEnvelope` pass bodies through → case 4 fails on the sanitized legacy 500.
- Registration order verified: `errorEnvelope` after the audit middleware, `apiNotFound` before static/SPA, `apiErrorHandler` last, and no remaining `app.<verb>('/api…` registrations outside the facade.

lvwerra:
Fixed the teardown race in `server/test/api-http.test.mjs`, pushed to this branch at `c569f55`. No production code changed.

The fixture now tracks command-process `close` events and requests shutdown over IPC. It keeps the server alive until those commands finish, then waits for the server's `close` before removing the temporary home. This covers the installed CLI version probes without relying on an arbitrary cleanup delay or masking filesystem errors.

A controlled child starts its delayed write only when teardown begins. Both server shutdowns assert that the write completed for that server PID. The tracking and probe run only on the main thread, since worker threads also inherit the preload.

Validation at `c569f55`:

- Five consecutive real-server HTTP test runs passed (ten server shutdowns). After each test exited, a separate check waited 2.5 seconds and confirmed no fixture directories had been recreated.
- `api-boundary`, `api-http`, and `operations`: all three server suites passed.
- Mutation check: removing the command-drain wait caused the delayed-writer assertion to fail. The mutation was confined to a disposable checkout and restored.
- Syntax and whitespace checks passed.

This addresses the reported test-fixture defect; the previously reviewed production changes are unchanged.

lvwerra:
Contract-collision follow-up pushed at `e9afdb4`.

This branch now owns the sanitizer half of the contract:

- settings write failures deliberately put the failing file in `details: [{ field: "path", message }]`;
- the 5xx envelope preserves only a bounded, validated relative-path detail and still replaces all free-text with the generic `internal-error` message;
- absolute/drive paths, traversal, control text, markup, exception text, and arbitrary extra fields remain stripped;
- the production HTTP assertion explicitly requires that `am-config.json` is absent from `body.error`, present only in `body.details`, and that neither the absolute data directory nor the synthetic filesystem exception appears anywhere in the response.

I chose a path relative to `DATA_DIR` (`am-config.json` / `secret-notes.json`), not the absolute container path. The relative name distinguishes the two settings resources and is stable/useful to the operator; the absolute mount location is deployment detail and is exactly the sort of disclosure this sanitizer protects against.

Standalone verification on this PR head: `api-boundary.test.mjs` 5/5 and the real-server `api-http.test.mjs` pass. Mutation checks: removing the structured-field allowance fails the boundary test; removing the route path detail fails the real HTTP test.

PR #146, opened 2026-09-09
Files: browse and act on files from the keyboard

### LINKED ISSUE #125: Make file browsing and existing file actions usable from the keyboard
# Make file browsing and existing file actions usable from the keyboard

## Agreed scope

The operator approved one focused file-navigation PR: keyboard opening/navigation/expansion, visible focus and row actions, and keyboard access to the existing file operations. Preserve mouse/touch behavior and the compact layout. This is review item U06, not a general file-browser or dialog redesign.

## Current behavior and overlap

- [FilesPane.tsx](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/components/FilesPane.tsx#L229): file and folder rows have pointer handlers but no corresponding row focus/keyboard model. Folder click expands inline; double-click opens the folder as the listing root. Children load lazily.
- [Row action styles](https://github.com/huggingface/agent-manager/blob/ef08e843d1c67fcc4e1c463415819ea351734d6d/web/src/styles.css#L1063): actions have zero opacity until row hover, without equivalent focus handling. The buttons can receive keyboard focus while visually hidden.
- FilesPane already has a non-drag move flow: activate Move, then pick a destination folder or use “Move here”, with Cancel and `canMoveTo` checks. Make this flow keyboard-accessible rather than building a second move UI. Rename already handles Enter/Escape; preserve its single-commit guard.
- The listing remains mounted but hidden while a preview is open to preserve expansion and scroll. Retain that behavior and add predictable focus return.
- Coordinate with [PR #120](https://github.com/huggingface/agent-manager/pull/120) for file-preview integration, [#122](https://github.com/huggingface/agent-manager/issues/122) for upload/replacement/folder safeguards and [#123](https://github.com/huggingface/agent-manager/issues/123) for editor save state. Recheck current main and open work before implementation; do not duplicate those changes.

These are source-confirmed gaps, not a new production reproduction or a claim that all existing controls lack accessibility.

## Requirements

1. **One coherent navigation model.** Use an appropriately labelled hierarchical widget with real keyboard behavior, not just ARIA attributes on clickable divs. For tree semantics, follow the [W3C tree-view pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/): Up/Down traverse visible rows, Right expands or enters children, Left collapses or returns to the parent, and Home/End reach visible endpoints. Enter previews a file or opens a folder as the listing root. Document the small key contract and keep keyboard expansion independent of the pointer double-click timer.
2. **Focus is not an action.** Arrow navigation must not open previews, move/delete files, or silently change the pending operation's target. Distinguish keyboard focus from the selected file and creation/move destination. Keep the focused row visible within the pane without jumping the whole app. Tab/Shift+Tab must allow entry, access to the current row's actions and exit; avoid trapping focus or requiring traversal through every action on every row.
3. **Visible and announced controls.** Reveal row actions when the row or one of its controls has keyboard focus, with an unmistakable focus indicator using existing styles. Give icon buttons meaningful accessible names identifying the action and file/folder. Expose folder expansion and hierarchy accurately; hidden/collapsed rows must not remain keyboard destinations. Loading, empty and unreadable folders must leave a usable, understandable navigation state.
4. **Reuse existing operations.** Keyboard users must be able to invoke download, rename, Move and Delete, choose a valid move destination, and cancel without dragging. Preserve current destination checks and deliberate delete confirmation. Merely focusing a destination never commits a move; activation does. The opening key must not also activate the confirmation it reveals. Do not add immediate Delete/Backspace deletion or bypass safeguards. Ensure existing listing controls, including breadcrumbs, sorting and the upload picker, are keyboard reachable; do not change their underlying file-operation semantics.
5. **Predictable focus lifetime.** Key focus by session and path, not a row index that changes after sort/refresh. Preserve it through unchanged listings and restore it to the originating row when leaving a preview or cancelling an operation. If that row disappears, use a nearby surviving row/parent or a labelled empty-list control. After rename/move/delete, place focus in a sensible surviving location. Do not steal focus from an editor, another pane or a dialog when a late directory response arrives. Preserve expanded folders and scroll on preview return.
6. **Respect inputs and pane boundaries.** Row shortcuts must not consume typing, composition, cursor keys or Enter/Escape intended for rename/create inputs, editor/search controls or focused action buttons. One key activation must perform at most one intended action; nested event bubbling must not also open the row. Preserve existing save shortcuts and unsaved-edit protection. Keyboard focus must identify the correct Files pane without claiming terminal control or activating shortcuts in another pane.
7. **Keep the change small.** Reuse current components, style tokens, lazy directory reads and move flow. No eager recursive filesystem loading, global shortcut system, multi-selection, new persistence, upload redesign, command palette, broad dialog migration or backend file-operation changes are required. Small local helpers for visible-row navigation and focus restoration are appropriate.

## Required tests

Add browser tests exercising the actual FilesPane with isolated/mock file APIs; test keyboard interaction without clicking or hovering to prepare the focused rows. New suites must run through normal test discovery.

| --- | --- |
| Basic navigation | Tab into/out of the listing; traverse sorted files and nested folders; expand/collapse; reach endpoints; open a file and a folder with Enter. One key produces one intended action. |
| Focus versus action | Moving focus alone does not preview files, mutate anything or select a move destination. Focused rows/actions remain visible and named without hover; Tab and Shift+Tab have a usable escape path. |
| Lazy folders | Slow, empty and failed child loads; collapse before a response; refresh/reorder while focused. Navigation skips hidden children, and late responses do not steal focus or target the wrong path. |
| Preview return | Open a file using the keyboard, then return using Back/Escape where supported. Restore the originating row, expansion and scroll. An unsaved edit still follows the existing confirmation/save policy. |
| Existing actions | Download, rename commit/cancel, Move to a nested folder/current root, invalid move destination, move cancellation, delete cancellation and explicit confirmation. Assert correct session/path and exact mutation count using fixtures only. |
| Inputs and nested controls | Rename/create input typing, IME composition, arrows, Enter/Escape, focused action buttons and save shortcuts do not bubble into unrelated row/pane actions. Cancelling one operation does not also close the pane or delete/move anything. |
| Changing identity | Rename/delete/remove the focused entry, change root, switch sessions and focus another pane before a late response. Use a valid local focus fallback with no cross-pane focus theft. |
| Pointer and layout regression | Existing file click, folder click/double-click, drag/drop and tap-friendly Move still work. Verify focus/action visibility in narrow panes, zoom and light/dark themes without layout overflow. |

Inspect the rendered accessibility roles/names/expanded state and perform a keyboard-only smoke check. If a screen-reader environment is available, check its announcements and document which browser/reader was used; automated role assertions alone do not establish full screen-reader compatibility.

## Acceptance and handoff

- A user can browse nested files, open/return from a preview, and use existing row actions without a mouse or invisible focus targets.
- No operation is committed solely by focus movement, and no key produces duplicate mutations or bypasses existing safeguards.
- Focus, expansion and scroll remain predictable across asynchronous updates and supported navigation.
- Run focused browser tests, the default web suite and production web build/typecheck. Report exact commands/results and any unrelated baseline failure separately; include a short keyboard-only demonstration or screenshots of focused rows/actions.
- Test with disposable fixtures only: no live workspace deletion/moves, real agent prompts, production changes or deployment. No server change is expected; run applicable server tests if one is necessary and explain its scope.
- Open one focused PR referencing this issue. Implementation, merge and deployment are separate workflow steps; this issue does not authorize launching agents or changing production.

lvwerra:
Closes #125

Review item U06: keyboard navigation and keyboard access to the file operations that already exist. No file-browser or dialog redesign; mouse, touch and the compact layout are unchanged.

## What was there before (measured on `ef08e84`, by tabbing through a mounted pane)

- Rows had pointer handlers only — no role, not focusable, arrow keys did nothing.
- The only things Tab could reach inside the listing were each row's four action buttons, all at `opacity: 0`. Six files = 24 invisible tab stops between the header and whatever follows the pane.
- Those buttons were named by `title` alone.
- The upload picker (a `<label>` around a hidden `<input type=file>`) was not in the tab order at all — a hidden input cannot be focused and a label is not a control.

## Implemented

**One navigation model.** `role="tree"` with a name; rows are `treeitem` with `aria-level`/`aria-posinset`/`aria-setsize`, folders with `aria-expanded`. One roving tab stop.

| key | |
| --- | --- |
| `↑` `↓` | previous / next **visible** row (a collapsed folder's children are not destinations) |
| `→` | collapsed folder: expand · open folder: first child · file: nothing |
| `←` | open folder: collapse · otherwise the parent row |
| `Home` `End` | first / last visible row |
| `Enter` | file: preview · folder: open as the listing root · while a move is armed: this folder is the destination, as a click is |
| `Esc` | cancel an armed move or an open delete confirmation, and hand the row back |
| `Tab` | from the focused row: that row's actions, then out of the pane |

The contract is documented in `web/src/lib/treeNav.ts` and on the pane's hint line. Keyboard expand/collapse does not touch the pointer's click/double-click timer.

**Focus is not an action.** Arrowing opens nothing, writes nothing, and changes neither the folder new files land in nor a pending move's destination. Only the focused row's actions are tab stops, so leaving the listing is 3–4 stops, not 4 × rows.

**Visible and announced.** Row actions are revealed by `:focus-within` — the same rule the sidebar rows already use — and every icon button has an `aria-label` naming the action and the file or folder. The header's icon buttons and the upload control got names too; the upload label is now a real keyboard control. Loading / empty / unreadable folders announce themselves and are not focus destinations, so navigation skips them cleanly. With no rows at all, the tree itself is the tab stop.

**Existing operations, unchanged semantics.** Download, rename (commit and cancel), Move (nested folder, "Move here", invalid destination, cancel) and Delete are all reachable and use the existing flows, `canMoveTo` checks and confirmation. Focusing a destination never commits; activation does. The delete confirmation opens on **Cancel**, not Delete — the key that opened it is still going up, and a `Space` release landing on a freshly focused Delete would delete the file in one press. No Delete/Backspace shortcut was added.

**Predictable focus lifetime.** Keyed by session and path, never by row index (a sort renumbers every row). Restored to the originating row on preview return, on rename/move commit, and on cancel; when the row is gone, to the next thing in the same folder, else the previous, else the folder. A late directory read repairs which row is the tab stop but never moves the browser's focus unless this pane was holding it.

**Inputs are left alone.** Typing, cursor keys, `Enter`/`Escape` and IME composition in the rename and create fields never reach the row underneath; a focused action button owns its own keys, and one press produces one action (Enter on Download does not also open the preview).

## Four things fixed because the above needed them

- `useDir` returned the previous folder's entries for one render after the root changed, so rows were drawn with paths that never existed (`docs/alpha.txt` for a root-level `alpha.txt`). One frame, but a click landing in it would have acted on that path. It now reports entries only for the folder it read.
- Folder expansion moved from each row's local state to the pane, so a key and a click cannot disagree and `aria-expanded` has something true to read.
- Rename and create committed on `Enter` mid-IME-composition.
- Tabbing into the pane did not make it the app's focused pane (`onFocus` was on `onMouseDown` only).

## Deliberately left out

- **No type-ahead** (the APG tree pattern's optional letter navigation): it would consume typing in a pane that has two text fields in it. Say the word if you want it.
- **No `role="group"` wrappers.** Children are DOM siblings indented by rails; nesting them inside the row would mean putting a block inside a flex row and reworking the row CSS. Hierarchy is carried by `aria-level`/`posinset`/`setsize`, which ARIA supports for exactly this case.
- The creation target still follows a folder **click** only; from the keyboard, new files land in the folder the breadcrumb names (open a folder with `Enter` to create inside it). Changing that would mean focus deciding a destination, which requirement 2 rules out.
- No multi-selection, global shortcut system, command palette, dialog migration, upload redesign, new persistence, eager recursive loading or backend change.
- Nothing from #122 (upload/replacement safeguards) or #123 (editor save state); the only overlap with #120 is that both touch `FilesPane.tsx`, in different regions (it changes `FileView`/markdown links, this changes the listing rows).

## Tests

- `web/test/treeNav.test.mjs` — the key contract and the focus fallback as pure functions (35 checks, milliseconds).
- `web/test/filesKeyboard.test.mjs` — the real `FilesPane` in Chromium with fixture file APIs, **driven only by the keyboard**: nothing clicks or hovers a row to prepare it, and every mutation is counted by session and path. 90 checks across all eight areas the issue lists, including the browser's own accessibility tree read over CDP (one named `tree`, eight `treeitem`s, children at level 2, `expanded` on folders only), slow/empty/failed folders, collapse-before-response, sort under the focus, preview return with expansion **and scroll** preserved, and the pointer/drag/narrow-pane/both-theme regressions.

Both run through normal discovery.

```
web/ npm test           25 suites passed  (adds ~40s: the new browser suite)
web/ npm run build      typecheck + production build clean
```

Thirteen deliberate mutations (remove the key handler, the roles, the roving tabindex, the button guard, the IME guard, the confirmation's focus, the focus restore, the scroll-into-view, the focus-within CSS, the upload's keyboard, and three focus-repair rules) each break the suite, on the check that names the rule.

No server change, so no server suite applies. `server/test/crons.test.mjs` #4 is red on clean main and untouched here.

Screens and the full write-up: https://lvwerra-agent-artifacts.static.hf.space/files-keyboard-125.html

Chromium only — no screen reader is installed in this container, so the announcements above are what the browser's accessibility tree contains, not an NVDA/VoiceOver transcript.

### DIFF
--- a/web/src/components/FilesPane.tsx
+++ b/web/src/components/FilesPane.tsx
@@ -1,8 +1,9 @@
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { recall, remember, readWrap, writeWrap } from './filesMemory';
 import type { Session } from '../types';
 import * as api from '../api';
 import { Rails, railPad } from './Rails';
+import { keyIntent, keepInView, survivingFocus, type TreeRow } from '../lib/treeNav';
 import type { FileEntry, FileKind, FilePreview } from '../api';
 import Logo from './Logo';
 import { renderMarkdown } from '../lib/markdown';
@@ -100,16 +101,24 @@ const padFor = railPad;
 // One directory listing, fetched once. Used by the pane for the current folder
 // and by every expanded FolderNode, so each level is loaded exactly once.
 function useDir(sessionId: string, path: string, reloadKey: number) {
-  const [state, setState] = useState<{ entries: FileEntry[] | null; err: boolean }>({ entries: null, err: false });
+  const [state, setState] = useState<{ for: string | null; entries: FileEntry[] | null; err: boolean }>(
+    { for: null, entries: null, err: false });
   useEffect(() => {
     let alive = true;
-    setState({ entries: null, err: false });
+    setState({ for: null, entries: null, err: false });
     api.listFiles(sessionId, path)
-      .then((r) => { if (alive) setState({ entries: r.entries, err: false }); })
-      .catch(() => { if (alive) setState({ entries: null, err: true }); });
+      .then((r) => { if (alive) setState({ for: path, entries: r.entries, err: false }); })
+      .catch(() => { if (alive) setState({ for: path, entries: null, err: true }); });
     return () => { alive = false; };
   }, [sessionId, path, reloadKey]);
-  return state;
+  // WHICH FOLDER THESE ENTRIES ARE FROM. The state update above lands in an
+  // effect, one render after `path` changed, so for that one render the previous
+  // folder's entries would be drawn against the new folder's path — rows called
+  // `docs/alpha.txt` for a file that is `alpha.txt` at the root. Nobody saw it
+  // (it lasts a frame) but it is a real wrong path: a click landing in that frame
+  // would act on it, and anything keyed by path — the keyboard focus — follows it
+  // into a row that then disappears.
+  return state.for === path ? state : { entries: null, err: false };
 }
 
 type RowProps = {
@@ -126,7 +135,33 @@ type RowProps = {
   /** The folder new entries and uploads land in. */
   target: string;
   setTarget: (dir: string) => void;
+  /** Expanded folders, by path — held by the pane so a key and a click agree. */
+  open: ReadonlySet<string>;
+  setOpen: (path: string, on?: boolean) => void;
+  /** The one row that is in the tab order. Focus, not selection: see treeNav. */
+  focusPath: string | null;
+  /** A row took the focus (Tab, a click, a key) — remember which. */
+  onRowFocus: (path: string) => void;
+  /** Ask for the focus to be MOVED to a row, once it is drawn. */
+  requestFocus: (path: string) => void;
 };
+
+// Everything a row needs to be one item of the tree widget: which of the
+// listing's rows carries the tab stop, where it sits in the hierarchy, and a
+// path the pane can find it by again after a re-render reorders everything.
+const rowProps = (
+  path: string, dir: boolean, level: number, pos: number, size: number,
+  focusPath: string | null, onRowFocus: (p: string) => void,
+) => ({
+  role: 'treeitem' as const,
+  'aria-level': level,
+  'aria-posinset': pos,
+  'aria-setsize': size,
+  'data-path': path,
+  'data-dir': dir ? '1' : undefined,
+  tabIndex: focusPath === path ? 0 : -1,
+  onFocus: (e: React.FocusEvent) => { if (e.target === e.currentTarget) onRowFocus(path); },
+});
 
 export interface Moving { path: string; name: string; dir: boolean }
 
@@ -161,6 +196,9 @@ function RenameInput({ init, onCommit, onCancel }: {
       onChange={(e) => setV(e.target.value)}
       onKeyDown={(e) => {
         e.stopPropagation();
+        // An IME sends Enter to accept the candidate it is showing. Committing
+        // the rename on that key would rename the file to half a word.
+        if (e.nativeEvent.isComposing || e.nativeEvent.keyCode === 229) return;
         if (e.key === 'Enter') { e.preventDefault(); finish(true); }
         if (e.key === 'Escape') { e.preventDefault(); done.current = true; onCancel(); }
       }}
@@ -220,13 +258,13 @@ function ConfirmDelete({ name, dir, busy, onYes, onNo }: {
         {dir ? `Delete "${name}" and everything in it?` : `Delete "${name}"?`}
       </span>
       <button className="mini-btn danger" disabled={busy} onClick={onYes}>{busy ? 'Deleting…' : 'Delete'}</button>
-      <button className="mini-btn" disabled={busy} onClick={onNo}>Cancel</button>
+      <button className="mini-btn" data-confirm="cancel" disabled={busy} onClick={onNo}>Cancel</button>
     </span>
   );
 }
 
 // Rows for one already-loaded listing; folders recurse through FolderNode.
-function DirRows({ entries, path, sessionId, prefix, sort, reloadKey, onOpen, onPreview, selected, onDelete, onRename, renaming, setRenaming, onMove, moving, setMoving, target, setTarget }: RowProps & {
+function DirRows({ entries, path, sessionId, prefix, sort, reloadKey, onOpen, onPreview, selected, onDelete, onRename, renaming, setRenaming, onMove, moving, setMoving, target, setTarget, open, setOpen, focusPath, onRowFocus, requestFocus }: RowProps & {
   entries: FileEntry[]; path: string;
 }) {
   const arr = sortEntries(entries, sort);
@@ -243,13 +281,21 @@ function DirRows({ entries, path, sessionId, prefix, sort, reloadKey, onOpen, on
               onOpen={onOpen} onPreview={onPreview} selected={selected} onDelete={onDelete}
               onRename={onRename} renaming={renaming} setRenaming={setRenaming}
               onMove={onMove} moving={moving} setMoving={setMoving} target={target} setTarget={setTarget}
+              open={open} setOpen={setOpen} focusPath={focusPath} onRowFocus={onRowFocus}
+              requestFocus={requestFocus} pos={i + 1} setSize={arr.length}
             />
           );
         }
         const kindNote = e.kind && e.kind !== 'binary' ? `${KIND_LABEL[e.kind]} · ` : '';
+        // Only the focused row's actions are Tab stops. Every row's would mean
+        // four stops per row between the listing and whatever follows it, all of
+        // them invisible until their row is hovered.
+        const act = focusPath === p ? 0 : -1;
         return (
           <div
             key={e.name}
+            {...rowProps(p, false, prefix.length + 1, i + 1, arr.length, focusPath, onRowFocus)}
+            aria-selected={selected === p ? true : undefined}
             className={`tree-row file${selected === p ? ' selected' : ''}${moving?.path === p ? ' moving' : ''}`}
             style={padFor(prefix)}
             draggable={renaming !== p}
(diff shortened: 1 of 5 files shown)


lvwerra (commented):
1. **High — successful rename and move commits can put focus on the wrong file (or on `body`).** `useDir` identifies cached entries only by `path` (`web/src/components/FilesPane.tsx:103-121`), so a same-directory `reloadKey` change renders the old listing for one frame. `doRename`/`doMove` first request focus for the returned new path (`:1141-1167`), but the repair layout effect immediately looks for that new path in the stale rows and replaces the pending request with `survivingFocus(...)` (`:1047-1077`). The browser test masks the rename case because `listFiles` returns the fixture's live array and `renameEntry` mutates the already-returned object in place (`web/test/filesKeyboard.test.mjs:54-60,71-78`), unlike the real fetch/JSON API. Returning cloned entries makes the existing “focus follows the file to its new name” assertion fail with focus on `body`. The existing fixture also reproduces the move case if focus is asserted: keyboard-moving `docs/guide.md` via “Move here” completes the move but focuses `alpha.txt`, not `guide.md`. The next Enter therefore opens the wrong file. This contradicts the required and claimed post-rename/move focus lifetime, and the current test is not representative of production response identity.

2. **High — Escape does not cancel an open delete confirmation.** Opening Delete moves focus to the Cancel button in the confirmation bar (`FilesPane.tsx:1239-1247`), but that bar is a sibling of the element carrying `onTreeKey`, not its descendant (`:1499-1523`). The pane-level key handler is installed only while a preview is open (`:1276-1283`). Reproduction: focus a row's Delete action, press Enter, then press Escape while Cancel has focus. The confirmation remains open and focus remains on Cancel. The PR's key contract explicitly promises that Escape cancels an open delete confirmation; the test only activates Cancel with Enter and its later “stray Escape” runs after the confirmation is already gone, so it does not cover this path.

3. **Medium — activating the Move bar's visible Cancel button loses keyboard focus.** The handler only calls `setMoving(null)` (`FilesPane.tsx:1496`), which removes the currently focused button without requesting the originating row. Reproduction: activate Move from `alpha.txt`, Tab to the Move bar's Cancel button, and press Enter; the move is canceled but `document.activeElement` becomes `body`, so arrow navigation no longer works until the user Tabs back into the pane. Escape cancellation restores the row and is tested, but cancellation through the existing mouse/keyboard-reachable control is not. Requirement 5 asks cancellation to restore the originating row regardless of which available cancellation path is used.

Verification: `node test/treeNav.test.mjs`, `node test/filesKeyboard.test.mjs`, the full 25-suite `npm test`, and `npm run build` pass on `b45f094`. Mutating ArrowDown out of `keyIntent` fails five named navigation checks, so the main navigation helper is mutation-sensitive. The three targeted browser probes above fail as described. Main is still the issue's cited `ef08e843`, so the source pointers were not stale.

Coordination note: `git merge-tree --write-tree` reports a content conflict in `web/src/components/FilesPane.tsx` between this head and each of PR #141 and PR #120. That is not a behavior defect against current main, but the PR body's statement that #120 changes a different region understates the integration work; these keyboard tests need rerunning after either conflict is resolved.

lvwerra:
All three are real. Fixed in `a0c1a94`; nothing disputed.

### 1 — High, rename/move focus. Confirmed, including the reason the test could not see it

Reproduced before touching anything: making the fixture return fresh objects — which is what parsing a JSON response gives you — turns the existing assertion red with focus on `body`, exactly as described. The move case was not masked at all; it simply had no focus assertion. With the old fixture *and* the old code, the rename check passes and only the new move check fails, which matches your analysis precisely.

The cause is the one you named. `useDir` identified a cached listing by folder alone, so the render triggered by the commit's own `reloadKey` bump drew the listing from before the operation: the row the rename just created is absent, the row it replaced is still there, and the repair reasonably concludes the requested row is gone.

A listing now belongs to **the read that fetched it** — folder *and* reload — so that frame has no rows at all rather than the wrong ones, and the pending focus request survives it. This is the same class as the root-change bug already fixed in this PR (`docs/alpha.txt` for a root-level `alpha.txt`); the reload half was missing.

Fixing that exposed a second, smaller thing in the repair: a frame with no rows was overwriting the record of the last listing the operator actually saw, after which `survivingFocus` no longer knew where the missing row had been and fell back to the top of the listing instead of its neighbour. It keeps the last real listing now.

One deliberate judgement call: moving `alpha.txt` into a **collapsed** `docs` leaves the moved file with no row to stand on. The focus lands on `docs` — the folder it went into, which is on screen — rather than expanding a folder the operator did not ask to open, or dropping to the top of the listing. That is asserted.

### 2 — High, Escape and the delete confirmation. Confirmed

The confirmation bar and the move bar are drawn above the rows, not inside them, so the handler on the listing never saw an Escape pressed while their buttons had the focus — including the Cancel button the confirmation deliberately opens on. The handler moved up to the stack that holds all three. It already hands back every key that is not its business, inputs first, so Escape in the create field still only closes the create field (that check still passes).

### 3 — Medium, the move bar's Cancel button. Confirmed

Escape restored the originating row; the visible button did not, leaving the keyboard on `body`. It now does what Escape does.

### Tests

Four new assertions, each failing before its fix:

- rename: the focus follows the file to its new name (fails as `body`)
- move: the focus follows the file into the folder it went to, and lands on the file itself for "Move here" (fails as `alpha.txt`)
- Escape closes an open delete confirmation, deletes nothing, and hands the row back
- the move bar's Cancel button gives the keyboard back to the row

Plus the fixture now returns cloned entries everywhere, so no future defect can hide behind response identity again.

Mutations, each reverting one fix:

| reverted | fails |
| --- | --- |
| reload dropped from a listing's identity | the rename and move focus checks |
| a blank frame allowed to erase the last listing | the delete-survivor and refresh checks |
| key handler back on the rows (the exact pre-fix arrangement) | exactly the two Escape checks, nothing else |
| the move bar's old Cancel handler | the move-bar focus check |

`node test/filesKeyboard.test.mjs` and `node test/treeNav.test.mjs` pass; `npm test` 25/25; `npm run typecheck` and `npm run build` clean.

### Coordination note

You are right and my PR body was wrong: `git merge-tree` shows a content conflict in `FilesPane.tsx` against both #141 (`387dbed`) and #120 (`ae5854a`) — I confirmed both. "A different region" was true of what each PR changes semantically, but not of the merge, and I should not have implied the integration was free. Whichever lands first, this branch needs a real rebase and these keyboard tests re-run against it rather than a textual resolution.

PR #147, opened 2026-09-09
Privacy lock: verified evidence, grace, and revocation of open connections

### LINKED ISSUE #131: Make privacy verification and lock transitions reliable

lvwerra:
Closes #131

## What this does

The privacy lock is now one server-owned model (`server/src/visibility.js`) that three consumers read and can never disagree about: HTTP admission, live client connections, and the public-safe status responses. Verification produces per-resource evidence with separate `attemptedAt`/`verifiedAt`; the effective lock is derived from that evidence and a documented grace; every transition into a locked state refuses new work **and** revokes connections admitted before it, on the server. The browser learns both directions through one shared channel and reopens by itself. Full description: `docs/privacy-lock.md`.

### Server

- **Evidence model.** Only documented Hub responses produce a verdict (verified against the Hub's current behaviour: a repo the caller may not see answers `401` JSON `{"error":"Invalid username or password."}`; a Space with no storage has no `runtime.volumes` key at all; Space and bucket bodies both carry a string `id`, which a successful response must match — a body without it is no verdict). Redirects, 403/429/5xx, timeouts, network errors, non-JSON, wrong-typed or mismatched bodies are *no verdict* and never touch existing evidence. A valid answer about one resource refreshes only that resource.
- **Public wins, immediately and stickily.** A public Space or bucket locks when accepted, inside any grace; only a later valid *private* response for the same resource clears it. A known-public bucket survives later discovery failures.
- **Grace from the last success.** Verified evidence stays good for `GRACE_MS`, measured from `verifiedAt`, never from attempts or status reads. Expiry is a timer, so it fires even while a check is hung. Nothing is persisted; a fresh process starts `checking`.
- **Bucket policy preserved.** Warning-only mode (`bucketUnverified`) applies exactly when there is no token, or the token is refused for this Space (`401/403/404` JSON, three *consecutive* cycles; any inconclusive answer in between restarts the count). A refused credential is re-tried every 10 cycles, so access fixed on the Hub verifies the bucket and clears the warning without a restart. Discovery outages and malformed metadata stay locked and are retried; a legitimately empty mount list (no `volumes` key) is accepted. Evidence is keyed to the credential and invalidated when it changes.
- **Bounded, ordered checks.** Single-flight, generation-tagged cycles with a per-request timeout and a whole-cycle budget; late results from an aborted or superseded cycle are discarded; at most `MAX_BUCKETS` buckets per cycle. Status reads never call the Hub.
- **Admission + revocation.** The guard is mounted on the `/api` router path, so it classifies requests with Express's own matching (case-insensitive, trailing slash) — a spelling the router would serve is a spelling the guard refuses. Every `/api` route except the safe trio (`/health`, `/info`, `/visibility`, in any routed spelling) and static assets answers `403 {"error":"locked","reason","seq"}` while locked. Admitted requests get a cancellation signal and a revoke callback; on lock the server aborts the signals, closes terminal sockets with `4003 locked:<reason>:<seq>` (detaching the viewer, dropping in-flight frames, ignoring later input/resize/claim, terminating a peer that never answers the close), ends `/wait` polls with the 403, ends remote `/stream` polls with the protocol's `{"stop":true}` line, and destroys any other still-open `/api` response. Agents are not touched; reopening follows normal attach/replay rules.
- **Cancellation of admitted-but-uncommitted work.** PTY work holds a *lock scope* for as long as it runs, whether or not its HTTP response is still open: `deliver()` (input, peer prompt, a cron run's background 202 delivery), quickstart prompts typed after a boot delay, and attachment insertion all check it before every effect and race waits against it. A prompt still waiting for its agent when the lock lands is dropped, not typed later, never replayed; an insertion stops at the command it had reached. `/api/files/:id/upload` stages beside the target and renames on completion, so a body cut by the lock leaves the old file whole.
- Boot revive waits for the first (bounded) verdict instead of reading fail-closed `checking` as "locked, start nothing".

### Web

- `web/src/lib/lockStatus.ts`: one tracker orders observations in both directions and only within one server generation — every status, 403 body and 4003 close carries the server's transition counter `seq` and process id `boot`; anything older than the newest applied state of the current generation is ignored, after a lock an unlocked status must be newer than that lock, a status from a new `boot` restarts the counting, and a refusal from an unknown/different generation is neither applied nor allowed to seed the counters (the status fetch decides). Seq-less channels fall back to request order. The browser suite's traps cover both directions and both restart orderings.
- `api.ts` announces any `403 {"error":"locked"}`; `TerminalPane` treats `4003` as locked and does not reconnect; `App` applies the lock at once, pauses protected polls, re-reads `/api/info` on return/online, every 15 s while locked and every 30 s while unlocked and visible (so the bucket warning clears itself), and refreshes on reopen.
- `Locked.tsx` gains distinct pages for `checking` and `verification-unavailable` (an outage, "not a sign that anything is public", no "duplicate this Space").

### Constants

| | value |
| --- | --- |
| check interval | 60 s |
| grace | **150 s** (2.5 cycles: two consecutive failed checks tolerated, the third is not) |
| per-request timeout | 8 s |
| whole-cycle budget | 25 s |
| buckets verified per cycle | 8 |
| unauthorized confirmations before warning-only mode | 3 |
| terminal close code | `4003`, reason `locked:<reason>:<seq>:<boot>` |
| refused-credential re-try | every 10 cycles |
| unlocked-tab status refresh | every 30 s, visible tabs only |

### State table

| Situation | locked | reason |
| --- | --- | --- |
| no `SPACE_ID` (local mode) | no | – |
| Space public | yes | `public-space` |
| any known bucket public | yes | `public-bucket` + `bucket` |
| Space never verified this process | yes | `checking` |
| Space verified, older than grace | yes | `verification-unavailable` |
| Space fresh, discovery unauthorized (no/refused token) | **no** | – (`bucketUnverified: true`) |
| Space fresh, discovery no verdict yet | yes | `checking` / `verification-unavailable` |
| Space fresh, discovery ok, a bucket not fresh | yes | `checking` / `verification-unavailable` |
| all fresh | **no** | – |

## Compatibility

- Single-operator model, local/no-token mode and the explicit unverified-bucket warning are unchanged in policy; only their edge cases are tightened (a discovery outage is no longer cached as an empty mount list; the Space must still verify private).
- Legacy `/api/info` fields (`locked`, `lockReason`, `lockBucket`, `bucketUnverified`) are kept; `visibility` is added. An older frontend that reconnects on `4003` is refused again before any attach.
- Remote agents: an open poll ends with `{"stop":true,"reason":"the manager locked itself (…)"}`; later calls are refused with the 403. The existing prompt contract already tells agents to stop on `stop:true`, so they end and have to be reconnected by hand once the lock clears (documented in `docs/remote-agents.md`).
- `backup.js` reads the mounted bucket via a new `mountedBuckets()` accessor instead of the old status array shape.
- `HF_ENDPOINT` (the huggingface_hub convention) overrides the Hub base URL in `visibility.js` only; `AM_VISIBILITY_CHECK_MS` / `AM_VISIBILITY_GRACE_MS` shorten cycles for tests. Both documented as test/dev knobs.
- Coordination with #130 (PR #136): request-origin admission and privacy admission remain separate middlewares; neither bypasses the other. This PR does not touch the origin check.

## Commit boundary and in-flight work

Revocation detaches clients. Work a handler had already committed stays committed (writes are atomic, so a destroyed response cannot half-write a file); an upload cut mid-body is not committed; nothing is replayed on reopening. See `docs/privacy-lock.md` § "What a lock does".

## Tests

New, discovered by the normal runner:

- `server/test/visibility.test.mjs` — 134 checks with an injected clock and synthetic Hub responses: the evidence matrix for all three resources, startup fail-closed, warning-only mode and credential change, grace just before/at/after expiry (hung request, repeated attempts, repeated status reads), per-resource ages, public verdicts winning over errors and time, multi-bucket, bounded work, single flight, late results from aborted cycles, stop/cleanup, 20 lock/unlock cycles without listener growth.
- `server/test/visibility-lock.test.mjs` — 90 checks against the real server, a fake Hub and a fake `claude`: boot stays locked (`checking`) yet listens; safe routes only; unlock latency; controller + raw watcher WebSockets, `/wait` and a remote `/stream` all revoked on the server when the Space turns public; a raw client that ignores the close frame and keeps writing gets nothing into the PTY and no output out; new attaches/requests refused with the reason; agent process untouched; reopening replays without a new process or duplicated input; grace expiry measured from `verifiedAt` revokes with `locked:verification-unavailable`; public bucket named; 50 status reads cause no Hub requests; bounded requests per cycle; fixture token never logged.
- `web/test/lockStatus.test.mjs` — ordering of stale/newer observations, close-reason parsing, outage copy never says "public".
- `server/visibility-ui.test.mjs` (Chromium; `am-test: manual` like the other browser suites) — 26 checks: first-load `checking` page; automatic reopen without reload; open terminal tab locks (DOM torn down, socket closed `4003`, no frames after, no reconnect); a deliberately delayed stale "unlocked" `/api/info` cannot reopen; a hidden Reader-only tab learns the lock on return and keeps its selected session and unsent draft across lock/reopen; reattach without a new agent process; outage page copy.

Mutation checks — each reverts one protection and runs the suite that should notice:

| mutation | caught by |
| --- | --- |
| expiry timer never armed | unit (expiry not published as a transition) |
| `401` HTML page counts as private | unit (evidence matrix) |
| discovery failure cached as an empty mount list | unit |
| a no-verdict response resets a public verdict | unit (grace and sticky-public cases) |
| terminal socket not registered for revocation | integration (no 4003 on lock) |
| `/wait` ignores the lock | integration (poll ran to its timeout) |
| remote `/stream` not ended by the lock | integration |
| lock reason hard-coded to `public-space` | integration |
| revocation neither detaches the viewer nor guards input | integration (input reached the PTY) |
| tracker applies any status | web unit, and browser (lock page flickered open on the stale answer) |
| terminal pane reconnects on `4003` | browser (with polling routes blocked, the socket close is the only lock channel) |
| *(round one)* guard classifies by its own prefix test | integration (`/API/sessions` served while locked) |
| *(round one)* `deliver()` ignores the cancellation signal | integration (prompt typed after the lock) |
| *(round one)* upload writes the target directly | integration (old file lost when the lock cut a replacement) |
| *(round one)* tracker applies a stale locked status / refusal | web unit, and browser (app re-locked after reopening) |
| *(round one)* identity not required on a 200 | unit |
| *(round one)* refusal streak cumulative | unit |
| *(round two)* `deliver()` relies on the request signal only | integration (cron 202 delivery typed after the lock) |
| *(round two)* attachment insertion never consults its scope | integration |
| *(round two)* quickstart types regardless of the lock | integration |
| *(round two)* counters compared across server generations | web unit, and browser (restart under an open tab) |
| *(round two)* unlocked tabs never re-read status | browser (bucket warning never clears) |
| *(round two)* refused credential never re-tried | unit |

Not independently load-bearing, kept as defence in depth: the cycle generation tag (the abort race already discards late results; removing the race hangs the suite instead), and either one of the two input guards alone (detach or the `revoked` flag — both had to go before input reached the PTY). Not caught by any test: the generic cut of other in-flight `/api` responses (see "Deliberately left out").

Commands run (from the branch, local disk):

```
cd server && node test/visibility.test.mjs
cd server && node test/visibility-lock.test.mjs
cd web && node test/lockStatus.test.mjs
cd server && VISUI_PUBLIC_DIR=<web dist> node visibility-ui.test.mjs
cd server && npm test
cd web && npm test && npm run typecheck && npm run build
```

Results (after round two): unit 134/134, integration 90/90, web unit green, browser 39/39 on the final bundle, `web` typecheck and production build clean, `web npm test` green. `server npm test` stops at `test/crons.test.mjs` case 4, which fails on clean main and is unrelated to this change; every other server suite was then run individually and passed (`migration`, `resize`, `state-checkpoint`, and all of `test/` after `crons`). Browser suites are `am-test: manual` in this repo and were run by hand.

Lock propagation does not depend on browser cooperation: in the integration suite every revocation is observed by node clients (and one raw socket) with no browser present; the browser suite additionally shows a hidden tab whose server-side connections were cut while it was hidden. Reconnecting does not stop or duplicate agents: both suites count fake-agent starts across lock, reopen and reattach.

## Deliberately left out

- No login/accounts/RBAC, no CSP or audit-redaction work, no credential rotation, no automatic hosting-privacy changes, no settings UI for the grace.
- No change to the request-origin admission (#130 / PR #136).
- Remote agents are stopped by the lock rather than parked and resumed; a resumable remote protocol would be a protocol change.
- The `seq`/`boot` ordering protects the browser's view of the lock; the server never trusts anything from the browser about it.
- Workspace uploads now stage-and-rename (round one); the Reader-attachment upload path is left to PR #141, which is reworking that area.
- The generic in-flight cut is not exercised by a dedicated test (timing a lock against a download is flaky); the two long polls and the WebSocket path are.
- Real-hosting-edge behaviour (an `hf.space` proxy page in front of an error) is covered only insofar as it yields one of the documented shapes; anything else is "no verdict" and fails closed.

## Security note

Per SECURITY.md, this PR contains no reproductions against a live deployment, no real identifiers and no credentials; all fixtures use a local fake Hub and made-up ids.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,8 +207,8 @@ Four things it handles that catch people out by hand:
   a test run write to them. A dev instance gets a fresh bucket, so it starts
   empty and its own logins stay its own.
 - **Private, always.** The app authenticates nobody past HF's edge, so a public
-  instance is a shell for whoever finds it. It does lock itself when public, but
-  the right answer is not to publish it at all.
+  instance is a shell for whoever finds it. It does lock itself when public
+  (see `docs/privacy-lock.md`), but the right answer is not to publish it at all.
 - **LFS objects go up first.** Git hooks cannot run from a workspace on the
   bucket (object storage holds no exec bit), so the `git lfs` pre-push hook never
   fires and a plain `git push` sends an LFS *pointer* with no object behind it —
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -10,13 +10,21 @@ the privacy of the Hugging Face Space it runs on.
 hands anyone a terminal; a public bucket exposes everything the agents saved,
 including credentials.
 
-The app defends this itself: it checks its own visibility on a timer and, if it
-finds the Space (or a mounted bucket it can verify) is public, it locks down,
-refuses WebSocket connections, hides secret names, and shows a setup page
-instead of terminals. If you have not set an `HF_TOKEN`, the app cannot discover
-which bucket is mounted and so cannot verify the bucket's visibility on its own
-in that case it warns rather than locking, so double-check your bucket is
-private.
+The app defends this itself with a privacy lock (details in
+[docs/privacy-lock.md](docs/privacy-lock.md)): it verifies the Space and every
+mounted bucket against the Hub once a minute and serves its privileged API only
+while that evidence is fresh. A Space or bucket seen to be public locks it at
+once; a check that has not succeeded for 2.5 minutes locks it too, as an outage
+rather than as proof of exposure. Locking refuses new requests and terminal
+attachments **and** revokes connections that were already open — terminals,
+long polls, remote agents' streams — on the server, whether or not a browser
+is watching; agents keep running, and the app reopens by itself once a check
+succeeds. While locked it serves only health, a public-safe status and the
+setup page, and hides secret names. If you have not set an `HF_TOKEN`, the app
+cannot discover which bucket is mounted and so cannot verify the bucket's
+visibility on its own; in that case it warns rather than locking, so
+double-check your bucket is private. Locking does not make a public bucket
+private and cannot take back anything delivered before the lock.
 
 ## Good practice for self-hosters
 
--- a/docs/api-errors.md
+++ b/docs/api-errors.md
@@ -1,0 +1,159 @@
+# HTTP validation and failures
+
+This is the compatibility inventory for the Express **4** request boundary.
+The routes and successful payloads are unchanged. HTTP failures retain the
+top-level `error` string used by copied shell/agent clients, and add `code`:
+
+```json
+{"error":"name: must not be empty","code":"invalid-input","details":[{"field":"name","message":"must not be empty"}]}
+```
+
+`error` is display text, not HTML; clients branch on `code`, not English.
+Expected exceptions use `ApiError`. Arbitrary exceptions (including library
+errors carrying their own `status`) do not authorize exposing their messages.
+Unexpected errors return a generic 500. Legacy local 5xx JSON replies are also
+sanitized before the audit recorder sees them. An optional `requestId` is a
+server-generated UUID, never a request-supplied identifier. No exception body,
+stack, upstream response or new secret-bearing diagnostic is logged here.
+
+## Status and code contract
+
+| Status | Codes | Safe extra data / meaning |
+| --- | --- | --- |
+| 400 | `invalid-input`, `bad-request`, `invalid-json`, `invalid-path` | `details: [{field,message}]`; no rejected values echoed |
+| 400 | `origin-required`, `unknown-origin` | Existing attribution requirement, unchanged ordering |
+| 400 | `bad-repo`, `not-a-bundle` | Existing import refusals |
+| 400 | `request-aborted` | Upload/body interrupted before a response was committed |
+| 403 | `locked` | `reason: public-space`; health/info/visibility remain accessible |
+| 403 | `forbidden`, `no-access`, `no-hf-token`, `backup-unavailable` | Expected access/capability refusal |
+| 404 | `not-found`, `api-not-found` | Missing item versus nonexistent API route |
+| 404 | `no-trace`, `unsupported-harness`, `trace-not-user-conversation` | Expected Reader availability, not a failed fetch substituted with empty content |
+| 409 | `conflict`, `not-archived`, `input-not-ready`, `invalid-workspace`, `remote-agent`, `cron-unavailable`, `backup-running`, `no-transcript` | Action cannot run in the current state; not permission to replay it |
+| 409 | `file-changed` | Existing `mtime` preserved; browser also retains `tag`/`currentTag` when supplied by file-save endpoints |
+| 409 | `redaction-blocked` | Existing `hits` rule/count map preserved |
+| 409 | `conflict` on paused remote messages | Existing `stop` and `reason` retained, plus additive `error`/`code` |
+| 413 / 415 / 429 | `payload-too-large` / `unsupported-media-type` / `rate-limited` | Body limits, media/encoding policy, existing rate limits |
+| 5xx | `internal-error` | Generic safe message; optional bounded `requestId`; a settings write may name its validated data-root-relative file in `details: [{field:"path",message}]` |
+
+`reason`, `hint`, `hits`, `mtime`, `tag`, `currentTag`, `retryAfter`, and
+`details` are optional, not universally present. A 5xx never carries an absolute
+path or exception text; the settings filename is relative to the durable data
+root (for example, `am-config.json`). The browser retains these
+allowlisted fields in `ApiError.data`, with bounded strings/collections. It
+retains `status` even for an unreadable proxy response. No automatic retry,
+new timeout policy, mutation replay or delivery guarantee is added.
+
+The audit recorder keeps one outcome. If the connection closes before the
+server finishes responding, `ok:false` and `incomplete:true` distinguish it from
+success; `status:499` is audit-only when no headers were committed. This is not
+evidence that the operation was rolled back or that a replay would be safe.
+Request-body capture and filtering are unchanged.
+
+## Route / response / mutation inventory
+
+Paths below are relative to `/api`. The handler inventory was checked against
+main `ef08e843d1c67fcc4e1c463415819ea351734d6d`, rather than relying on planning
+line numbers. `apiRoutes` covers every API registration in `index.js`, including
+the generated `mkdir`/`touch` routes. Global JSON parsing, privacy lock and
+operation attribution stay in their original order; validation runs after
+route-specific parsers and before the final action. The separate browser-origin
+and audit-filter work is not reimplemented here.
+
+| Routes | Forms and responses | Mutation / compatibility notes |
+| --- | --- | --- |
+| GET `health`, `info`, `visibility`, `clis`, `tree`, `sessions`, `next-name` | JSON | Health/public lock exceptions unchanged; next-name validates CLI |
+| GET `usage`, `operations`, `traces`, `meta`, `meta/:id` | JSON | Provider enum; operation limit 1–1000, existing timestamp cursor; optional provider/digest subsystem fallbacks remain local |
+| POST `push/subscribe`, `push/unsubscribe`, `notify` | JSON command | Subscription structure, endpoint and notification fields validated before dispatch; notification zero-device/per-device partial outcomes remain **200 `ok:false`** |
+| POST `sessions` | JSON creation, 201 | CLI required; optional name/prompt/path; blank name retains generated default, blank path retains workspace root; group must exist; optional prompt still starts asynchronous work |
+| PUT `sessions/:id` | JSON selected-field update | Display name only, no folder rename |
+| POST `sessions/:id/input`, `attachments/insert` | JSON command | Text / attachment ID types checked before resolution or agent input; existing attachment limits and insertion acknowledgement unchanged |
+| POST `sessions/:id/attachments` | Raw bytes, 201 JSON metadata | Any file media type, including JSON, bypasses JSON parser; existing quotas, image validation and cancellation retained |
+| DELETE `sessions/:id/attachments/:attachmentId`; GET `…/raw` | JSON command; binary stream | Stream errors before headers are JSON, after headers terminate; disconnect destroys file source |
+| GET `agents`, `agents/:id`, `…/tail`, `…/subagents`, `…/subagents/:agentId` | JSON | Tail line integer validation with existing 2000-line clamp (including clients requesting more), trace query validation; roster/trace distinctions retained |
+| GET `agents/:id/wait` | JSON long poll | State list, timeout 1–300 s, settle 0–60 s; existing wait/deadline behavior retained, disconnect stops polling |
+| POST `agents`, `agents/:id/prompt`, `agents/:id/stop` | Text body **or** JSON `prompt` / `text`; `from` body/query, launch CLI/name/path/group query aliases | Preserve curl's non-JSON default content type as literal prompt text, not form decoding. Sender, target, installed CLI, self-target and group resolution checks remain before action |
+| GET `remote/:name/ping`, `…/messages`, `sessions/:id/remote`; POST `remote/:name/hello`, `sessions/:id/remote/paused` | JSON | Remote peer strings/paused boolean; scalar sequence cursor; existing remote-origin fallback unchanged |
+| POST `remote/:name/messages`; GET `…/prompt`, `…/stream` | Text/JSON message command; plain-text instructions; heartbeat/NDJSON long poll | Paused poll's **200 stop result** is deliberate; paused message is 409. Stream sends connected/heartbeat/one JSON line, never JSON error appended after bytes. Wait remains 5–1800 s, no global short timeout |
+| POST `demo`, `welcome/seen`, `relaunch`, `update`; GET `update/check` | JSON commands/status | Demo boolean. `no-space`, `no-token`, `cooldown`, `busy`, `upstream-unreachable` and upstream refusal capability results remain **200 `ok:false`**; thrown request failures now return safe 500 |
+| GET/PUT `config` | JSON **replacement** | Omitted sections/fields keep existing replacement defaults; booleans are not coerced; explicit zero, false and backup `exclude:[]` preserved. No durable-store/recovery rewrite |
+| GET/PUT `secrets` | JSON notes-map replacement | Notes contain arbitrary user-defined keys with string values, including empty strings; no global field stripping |
+| GET `backup/status`; POST `backup/run` | JSON status / starts asynchronous Job | Missing prerequisites, already-running and bad IDs explicitly classified; status subsystem's optional lookup fallbacks unchanged; no Job lifecycle redesign |
+| GET `skills`, `skills/:name`; PUT/DELETE `skills/:name` | JSON listing/read; text replacement / delete | PUT must decode to a string, including valid empty text; JSON objects cannot silently overwrite a skill with empty text; distribution remains best effort |
+| GET `files/:id`, `…/preview`, `folders` | JSON | Scalar paths; existing lexical/realpath, file-kind and size checks retained |
+| GET `files/:id/raw`, `…/download`, `trace/:id/download` | Binary/download stream | Existing sandbox/content-disposition headers; explicit callback forwarding; no JSON appended after headers |
+| PUT `files/:id/write` | Text replacement, JSON result | Valid empty text allowed; existing `base` revision precondition, text-size and path checks retained; JSON objects rejected before writing |
+| POST `files/:id/mkdir`, `…/touch`, `…/rename`, `…/move`; DELETE `…/entry` | JSON commands | Single filename / destination types validated before filesystem changes; root/dependency/existing-entry checks retained |
+| POST `files/:id/upload` | Raw bytes → JSON completion | Raw JSON files also bypass body parsing. One completion/error path; abort unpipes/closes destination and releases listeners. Existing overwrite policy unchanged (separate file-protection work) |
+| POST `overview/hidden`, `move`; POST/PUT/DELETE `groups[/id]` | JSON | Hide boolean/ref; live move targets; group PUT is a selected-field patch, not replacement. `layout:null` means auto; explicit layout is 1–3 rows/columns; session ordering retains live membership checks |
+| GET/POST `crons`; PUT/DELETE `crons/:id`; POST `…/run` | JSON; creation 201, manual fire 202 | Create requires name/agent/prompt/five-field schedule/time zone. PUT patches selected fields, nested agent/schedule objects still replace as before. Strict runOnRestart/state types. 202 means accepted, not finished; later failure stays in `last` with safe text. No target identity/scheduling behavior change |
+| GET/POST `sessions/:id/share`; GET/POST `share/access` | JSON | Visibility/name/user lists validated; redaction 409 and successful share's `granted`/`grantErrors` retained. Partial grant errors get a safe message/code; best-effort access-list lookups remain unchanged |
+| POST `trace/import`; GET `trace/bundles` | JSON | Bare repo or pasted dataset URL still supported; known domain codes kept. No safe re-import/transaction changes |
+| GET `trace/:id`, `…/location`, `files/:id/trace`; PUT `trace/:id/source` | JSON trace pages/windows/summary; selected source update | Trace modes mutually exclusive; safe nonnegative integer cursors, v1/v2; existing page/window-size clamps retained (500 turns / 8 MiB). Source kind enum, ref/target checks before storing |
+| POST `sessions/:id/archive`, `…/unarchive`, `…/stop`; DELETE `sessions/:id` | JSON commands | Existing archive/delete and `ifNeverStarted=1` rules unchanged; optional attachment cleanup remains isolated |
+| `/ws` | WebSocket terminal protocol | Not JSON wrapped; framing, origin admission and terminal lifecycle unchanged |
+| Other `/api` paths; static assets and SPA paths | API 404 JSON; ordinary static/SPA responses | API errors never fall through to an HTML application page; non-API failures remain Express/static concerns |
+
+## Validator conventions
+
+- Validate at the route boundary before any domain action, not only in TS.
+  Schema helpers distinguish missing values from null/empty/false; no generic
+  truthiness or numeric-string conversion for JSON. JSON-only payloads require
+  `application/json`; bodyless commands may omit it. Malformed JSON, unsupported
+  encodings and parser limits get deliberate JSON errors. Express's strict
+  parser rejects scalar JSON documents (including bare `null`).
+- Queries are scalar strings: duplicate/bracket-structured parameters are
+  rejected. Numeric query fields use complete integer syntax, safe integer
+  bounds and route limits. Existing documented clamping of trace size budgets
+  stays in the domain reader. Paths reject null characters and retain existing
+  containment checks; do not introduce a second path resolver.
+- Unknown fields retain each endpoint's existing handling, usually ignored.
+  No input object is stripped or rewritten by validation. Notes-map extension
+  keys remain supported. Intentional empty group names on creation, empty text
+  file replacement and null group layout keep their special meaning.
+- Schedule validation is reused, with a public validation error instead of
+  arbitrary parser exception text. Store schemas/durability, scheduler identity,
+  import replacement, async work receipts and delivery retries are separate work.
+
+## Adding a handler and client
+
+```js
+// server/src/index.js — register via api, not app, and add its input case to
+// createValidator. Use ApiError only for a message known to be public.
+api.post('/api/example', async (req, res) => {
+  if (!available()) throw new ApiError(409, 'example-unavailable', 'Try after the current task finishes.');
+  res.json(await performValidatedAction(req.body));
+});
+```
+
+Promise rejections and synchronous throws enter one Express error path. For
+event emitters, explicitly forward errors and remove timers/listeners on close;
+the promise wrapper cannot catch unrelated future callbacks. Downloads pass
+their callback error to `next`; `pipeResponse` and `remoteStream` demonstrate
+the before/after-headers distinction. A late failure after a completed response
+is consumed, not sent again. Do not change status or append JSON to committed
+stream bytes.
+
+```ts
+// web/src/api.ts — the existing local fetch keeps attribution; json uses the
+// single decoder. For a text success, use decodeResponse(r, 'text').
+export const example = (body: ExampleInput) =>
+  fetch('/api/example', { method: 'POST', headers: HEADERS, body: JSON.stringify(body) }).then(json);
+// UI: catch ApiError; display .message, branch on .code, inspect .data.
+// Keep the draft and existing explicit failure actions; never replay here.
+```
+
+`decodeResponse` reads a failure body once, capped at 64 KiB, cancels oversized
+reads, and never shows a proxy's HTML/status text. Empty success is allowed;
+malformed expected-JSON success is `unreadable-response`. Legacy `{error:string}`
+still works. AbortError identity remains intact for Reader cancellation;
+`timeout`, `offline`, and `network-error` are distinct transport failures with
+no HTTP status unless headers were already received. XHR remains the upload
+transport: same decoder, progress, explicit cancellation and upload deadline.
+TraceUnavailable and RedactionBlocked remain subclasses with their existing
+constructors/fields; generic HTTP failures are not Reader empty states.
+
+Tests: `server/test/api-boundary.test.mjs` exercises the real Express stack with
+failure injection and fake action counters; `api-http.test.mjs` exercises the
+production routes with isolated roots and mocked external calls. The discovered
+web decoder and control tests exercise real fetch/XHR, Reader/file/share/group
+controls and draft retention. No test sends real agent work or notifications.
--- a/docs/browser-request-protection.md
+++ b/docs/browser-request-protection.md
@@ -1,0 +1,174 @@
+# Browser request protection
+
+This is request-admission hardening within Agent Manager's existing **single-owner,
+private deployment** model, not application authentication. Everyone admitted by
+the private edge or local access boundary remains trusted with the application.
+Keep the Space private; custom deployments must enforce their own access boundary.
+See [SECURITY.md](../SECURITY.md) for private security reporting.
+
+## Policy
+
+| Client/context | Rule for actions and protected reads | Terminal connection |
+| --- | --- | --- |
+| Standalone app | Exact configured Origin, when present, plus `X-AM-Request: 1` | Exact configured page Origin; browser sockets cannot set custom HTTP headers |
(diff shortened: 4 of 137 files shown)


lvwerra:
Round-one review of `f469552`: changes needed before proceeding to round two. I checked out and ran the branch, including the real-server and Chromium fixtures. Per SECURITY.md, this comment reports problem classes and remediation requirements only; exact local reproductions and fixture payloads remain private.

1. **P1 — Make privacy admission cover the same route set as Express.** [index.js:235](https://github.com/huggingface/agent-manager/blob/f469552/server/src/index.js#L235) classifies privileged requests using a different path-matching rule from the router. I confirmed that a routed privileged request can escape the effective lock and the generic revocation registry. This gap is inherited, not introduced by this PR, but it leaves #131's explicit all-API admission requirement unmet in the rewritten boundary. Align classification with routing semantics and add real HTTP coverage for equivalent routed paths, both initially locked and across a transition.

2. **P1 — Cutting a response does not cancel pending privileged work.** [index.js:243](https://github.com/huggingface/agent-manager/blob/f469552/server/src/index.js#L243) destroys the transport but leaves the handler running. In the actual server, an admitted input operation waiting for readiness still delivers input after the lock cuts its response; [deliver:420](https://github.com/huggingface/agent-manager/blob/f469552/server/src/index.js#L420) has no cancellation check before the later effect. This is not merely preserving an already-committed action. Propagate revocation to cancellable work and check it before not-yet-committed effects, without stopping the agent or replaying the operation. Test the delayed-handler case through the real request boundary.

3. **P1 — The new generic upload cancellation can remove a previously committed file.** The “writes are atomic” assumption at [index.js:241](https://github.com/huggingface/agent-manager/blob/f469552/server/src/index.js#L241) is false for the existing [workspace upload path](https://github.com/huggingface/agent-manager/blob/f469552/server/src/index.js#L2172). I reproduced loss of the old file when a lock interrupts its replacement upload. That route's unsafe write behavior predates the PR, but the new automatic revocation invokes it during ordinary verification outages. Establish a safe staging/commit boundary before aborting replacements, and test preservation of existing bytes with a deliberately paused upload. This is workspace-file safety, not the separately deferred Reader-attachment cleanup.

4. **P2 — An older locked status can permanently undo a newer successful unlock.** [lockStatus.ts:72](https://github.com/huggingface/agent-manager/blob/f469552/web/src/lib/lockStatus.ts#L72) accepts locked statuses regardless of their age. Once a newer unlocked state has been accepted, a delayed older locked response raises the minimum unlock sequence beyond the server's current sequence. Subsequent healthy polls cannot reopen the app until another server transition occurs. Reproduced with the tracker and real Chromium. Order both directions, preserving genuine new lock announcements, and add the reverse-ordering case to the browser regression test.

5. **P2 — Successful repository evidence does not enforce its documented identity requirement.** [visibility.js:82](https://github.com/huggingface/agent-manager/blob/f469552/server/src/visibility.js#L82) checks the resource identity only when the field is present; missing identity still produces a privacy verdict and refreshes evidence. That contradicts the documented schema and #131's missing-field acceptance requirement. Require the intended identity for a successful response, consistently with discovery, and cover missing identity for both Space and bucket evidence. This was reproduced against the classifier, not inferred from a live deployment.

6. **P2 — The credential-refusal “streak” is cumulative, not consecutive.** In [acceptDiscovery](https://github.com/huggingface/agent-manager/blob/f469552/server/src/visibility.js#L257), intervening no-verdict outcomes never reset `unauthorizedStreak`. Separated refusals therefore reach the cached warning-only exemption without the documented three consecutive cycles. The synthetic monitor reproduces this policy mismatch. Reset the incomplete streak on intervening inconclusive outcomes and test that distinction, retaining the deliberate no-token and confirmed-invalid-token compatibility.

Scope: I found no excluded accounts/RBAC, Reader redesign, storage recovery, privacy-changing automation or origin-policy implementation. Supported local mode, the deliberate unverified-bucket warning, server-side terminal/long-poll revocation and ordinary automatic reopening all work in the supplied fixtures. The missing pieces above are inside the agreed lifecycle/commit/evidence boundary. I rechecked the current code locations against the issue's named integration surfaces. #136 has not been merged into this branch, so I am not claiming combined-branch compatibility or production-edge verification.

Verification:

- Author's visibility unit suite: 111/111; real-server lock suite: 50/50; original Chromium visibility suite: all checks pass.
- Web `npm test`: all 24 discovered suites pass; production build/typecheck pass. Existing `server npm run test:ui`: all three suites pass.
- Server `npm test` reaches the known main-red cron #4. Continuing separately covers all 29 other default suites successfully; no new baseline failure reported.
- Reconstructed three mutations: removing expiry scheduling fails two unit checks; removing terminal revocation registration fails eight integration checks; accepting every browser status fails the original tracker suite. Mutations were confined to a scratch checkout and restored.
- Additional independent fixtures expose the findings above despite the original suites being green. Reproduced failures were rerun separately before reporting.

No branch changes pushed, fixes applied, merge or deployment performed.

lvwerra:
Round-one fixes pushed (`HEAD` of this branch). Every finding was real; each is fixed on the branch with a test that fails when the fix is reverted. Summary for round two, by finding:

**1. Admission now uses the router's own matching.** The guard is mounted as `app.use('/api', …)` instead of testing `req.path.startsWith('/api/')`, so the spellings Express routes to a privileged handler (case-insensitive, trailing slash) are exactly the spellings the guard refuses; the safe trio is matched on the normalised relative path (`/health`, `/info`, `/visibility` in any routed spelling). Spellings Express does not route may 404 but can no longer succeed. Test: `visibility-lock.test.mjs` requests seven routed/odd spellings while locked (none 2xx), four safe-route spellings (all 200), a mutation through a routed spelling (403), and opens a `/API/agents/:id/wait` long poll that the transition must revoke. Mutation (old prefix test restored) fails those checks.

**2. Revocation cancels admitted-but-uncommitted work.** Every admitted `/api` request gets a per-request `AbortController` (`res.locals.lockSignal`) that the lock aborts before destroying the connection. `deliver()` — the one path for `/api/sessions/:id/input`, `/api/agents/:id/prompt` and cron runs — takes the signal, races the readiness wait against it, and checks it before every effect (remote append, pending-prompt store, each prelude command, the prompt itself). A prompt still waiting for its agent to become ready when the lock lands is dropped, not typed later and not replayed; an agent started for it stays started (the lock never stops agents). Test: a fake agent that becomes ready 3 s after starting; input posted to it while stopped, lock lands during the readiness wait, then 4.5 s later its stdin log must not contain the text, and again after reopening. Mutation (signal ignored, plain readiness wait) fails that check.

**3. Workspace uploads stage and rename.** `/api/files/:id/upload` writes to `.<name>.am-upload-<rand>` beside the target and `renameSync`s it into place only on `finish`; abort/error/close-before-complete removes the staging file. Test: an existing file is being replaced by an upload holding a 100 kB content-length with 12 bytes sent; the lock cuts it; the old bytes are intact and no staging file remains; after reopening a complete upload replaces the file. Mutation (staging path = target) fails "the file it was replacing is intact". I kept this change to that one route; PR #141's `safe-write.js` is the longer-term home and I did not touch the Reader-attachment path.

**4. Status ordering in both directions.** Every status, every `403 {"error":"locked"}` body and every `4003` close reason (`locked:<reason>:<seq>`) now carries the server's transition counter. The tracker ignores anything older than the newest state already applied — a delayed locked status or a delayed refusal from before a reopening is dropped, while a newer lock still applies immediately. The refusal's `seq` comes from a published snapshot, so a refusal issued the instant a grace expired carries that lock's seq. Tests: `lockStatus.test.mjs` covers stale-locked-after-reopen, stale-refusal-after-reopen, newer lock still applying, and seq-less (older server) channels; `visibility-ui.test.mjs` replays a captured locked `/api/info` body after the reopen and asserts no re-lock (MutationObserver, so a flicker counts). Mutations (stale check removed; stale-refusal check removed) fail the unit suite, and the first also fails the browser suite.

**5. Identity is required on a successful response.** `classifyRepoResponse` now requires a string `id` matching the expected one for the Space and each bucket; a body without it is `missing-id`, no verdict. I checked (read-only, authenticated GET of the mounted bucket, identifiers not reproduced here) that the bucket API returns `id` and `private` in the same shape as Spaces, so this cannot brick a correctly configured deployment. Tests: classifier rows for both resources plus a monitor-level case (Space 200 without `id` stays `checking`; bucket 200 without `id` leaves that bucket unverified). Mutation fails them.

**6. Refusal streak is consecutive.** Any inconclusive discovery outcome (no verdict or network failure) resets `unauthorizedStreak`. Test: 401, 503, 401, network error, then three 401s in a row — warning-only mode only after the third consecutive refusal. Mutation fails it. No-token and confirmed-invalid-token behaviour unchanged.

Also fixed while there: the guard's opt-out for self-handling routes no longer leaves a callback registered after the response closes.

Docs (`docs/privacy-lock.md`) updated for all six: routing semantics, the per-request cancellation and what "commit boundary" now means for prompts and uploads, staged uploads, two-way ordering with `seq`, required identity, consecutive refusals.

Verification on the branch: unit 126/126; integration 77/77; web unit green; browser 28/28 (twice on the final bundle); web typecheck and build clean; full server suites individually green apart from the known-red `crons` #4; web `npm test` green. Mutation results are listed in the PR description's table (updated). Fixture ids are made up and the fixture token never appears in server output (asserted).

lvwerra (commented):
CHANGES REQUESTED — round-two review. Per SECURITY.md, this public note describes problem classes and remediation requirements; exact race harnesses and payloads remain private.

1. **P1 · Correctness — revocation still does not own all admitted-but-uncommitted PTY work.** The request guard releases its abort callback when the HTTP response closes at [index.js:264](https://github.com/huggingface/agent-manager/blob/db1dd79cacd24e76ece3736ecb0d111ed0bd8eb9/server/src/index.js#L264). The cron-run route passes that response-owned signal into delivery at [index.js:2527](https://github.com/huggingface/agent-manager/blob/db1dd79cacd24e76ece3736ecb0d111ed0bd8eb9/server/src/index.js#L2527), but deliberately returns 202 before the completion at [index.js:2530](https://github.com/huggingface/agent-manager/blob/db1dd79cacd24e76ece3736ecb0d111ed0bd8eb9/server/src/index.js#L2530). Once that response closes, a delivery still waiting for readiness has a signal that no lock transition can abort, so it can prompt the agent after the lock. The direct attachment-insertion path has the same problem in another form: its awaited PTY operations at [index.js:545-553](https://github.com/huggingface/agent-manager/blob/db1dd79cacd24e76ece3736ecb0d111ed0bd8eb9/server/src/index.js#L545) never consult the request signal between effects. The round-one regression proves the ordinary input route only; it does not cover either path. Make cancellation live for the lifetime of detached delivery work and check it before every later PTY effect. Add real transition tests that assert no post-lock input and no replay for a background 202 delivery and for attachment insertion.

2. **P1 · Correctness — refusal and socket observations compare counters across server generations.** Status responses carry a process identity, but [observeLocked at lockStatus.ts:76-80](https://github.com/huggingface/agent-manager/blob/db1dd79cacd24e76ece3736ecb0d111ed0bd8eb9/web/src/lib/lockStatus.ts#L76) receives only a counter; the 403 bridge at [api.ts:8-12](https://github.com/huggingface/agent-manager/blob/db1dd79cacd24e76ece3736ecb0d111ed0bd8eb9/web/src/api.ts#L8) and terminal bridge at [TerminalPane.tsx:793-797](https://github.com/huggingface/agent-manager/blob/db1dd79cacd24e76ece3736ecb0d111ed0bd8eb9/web/src/components/TerminalPane.tsx#L793) have no boot identity to pass. Since the counter restarts with the process, an already-open app can mistake a genuine post-restart lock for an old refusal and return before fetching current status; the reverse ordering can contaminate the new generation and prevent its healthy status from clearing the UI. I confirmed both tracker outcomes; the supplied browser traps stay within one process and cannot catch this. Scope every ordered observation to the server generation, or reconcile unscoped announcements without letting their counter seed the current generation. Add a browser lifecycle test across a backend restart in both orderings, asserting that protected cached views unmount on lock and current verification reopens them.

3. **P2 · Scope adherence — the unverified-bucket status does not clear automatically in an already-open, visible app.** [App.tsx:401-405](https://github.com/huggingface/agent-manager/blob/db1dd79cacd24e76ece3736ecb0d111ed0bd8eb9/web/src/App.tsx#L401) polls safe status only while locked. The supported bucket-warning state is explicitly unlocked, so when credential recovery lets the server verify the bucket, there is no lock/unlock render transition and no status fetch. Settings can keep showing the old warning indefinitely until a reload, tab return, or network event. That misses the issue goal that accurate status clears itself after successful verification and the requirement to preserve this warning policy. Give unlocked tabs a bounded way to refresh this safe status without introducing Hub work per tab, and test warning-only → fully verified while the tab remains visible.

I found no regression in the six round-one fixes: the focused policy suite passed 126/126, the real HTTP/WebSocket suite 77/77, and the full same-process Chromium lifecycle passed. Web tests and production build passed. The aggregate server run stopped only at the documented main-red cron case; the remaining 23 discovered non-manual suites passed when continued separately. No branch changes, merge, deployment, or live-service probes were performed.

lvwerra:
Round-two fixes pushed. All three findings were real; each is fixed with a test that fails when the fix is reverted (mutation results at the end).

**1. P1 — cancellation now lives as long as the work, not the response.** New `lockScope()` in `index.js`: any PTY work that outlives (or never had) an HTTP response registers its own revoke callback in the same registry the lock walks, and releases it only when the work ends. `deliver()` holds one for its whole run — so a cron run's background delivery (202 already sent) and a quickstart prompt typed after the CLI's boot delay are cancelled exactly like an awaited input — and checks it before every effect (readiness race, each prelude command, the prompt). Attachment insertion holds one across its per-image commands and checks it before each `sendInput` and the final paste. A scope opened while already locked is aborted immediately. Tests in `visibility-lock.test.mjs`: four kinds of detached work are fired back to back right before the Space turns public — an input to a stopped slow-to-ready agent, a cron `run` (202), a quickstart on a CLI without a prompt flag, and a five-image insertion — and after the lock none of the prompts is ever typed, the insertion stops at the command it had reached (asserted before and after a 4.5 s window), the agents keep running, and nothing is replayed after reopening. Mutations: caller-signal-only `deliver`, insertion without checks, quickstart without the check — each fails its check.

**2. P1 — counters are compared only within one server generation.** The `403 {"error":"locked"}` body and the `4003` close reason (`locked:<reason>:<seq>:<boot>`) now carry the process id alongside the counter, and the tracker's `observeLocked` takes both. A counter without a generation is treated as seq-less; a refusal from an unknown or different generation returns `refetch`: it is neither applied to the UI nor allowed to seed the counters — the app fetches `/api/info`, whose `boot` decides (a new generation resets the counting; a stale old-generation refusal changes nothing). So a restarted server's first refusal (seq 1) locks an open tab, and a delayed refusal from the previous generation (seq 999) cannot wedge it. Tests: `lockStatus.test.mjs` covers both orderings plus the seq-without-boot and first-observation cases; `visibility-ui.test.mjs` now restarts the backend under the open app on the same port in both orderings — the new generation's first refusal locks the tab and unmounts terminal/reader/composer, its verification reopens it, then an injected old-generation refusal with a high counter neither locks nor wedges it, and a real lock still lands afterwards. Mutation (cross-generation comparison restored) fails the unit suite and the browser restart case.

**3. P2 — the unverified-bucket warning clears in an open, visible tab.** Two parts. Server: a refused credential is re-tried every `UNAUTHORIZED_RETRY_CYCLES` (10 cycles ≈ 10 min); while still refused the cached exemption stands (no re-count, no lock), and once the Hub accepts it the bucket is verified for real and the warning drops. Without this the cached refusal never expired within a process, so the scenario was unreachable except across a restart. Web: while unlocked and visible, `/api/info` is re-read every 30 s (cached server state; no Hub work per tab), in addition to the existing load/return/online reads and the 15 s locked poll. Tests: unit (no re-try before the window, one re-try per window, still-refused keeps the exemption, fixed access verifies and clears); browser (server in warning-only mode, Settings shows the warning, fixture token accepted again, warning disappears in the same tab with no reload and no tab switch). Mutations: re-try disabled fails the unit suite; unlocked poll disabled fails the browser case.

Docs updated accordingly (`docs/privacy-lock.md`: lock scopes and what they cover, generation-scoped ordering, credential re-try, the unlocked status cadence).

Verification on the branch: unit 134/134; integration 90/90; web unit green; browser 39/39 on the final bundle; typecheck and build clean; server suites individually green except the known-red `crons` #4; web `npm test` green. Fixture identifiers are made up; the fixture token never appears in server output (asserted).

PR #148, opened 2026-09-09
Load Files, Trace and Settings panels on demand, measured (#133)

### LINKED ISSUE #133: Speed up app startup with measured on-demand panel loading
## Goal

Make a cold app visit usable sooner by keeping infrequently used panels out of the initial JavaScript work. Measure the improvement in the production build, while preserving fast access to the user's actual restored view, drafts, session identity and Reader/terminal behavior.

The operator approved this as a separate PR-sized performance task. This is frontend startup work, not the deferred build/CI project, a Reader rewrite or a change to coding-tool updates.

## Starting points

Planning inspected main at `ef08e843d1c67fcc4e1c463415819ea351734d6d`; recheck before implementing. `App.tsx` imports SettingsView, FilesPane and TracePane eagerly. SettingsView in turn imports UsagePanel, SkillsEditor, CronSettings and ApiLog. FilesPane imports the TraceView implementation as well as lightweight types from TracePane. CodeMirror core/languages and PDF loading already have dynamic imports; preserve those boundaries rather than claiming the app has no code splitting.

These are investigation targets, not a requirement to split every named component. Inspect the actual production module graph and measure baseline startup before choosing boundaries. Earlier review bundle-size figures predate the Reader redesign and must not be presented as current measurements.

## Requirements

| Area | Required behavior |
| --- | --- |
| Useful startup measurement | Measure cold Overview startup and startup restoring a Reader session, plus direct/restored Files or Trace selection where supported. Report initial requested/compressed JavaScript, parse/evaluation work, time to usable navigation and time to the selected view's first useful interaction. Use the same browser, fixture data, device/network settings and production build method before/after. A smaller filename/chunk alone is not proof of faster startup. |
| Deliberate loading boundaries | Defer substantial, infrequently used panel implementations and, where useful, Settings subpages until needed. Keep lightweight navigation, shared types/data helpers and essential shell controls available without importing the panel implementation. Check transitive/static imports and bundler preloads so a new dynamic import does not merely disguise an unchanged initial dependency graph. Extract only the shared seams needed for these boundaries; no general component rewrite. |
| Prioritize the selected view | A restored active panel must start loading promptly rather than waiting behind an arbitrary timer, hover or unrelated background work. Core Reader/composer readiness, sidebar navigation and lock/setup status must not wait for unopened Settings/Usage/Skills/API-log code. Do not push all existing startup work into a new serial request waterfall. Preserve independent progress in multi-pane layouts. |
| Local loading and failure states | Place loading/error handling at the selected panel boundary so one slow/failed chunk does not blank the whole application or strand the user behind an invisible Settings overlay. Keep Close/Back or another way out reachable. Reserve the intended panel geometry, use existing visual/focus styles, and avoid unstyled content or misleading empty states. A module failure must settle visibly; closing or changing selection while loading must win over a late completion. |
| Realistic module retry | Handle failed imports and stale deployment chunk URLs honestly. Do not label resetting a React error boundary as Retry unless it actually retries a recoverable load. React caches lazy-load promises; its guidance also requires stable component declarations outside renders. Use a bounded, tested retry mechanism where supported; when a newer deployment genuinely requires page reload, explain that and offer an explicit user action instead of a reload loop. Never automatically discard unsent work. See the official [React lazy reference](https://react.dev/reference/react/lazy). |
| Preserve state and ownership | Changing chunk boundaries must not change mount/identity semantics unnecessarily. Preserve panel selection, file/editor drafts and save status, Reader anchors/follow-latest intent, pending attachments and session isolation through open/close/switch operations. Keep existing save controllers alive as required by #123. Do not mount hidden panels just to preload code or let module completion reopen a closed panel. |
| Preserve terminal separation | No hidden terminal connection, input, ownership claim, resize, process start or restart merely to improve Reader startup or prefetch a panel. Opening/closing Settings must not unexpectedly rebuild terminal viewers or lose canonical restore/ownership state. Preserve the Reader-only no-PTY behavior from #119 and the actual mode-switch requirements in #127. |
| Restrained preloading | Load-on-demand is the default for unused panels. Optional intent-driven preloading is acceptable if measurements justify it and it does not fetch every panel immediately after startup. Bound/deduplicate work; fetching code must not mount a view, start its API polling, send prompts or acknowledge unread output. Repeated visits should reuse already loaded code without adding an unbounded retained-view cache. |
| Compatibility and scope | Keep the existing React/Vite stack, feature behavior, sanitization, embedding, theme/zoom and storage-denied fallback. Preserve existing editor/PDF splitting and shared renderer behavior. No router migration, service-worker/cache redesign, framework upgrade, API rewrite, new persistence system or Docker/CI changes. Correctness must not depend on eager availability of a component that is now deferred. |

## Required tests and performance evidence

| --- | --- |
| Production dependency graph | With cold browser cache, unopened panel-specific code is absent from the initial requests/work; first activation fetches the needed code and repeat activation reuses it. Inspect actual built chunks/network entries, not only mocked import calls. Shared code required by the active view is an explicitly documented exception. |
| Cold and restored visits | Overview, restored Reader, Files/Trace selection, a group with mixed panes, first Settings opening and each deferred Settings subpage. Measure initial readiness and first panel-open latency separately so work is not simply shifted into a worse interaction. Include empty/new sessions with a usable composer and locked/setup mode. |
| Loading races and errors | Delayed chunk, failed chunk, supported retry, missing old deployment chunk, offline first open, close before completion, rapid panel/subpage changes, and a failure in one pane while another remains usable. No stuck overlay, blank whole app, stale selection restoration, reload loop or repeated side effects. |
| State continuity | Unsaved file text, immediate settings saves, pending save/error state, prompt drafts/attachments, Reader manually scrolled back and following Latest, and actual Reader/Terminal/Settings round trips. Preserve the right state per session even with late module loads. Test current behavior and coordinate the approved #123/#127 changes rather than implementing a second controller. |
| Browser and layout | Desktop, narrow/mobile viewport, zoom, keyboard navigation/Back, light/dark themes, denied browser storage and supported iframe presentation. Loading/failure states have a usable exit and stable layout; no terminal bleed or unintended focus loss. Use harmless local fixtures, not live sessions. |
| Side effects and resources | Count API polls, module requests, WebSocket attaches/input/claims/resizes and retained listeners after repeated navigation. Optional code prefetch performs no feature action, and inactive deferred panels do not start background work. Unread tracking is not consumed by preloading. |

- Provide before/after production measurements for a normal desktop and a reproducibly constrained mobile-like profile, using repeated samples and documenting cache/data/network conditions. Show meaningful initial JavaScript/work reduction and improved or non-regressed first usable view; report first-open and warm-open costs for deferred panels. Set practical regression budgets from those measurements, without claiming universal millisecond targets from one machine.
- Add normally discovered regression tests. Run the relevant app/settings/files/Reader/terminal browser suites, normal web tests, frontend typecheck and production build. Run affected server integration suites if the test harness changes; report exact commands and unrelated baseline failures separately. This task does not authorize adding the deferred CI system.
- Check overlap with file-preview [PR #120](https://github.com/huggingface/agent-manager/pull/120), saves [#123](https://github.com/huggingface/agent-manager/issues/123), return/reconnect [#124](https://github.com/huggingface/agent-manager/issues/124), mode switching [#127](https://github.com/huggingface/agent-manager/issues/127), unread [#128](https://github.com/huggingface/agent-manager/issues/128), initial Reader history [#129](https://github.com/huggingface/agent-manager/issues/129), and lock transitions [#131](https://github.com/huggingface/agent-manager/issues/131). Their scopes remain unchanged; app startup optimization must not delay the Reader's required first stable viewport or create duplicate refresh loops.

Open one focused PR with the chosen loading boundaries, measurements and tests. No live prompts/stops/restarts, notifications, privacy changes, paid jobs or deployment. Do not merge without separate authorization.

lvwerra:
Closes #133

## What this does

A cold visit no longer parses the code of panels most visits never open. `FilesPane`, `TracePane` and `SettingsView` are fetched the first time they are shown, and each Settings subpage (`UsagePanel`, `ApiLog`, `SkillsEditor`, `CronSettings`) is its own chunk behind its tab. The terminal/Reader pane, the sidebar, the Overview and the lock screen stay in the entry.

**Boundaries** (`web/src/lib/lazyModule.ts` + `web/src/components/LazyPanel.tsx`)
- One module-level loader per panel keyed into a small cache: one request however many tiles show the panel, and every later show reuses the module. No hidden mounting, no prefetch: nothing is fetched until a panel is actually rendered, and fetching never mounts a view or starts its polling.
- The boundary reserves the panel's geometry (tile / settings page / file viewer), keeps the panel's own Close reachable, and settles visibly on failure. Open/closed stays the parent's state, so a chunk landing after Close cannot reopen anything.
- **Honest retry.** Chromium caches a *failed* module fetch in its module map — verified here: after one aborted fetch, a second `import()` of the same URL rejects without touching the network. So a plain re-import is not a retry. The loader takes the chunk URL from the browser's error (Chromium/Firefox name it; WebKit does not), makes one automatic attempt under `?am-retry=N` (400 ms later), and offers **Try again** only when that mechanism is available. When the page's own HTML, re-read with `cache: no-store`, names a different entry script, the panel says a newer version was deployed and offers an explicit **Reload** — it never reloads by itself and never offers a retry that cannot work.
- Settings' Back/title/tabs move into an eager `SettingsShell`, so the operator is never behind an invisible overlay while the page body loads (`.app-suspended` hides the fleet as before) and can already pick a tab.
- `FilesPane` now takes its trace types from `lib/traceWindows` and loads `TraceView` on demand only when a `.jsonl` file is opened, so opening Files no longer pulls the trace renderer and a failure in one does not take the other.

**Kept as-is:** the CodeMirror and PDF dynamic imports; `RemotePane`, `Locked`, `TraceInfo` (the Reader's info popover) stay in the entry — the first two are small and the last is part of the active pane. xterm/react-dom/marked/dompurify remain in the entry because the terminal and the Reader need them.

## Measurements (production build, this machine)

Evidence page with all cells, ranges and conditions: https://lvwerra-agent-artifacts.static.hf.space/issue-133-startup.html

Harness: `node scripts/startup-bench.mjs --dist <dist> --label before|after --runs 5` — a real backend over a fixture fleet (`web/test/helpers/fixture-server.mjs`: two Claude sessions, one with a 14-turn transcript, a Files pane, a Trace pane, a group with both), Reader mode only, no PTY ever attached. Fresh browser context per visit, HTTP cache disabled; markers stamped in-page by a MutationObserver. Desktop 1280×800 unthrottled; mobile-like 390×844 with 4× CPU slowdown and Chrome's "Slow 4G" (1.6 Mbps / 150 ms). Medians of 5.

**Desktop, unthrottled** — cold visit, medians of 5 (range)

| scenario | JS on the wire | gzip | V8 script ms | sidebar usable ms | view usable ms |
|---|---|---|---|---|---|
| Overview (cold) | 795 kB → 697 kB | 234 kB → 206 kB | 83 → 57 | 167 → 128 | 167 (126–205) → 128 (113–153) |
| Restored Reader | 795 kB → 697 kB | 234 kB → 206 kB | 110 → 109 | 94 → 84 | 209 (195–277) → 200 (196–354) |
| Restored empty session | 795 kB → 697 kB | 234 kB → 206 kB | 97 → 66 | 110 → 80 | 224 (194–274) → 154 (151–163) |
| Restored Files | 795 kB → 729 kB | 234 kB → 216 kB | 94 → 76 | 105 → 81 | 225 (186–274) → 164 (159–176) |
| Restored Trace | 795 kB → 709 kB | 234 kB → 210 kB | 188 → 108 | 146 → 79 | 387 (277–443) → 188 (181–291) |
| Restored group (Files+Trace) | 795 kB → 741 kB | 234 kB → 221 kB | 178 → 136 | 146 → 96 | 336 (214–486) → 235 (215–308) |

desktop: first open of a deferred panel from a warm Overview, before → after (ms), at first *useful* content (Usage: a populated trace row; API log: a real row)

| panel | first open | reopen |
|---|---|---|
| settings | 108 → 116 | 35 → 35 |
| usage | 68 → 75 | 73 → 74 |
| apilog | 62 → 70 | 67 → 64 |
| skills | 41 → 58 | 41 → 42 |
| cron | 49 → 67 | 38 → 38 |
| files | 68 → 57 | 56 → 54 |
| trace | 99 → 119 | 63 → 62 |

**Mobile-like (4× CPU, 1.6 Mbps / 150 ms)** — cold visit, medians of 5 (range)

| scenario | JS on the wire | gzip | V8 script ms | sidebar usable ms | view usable ms |
|---|---|---|---|---|---|
| Overview (cold) | 795 kB → 697 kB | 234 kB → 206 kB | 249 → 257 | 5250 → 4760 | 5250 (5240–5260) → 4760 (4728–4760) |
| Restored Reader | 795 kB → 697 kB | 234 kB → 206 kB | 482 → 492 | 5244 → 4733 | 6029 (6014–6060) → 5515 (5500–5573) |
| Restored empty session | 795 kB → 697 kB | 234 kB → 206 kB | 324 → 304 | 5244 → 4732 | 5646 (5623–5671) → 5121 (5103–5161) |
| Restored Files | 795 kB → 729 kB | 234 kB → 216 kB | 365 → 356 | 5250 → 4736 | 5905 (5889–5917) → 5739 (5736–5774) |
| Restored Trace | 795 kB → 709 kB | 234 kB → 210 kB | 690 → 622 | 5257 → 4734 | 6014 (6006–6037) → 5667 (5663–5692) |
| Restored group (Files+Trace) | 795 kB → 729 kB | 234 kB → 216 kB | 355 → 354 | 5254 → 4741 | 5907 (5876–5947) → 5750 (5723–5781) |

mobile: first open of a deferred panel from a warm Overview, before → after (ms), at first *useful* content (Usage: a populated trace row; API log: a real row)

| panel | first open | reopen |
|---|---|---|
| settings | 262 → 948 | 84 → 83 |
| usage | 916 → 875 | 865 → 865 |
| apilog | 314 → 838 | 332 → 332 |
| skills | 53 → 224 | 46 → 46 |
| cron | 127 → 327 | 92 → 94 |

Entry chunk 795 kB → 697 kB minified (gzip 234 kB → 206 kB), −12 %. New on-demand chunks (minified): FilesPane 31 kB, SettingsView 24 kB, ApiLog 15 kB, CronSettings 12 kB, TracePane 12 kB, UsagePanel 5.6 kB, SkillsEditor 3.6 kB.

Reading it: on the throttled profile every view that needs no panel is usable ~0.5 s sooner (−9 to −10 %), and the restored Files/Trace/group views 0.15–0.35 s sooner even though their chunk now follows the entry (ranges are 20–60 ms wide, so these are real). Desktop cells are noise around zero — see the reviewer's reproduction below. The cost moved to first open is visible on the throttled profile where the panel's own data is quick: Settings 262 → 948 ms, API log 314 → 838 ms, Skills 53 → 224, Cron 127 → 327 (the chunk shares the slow link with the Overview's polls). Usage is ~0.9 s cold on **both** builds (916 → 875) because its first useful row waits on its own seven usage/trace requests — the chunk is not what the operator waits for there. Desktop first opens are 0–20 ms slower. Reopen is unchanged everywhere (module cached).

Conditions: shared box, 8 cores available to the container; the harness records the 1-minute load average at the start and after each profile and a run is kept only if it stayed under ~10. Cold-visit cells: before 6.3 → 6.8 → 9.5, after desktop 7.5 → 8.2 and mobile 6.4 → 6.35 (a first mobile pass that ended at load 18.1 was discarded, as was an earlier full pair taken at load 11–20). Panel-open cells (re-measured after the review's marker fix): before 5.1 → 6.6, after 6.5 → 6.4.

Regression budgets from these runs: entry chunk ≤ 720 kB minified and free of the seven panel modules (both asserted by `web/test/lazyPanels.test.mjs` on the built chunks); zero panel chunks on a cold Overview or restored Reader visit (asserted); first open of a deferred panel within ~1.2× the **after** medians above (mobile-like: Settings ≲ 1.15 s, API log ≲ 1.05 s, other subpages ≲ 0.45 s; desktop ≲ 150 ms) — that first-open cost is the accepted trade, and if it must come down the boundary needs tuning (intent prefetch on the Settings button, or one chunk for Settings + API log), not a bigger entry; warm reopen equal to before (cache intact).

**Read the gain as less to download and wait for at startup, not a CPU speedup.** The reviewer's independent rerun reproduced the mobile-like numbers (Overview 5246 → 4752 ms, Reader 6028 → 5535) and the first-open trade, and found desktop flat (Overview 88 → 81 ms, Reader 195 → 195, mixed group 195 → 220) — so the desktop cells in my run are noise around zero, and V8 script time is broadly flat for Overview/Reader in both runs.

## Tests

- `web/test/lazyModule.test.mjs` (node, no browser): dedupe, bounded automatic retry under a busted URL, manual retry only where it can work, stale deployment read from the page HTML, unknown when offline. Mutation-checked: reintroducing the retry-URL bug this caught (`?am-retry=2&am-retry=3`) fails it.
- `web/test/lazyPanels.test.mjs` (Chromium, production build, real server, port 7904): cold Overview requests no panel chunk and the entry contains no panel-only literal; unopened panels do no API polling; first Settings open + each subpage fetch once and reopen from cache, closing stops their polling; Files fetches Files only, Trace its own; restored Files / Reader / group; a failed Files chunk fails in its tile while the Trace tile, sidebar and Close work and Try again really loads it (asserting the `?am-retry=` URLs on the wire); a chunk removed by a newer deployment shows Reload, no Try again, and no self-navigation; offline first open then Try again online; Settings closed while loading stays closed when the code lands and the shell's tabs work meanwhile; two in-flight subpage switches — the last choice wins; phone width shell with Back during load; denied `localStorage`.
- `web/test/fixtureServer.test.mjs` (review fix): the fixture server refuses a port that is not free and only calls itself ready when `/api/info` names its own `DATA_DIR`; the test holds a stub on the port and requires a rejection with zero requests to it.
- `web/test/helpers/markers.mjs` (review fix): one definition of "usable" shared by the benchmark and the browser suite. Usage counts a populated trace row and the API log a real row, not the frame or its skeletons; the suite holds the Usage data back and requires the marker to stay silent while skeletons show. The Usage/API-log first-open cells below were re-measured with that marker.
- `web/test/settingsMobile.test.mjs` and `cronSettings.test.mjs` adapted: they mount `SettingsShell` around `SettingsView`, as App does.

Commands run and results: 
- `cd web && npx tsc --noEmit` — ok
- `cd web && npx vite build` — ok (chunk list above)
- `cd web && npm test` — 25 suites passed (3 render suites skipped as manual by the runner); includes the two new suites
- `cd server && npm test` — stops at `test/crons.test.mjs` (#4, known-red on main); every other `server/test/*.test.mjs` suite run individually via `node ../scripts/run-suites.mjs test/<name>.test.mjs` — all pass
- `node scripts/startup-bench.mjs` on both builds as described above

## Deliberately left out
- No intent-driven prefetch (hover/pointerdown): measurements did not justify it — see the first-open table; the restored Files/Trace case starts its chunk as soon as the tree names the pane, without a timer.
- No change to what the Reader/terminal pane imports (xterm stays in the entry), no Reader/composer changes, no router, service worker or Vite config changes, no CI.
- `RemotePane` not deferred (small; not named in the issue).
- Locked/setup mode: unchanged code path (`Locked` is in the entry); not separately benchmarked.

Known-red on main, not touched: `server/test/crons.test.mjs` #4.

### DIFF
--- a/scripts/startup-bench.mjs
+++ b/scripts/startup-bench.mjs
@@ -30,6 +30,7 @@ import os from 'node:os';
 import { chromium } from '../web/node_modules/playwright/index.mjs';
 import { chromiumLaunchOptions } from './test-chromium.mjs';
 import { startFixtureServer } from '../web/test/helpers/fixture-server.mjs';
+import { MARK } from '../web/test/helpers/markers.mjs';
 
 const arg = (k, d) => { const i = process.argv.indexOf(`--${k}`); return i > 0 ? process.argv[i + 1] : d; };
 const DIST = path.resolve(arg('dist'));
@@ -48,17 +49,7 @@ const server = await startFixtureServer({ port: PORT, publicDir: DIST, tag: 'am-
 const { ids } = server;
 const browser = await chromium.launch(chromiumLaunchOptions());
 
-// The markers: a view is "usable" when its first interactive content is in the DOM.
-const MARK = {
-  nav: '.sidebar .row.session',
-  overview: '.ov-tile, .ov-card',
-  reader: '.pane-reader .cx-prompt',
-  composer: '.pane-reader .ov-composer textarea:not([disabled])',
-  files: '.files-body.tree .tree-row',
-  trace: '.trace-body .cx-prompt, .trace-body .tv-md',
-  settings: '.settings-page .setting-row',
-  usage: '.usage', apilog: '.al-tbl, .al-head', skills: '.skills', cron: '.cron-form',
-};
+// A view is usable at its first USEFUL content — see web/test/helpers/markers.mjs.
 const SCENARIOS = [
   { id: 'overview', restore: 'overview', view: ['overview'] },
   { id: 'reader', restore: `s:${ids.cadence}`, view: ['reader', 'composer'] },
--- a/scripts/startup-bench.mjs
+++ b/scripts/startup-bench.mjs
@@ -155,11 +155,18 @@ async function panelOpens(page, mobile) {
 }
 
 const gz = (f) => zlib.gzipSync(fs.readFileSync(f), { level: 6 }).length;
-const results = { label: LABEL, dist: DIST, at: new Date().toISOString(), runs: RUNS, node: process.version, chromium: browser.version(), mobile: MOBILE, profiles: {} };
+// Resumable: a partial file from an interrupted run of the same build is
+// continued, cell by cell, rather than started over.
+const prior = fs.existsSync(OUT) ? JSON.parse(fs.readFileSync(OUT, 'utf8')) : null;
+const results = prior && prior.dist === DIST && prior.runs === RUNS
+  ? { ...prior, resumedAt: new Date().toISOString() }
+  : { label: LABEL, dist: DIST, at: new Date().toISOString(), runs: RUNS, node: process.version, chromium: browser.version(), mobile: MOBILE, profiles: {} };
+const save = () => fs.writeFileSync(OUT, JSON.stringify(results, null, 1));
 try {
   for (const profile of PROFILES) {
-    const P = (results.profiles[profile] = { scenarios: {}, panels: null });
+    const P = (results.profiles[profile] ||= { scenarios: {}, panels: null });
     for (const sc of SCENARIOS) {
+      if (P.scenarios[sc.id]) continue;
       const samples = [];
       for (let i = 0; i < RUNS; i++) {
         const v = await visit(profile, sc);
@@ -173,7 +180,9 @@ try {
         script: stat(samples.map((s) => s.script)), compile: stat(samples.map((s) => s.compile)), task: stat(samples.map((s) => s.task)),
         js: { count: samples[0].js.count, bytes: samples[0].js.bytes, gzip: files.reduce((n, f) => n + gz(path.join(DIST, 'assets', f)), 0), files },
       };
+      save();
     }
+    if (P.panels) continue;
     console.log();
     const opens = [];
     for (let i = 0; i < RUNS; i++) {
@@ -188,14 +197,14 @@ try {
       P.panels[pass] = {};
       for (const k of Object.keys(opens[0][pass])) P.panels[pass][k] = stat(opens.map((o) => o[pass][k]));
     }
-    fs.writeFileSync(OUT, JSON.stringify(results, null, 1)); // a later crash keeps this profile
+    save();
   }
   results.chunks = Object.fromEntries(fs.readdirSync(path.join(DIST, 'assets')).filter((f) => f.endsWith('.js')).map((f) => [f, { bytes: fs.statSync(path.join(DIST, 'assets', f)).size, gzip: gz(path.join(DIST, 'assets', f)) }]));
 } finally {
   await browser.close();
   await server.stop();
 }
-fs.writeFileSync(OUT, JSON.stringify(results, null, 1));
+save();
 
 const kb = (n) => `${(n / 1024).toFixed(0)}k`;
 const ms = (s) => `${s.median.toFixed(0)}ms`;
--- a/scripts/startup-bench.mjs
+++ b/scripts/startup-bench.mjs
@@ -1,0 +1,192 @@
+// Startup cost of a production build, measured in a real Chromium against the
+// fixture fleet (web/test/helpers/fixture-server.mjs), before and after a change
+// to how the app loads. Every number here is a measurement of THIS machine under
+// the stated conditions — compare two runs of this script, never a run to a
+// figure from somewhere else.
+//
+// For each profile × scenario, `--runs` cold visits (fresh browser profile,
+// HTTP cache disabled, nothing preloaded) record:
+//   js.bytes / js.count   script bytes and requests until the view was usable
+//   script / compile      V8 ScriptDuration and V8CompileDuration (ms) at that point
+//   nav                   ms from navigation start until the sidebar lists sessions
+//   view                  ms until the selected view's first useful interaction:
+//                         Overview cards, the Reader's first exchange + composer,
+//                         the Files listing, the Trace's first turn, both group tiles
+// then, from a warm Overview, the cost of opening each deferred panel the first
+// time (its chunk on the wire) and again (from the module cache): Settings and
+// each subpage, Files, Trace.
+//
+// Profiles: desktop (1280×800, no throttling) and mobile-like (390×844, 4× CPU
+// slowdown, 1.6 Mbps down / 750 kbps up / 150 ms RTT — Chrome's "Slow 4G").
+// The server does not compress; gzip sizes are computed from the files.
+//
+//   node scripts/startup-bench.mjs --dist <web/dist> --label before --runs 5 --out before.json
+//
+// am-test: manual — a benchmark, not a pass/fail suite.
+import fs from 'node:fs';
+import path from 'node:path';
+import zlib from 'node:zlib';
+import { chromium } from '../web/node_modules/playwright/index.mjs';
+import { chromiumLaunchOptions } from './test-chromium.mjs';
+import { startFixtureServer } from '../web/test/helpers/fixture-server.mjs';
+
+const arg = (k, d) => { const i = process.argv.indexOf(`--${k}`); return i > 0 ? process.argv[i + 1] : d; };
+const DIST = path.resolve(arg('dist'));
+const LABEL = arg('label', path.basename(DIST));
+const RUNS = Number(arg('runs', 5));
+const OUT = arg('out', `startup-${LABEL}.json`);
+const PORT = Number(arg('port', 7905));
+const PROFILES = arg('profiles', 'desktop,mobile').split(',');
+
+const MOBILE = { cpu: 4, latency: 150, down: 1.6e6 / 8, up: 750e3 / 8 };
+const med = (a) => { const s = [...a].sort((x, y) => x - y); return s.length ? (s.length % 2 ? s[(s.length - 1) / 2] : (s[s.length / 2 - 1] + s[s.length / 2]) / 2) : NaN; };
+const stat = (a) => ({ median: med(a), min: Math.min(...a), max: Math.max(...a), n: a.length });
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+const server = await startFixtureServer({ port: PORT, publicDir: DIST, tag: 'am-bench-' });
+const { ids } = server;
+const browser = await chromium.launch(chromiumLaunchOptions());
+
+// The markers: a view is "usable" when its first interactive content is in the DOM.
+const MARK = {
+  nav: '.sidebar .row.session',
+  overview: '.ov-tile, .ov-card',
+  reader: '.pane-reader .cx-prompt',
+  composer: '.pane-reader .ov-composer textarea:not([disabled])',
+  files: '.files-body.tree .tree-row',
+  trace: '.trace-body .cx-prompt, .trace-body .tv-md',
+  settings: '.settings-page .setting-row',
+  usage: '.usage', apilog: '.al-tbl, .al-head', skills: '.skills', cron: '.cron-form',
+};
+const SCENARIOS = [
+  { id: 'overview', restore: 'overview', view: ['overview'] },
+  { id: 'reader', restore: `s:${ids.cadence}`, view: ['reader', 'composer'] },
+  { id: 'reader-empty', restore: `s:${ids.fresh}`, view: ['composer'] },
+  { id: 'files', restore: `s:${ids.files}`, view: ['files'] },
+  { id: 'trace', restore: `s:${ids.trace}`, view: ['trace'] },
+  { id: 'group', restore: `g:${ids.group}`, focused: ids.groupFiles, view: ['files', 'trace'] },
+];
+
+async function visit(profile, scenario) {
+  const mobile = profile === 'mobile';
+  const ctx = await browser.newContext({ viewport: mobile ? { width: 390, height: 844 } : { width: 1280, height: 800 } });
+  const page = await ctx.newPage();
+  const cdp = await ctx.newCDPSession(page);
+  await cdp.send('Network.enable');
+  await cdp.send('Network.setCacheDisabled', { cacheDisabled: true });
+  await cdp.send('Performance.enable');
+  if (mobile) {
+    await cdp.send('Emulation.setCPUThrottlingRate', { rate: MOBILE.cpu });
+    await cdp.send('Network.emulateNetworkConditions', { offline: false, latency: MOBILE.latency, downloadThroughput: MOBILE.down, uploadThroughput: MOBILE.up });
+  }
+  const js = new Map(); // requestId -> {url, bytes}
+  cdp.on('Network.requestWillBeSent', (e) => { if (e.type === 'Script') js.set(e.requestId, { url: e.request.url, bytes: 0, t: null }); });
+  cdp.on('Network.loadingFinished', (e) => { const r = js.get(e.requestId); if (r) { r.bytes = e.encodedDataLength; r.t = e.timestamp; } });
+  await page.addInitScript(({ restore, focused, mobile, marks }) => {
+    localStorage.setItem('am-active-ref', restore);
+    if (focused) localStorage.setItem('am-focused-id', focused);
+    localStorage.setItem('am-pane-mode', 'reader');
+    if (mobile) localStorage.setItem('am-mobile-stage', '1');
+    // First appearance of each marker, stamped by a MutationObserver rather than
+    // by polling from outside, so the numbers are the DOM's, not the harness's.
+    const seen = (window.__marks = {});
+    const look = () => { for (const [k, sel] of Object.entries(marks)) if (!(k in seen) && document.querySelector(sel)) seen[k] = performance.now(); };
+    // `document` itself: init scripts run before <html> exists.
+    new MutationObserver(look).observe(document, { childList: true, subtree: true, attributes: true });
+    look();
+  }, { restore: scenario.restore, focused: scenario.focused || null, mobile, marks: MARK });
+
+  await page.goto(server.origin + '/');
+  const want = ['nav', ...scenario.view];
+  await page.waitForFunction((want) => want.every((k) => k in window.__marks), want, { timeout: 60_000 });
+  const marks = await page.evaluate(() => window.__marks);
+  const perf = Object.fromEntries((await cdp.send('Performance.getMetrics')).metrics.map((m) => [m.name, m.value]));
+  const startup = {
+    nav: marks.nav,
+    view: Math.max(...scenario.view.map((k) => marks[k])),
+    js: { count: js.size, bytes: [...js.values()].reduce((n, r) => n + r.bytes, 0), files: [...js.values()].map((r) => path.basename(new URL(r.url).pathname)) },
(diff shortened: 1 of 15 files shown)


lvwerra:
Reviewed `adb843751e716b3bc093871dc299efc563326842` against `ef08e843d1c67fcc4e1c463415819ea351734d6d`, with both production builds run locally. Two fixes needed in the new measurement/test tooling; I did not find a new defect in the production loading boundaries.

1. **P2 — Verify the fixture server owns its port before seeding it.** [fixture-server.mjs:52–69](https://github.com/huggingface/agent-manager/blob/adb843751e716b3bc093871dc299efc563326842/web/test/helpers/fixture-server.mjs#L52-L69) accepts any successful `/api/health` response without checking that the spawned child actually started. With the port already occupied, the child reports `EADDRINUSE`, but the helper still sends its fixture writes to the existing listener and returns successfully. I confirmed this with a disposable occupied-port HTTP stub: it received ten non-GET requests and the helper returned its IDs despite the child bind failure. This can mutate another test instance and produce evidence from the wrong build. Fail before seeding unless readiness belongs to this child; do not reuse or terminate an unrelated listener. Add an occupied-port regression test asserting rejection and **zero fixture writes** to that listener.

2. **P2 — The Usage first-useful-content marker measures skeletons, not usable usage data.** [startup-bench.mjs:60](https://github.com/huggingface/agent-manager/blob/adb843751e716b3bc093871dc299efc563326842/scripts/startup-bench.mjs#L60) stops at `.usage`, which exists before any of the panel's data has arrived. On this production build, I held all seven Usage/Traces requests: the marker completed in 70 ms, with 21 skeleton elements and no populated trace rows still present a second later. That does not support the evidence page's statement that these timings include each panel's API round trips. Keep a separate frame-painted metric if useful, but measure the first useful fixture data for Usage and rerun its cold/warm cells. Add a delayed-data check that cannot report useful readiness while only skeletons are present. This does **not** invalidate the independently reproduced Overview/Reader startup gain below.

Non-blocking documentation correction: the proposed mobile first-open budget of approximately 1.5× **before** already fails the PR's own Settings/API-log/Skills numbers. Base an achievable budget on the accepted measured tradeoff, or explicitly say the boundary needs further tuning. Also, the entry-size limit is described as enforced by `lazyPanels.test.mjs`, but that test checks module literals/requests, not a 720 KiB size limit. Correct that claim or add the assertion.

### Independent performance check

Same Chromium `153.0.8010.12`, production `npm run build` on both SHAs, identical dependencies and fixture fleet, five fresh contexts per cell, HTTP cache disabled. Same desktop and mobile settings as the PR: mobile 390×844, 4× CPU slowdown, 1.6 Mbps down/750 kbps up, 150 ms latency. Like the submitted benchmark, the local server is uncompressed; gzip figures are computed sizes, not measured compressed transfer. No concurrent builds/suites during measurement. Ran desktop before→after, then mobile after→before. Desktop load averages at start/end were 6.59→7.24 and 7.34→7.72; mobile after 6.90→6.66, before 6.30→5.24, on the 8-CPU quota. My intermediate load checks stayed below 8.

Mobile selected-view readiness, median ms (min–max):

| View | Before | After |
| --- | ---: | ---: |
| Overview | 5246 (5232–5268) | 4752 (4737–4787) |
| Restored Reader | 6028 (6012–6053) | 5535 (5510–5632) |
| Empty Reader/composer | 5608 (5595–5615) | 5135 (5096–5145) |
| Restored Files | 5880 (5855–5915) | 5743 (5736–5748) |
| Restored Trace | 5995 (5981–5998) | 5662 (5658–5708) |
| Mixed group, focused Files on mobile | 5886 (5864–5896) | 5758 (5725–5775) |

The mobile improvement is reproducible and outside these sample ranges. Entry transfer is again approximately 795→697 KiB, computed gzip 234→206 KiB; the seven deferred panel markers are present in the baseline entry and absent in the PR entry. CPU script time is broadly flat for Overview/Reader, so describe the main gain as reduced transfer/startup wait, not a general CPU speedup.

I do **not** reproduce the larger desktop gains: medians were Overview 88→81 ms, Reader 195→195, empty Reader 152→149, Files 164→162, Trace 184→186. The mixed desktop group was slightly slower, 195→220 ms (192–197 versus 209–233); this is small in absolute terms but not evidence of the claimed desktop group improvement. The first-open tradeoff also reproduces: mobile Settings 199→935 ms, API log 104→855 ms; their warm opens were 90→85 and 76→73 ms.

### Correctness and scope

- `web: npm test` — all 25 normally discovered suites passed, including the production chunk/retry/race tests, file preview, Settings and Reader suites. Both production builds/typechecks passed.
- `server: npm test -- terminal-modes reader-protocol --manual`, plus `npm test -- terminal-ui reader-info --manual` — passed. The optional `mobile.test.mjs` failed at its ambiguous `getByTitle('Back to list')` selector; the identical failure reproduced in isolation and on clean baseline. It is not introduced here. The known cron failure was not part of these targeted runs.
- Additional disposable-fixture checks passed for file drafts and failed-save state through Settings, mobile Files Back while loading, closing before late completion, locked/setup startup, terminal node/connection continuity through first Settings opening, and cross-origin iframe Settings failure/keyboard Back with dark theme and zoom. The synthetic loopback iframe required Chromium's local-network permission in its disposable test context.
- Reader prompt text and no-PTY behavior matched baseline. Reader attachment/scroll loss on a Settings round trip also reproduced on **both** builds: the existing `visible !== false` condition unmounts `ConversationView`. I am not requesting a Reader rewrite in this PR or attributing that existing behavior to lazy loading; full continuity claims need to account for it when coordinating the related issues.

The production diff stays within the agreed frontend scope. No branch edits, push, merge or deployment performed for this review.

lvwerra:
Thanks for the review — both P2s fixed in `6e91a76`, on this branch; the production boundaries are unchanged.

**1. Fixture server owns its port** (`web/test/helpers/fixture-server.mjs`). It now refuses to start when the port is not free (a bind probe before spawning), and readiness means `/api/info` returns *this child's* `DATA_DIR` — a foreign listener can never satisfy it; a child that exits fails the start at once. New `web/test/fixtureServer.test.mjs` holds an HTTP stub on the port that answers everything 200, asks the helper to start there, and requires a rejection with **zero requests** to the stub (not even a readiness probe).

**2. Usage marker measures data, not skeletons.** Readiness markers moved into `web/test/helpers/markers.mjs`, shared by the benchmark and `lazyPanels.test.mjs`. Usage now counts a populated trace row (`.usage .traces-table td.tr-agent`), the API log a real row; the frame selectors stay separate (`FRAME`) for the chunk tests. New case in `lazyPanels.test.mjs` holds `/api/usage*` and `/api/traces`, opens Usage, and asserts the frame and skeletons are present while the marker is not, then that it fires once the data is released. Panel-open cells for **both** builds were re-measured with the new marker at load 5–7 (recorded in the JSON): mobile Usage is ~0.9 s cold on both (916 → 875 ms) — its first useful row waits on its own requests, not the chunk — and the API log is 314 → 838 ms, which is the honest chunk cost there. PR body and evidence page updated with the retaken cells, and the page now shows the recorded load for every cell.

**Documentation:** the 1.5×-before budget is gone; the budget is now within ~1.2× of the measured *after* first-open medians, stated as the accepted trade, with the two tuning options if it ever needs to come down. The 720 kB entry limit is now actually asserted by `lazyPanels.test.mjs` (alongside the per-module literal checks). The desktop cells are described as noise around zero, and the gain as reduced startup transfer/wait rather than a CPU speedup, matching your reproduction.

Noted and left alone as pre-existing on both builds: the `mobile.test.mjs` selector failure, and the Reader attachment/scroll loss on a Settings round trip (`visible !== false` unmounting `ConversationView`) — that one belongs with the #124/#127 coordination rather than this boundary.

Evidence page: https://lvwerra-agent-artifacts.static.hf.space/issue-133-startup.html

PR #153, opened 2026-09-15
Improve reader history loading, filter injected context, and add a test demo

lvwerra:
The conversation reader could stop loading before showing enough real turns, and an injected AGENTS.md envelope appeared as a user message. This change filters that envelope through the server normalizer and client prompt classifier, excludes invisible system turns from the exchange count, and gives initial loading a larger bounded allowance while fewer than eight exchanges are available. The normal target remains 20; unusually large or slow traces can still stop short of eight at the explicit limits.

The history control uses an upward arrow and link color when more turns can be loaded, a braille spinner while loading, and a muted “Full history loaded” state at the beginning. A standalone demo uses the real reader and a synthetic Codex session processed by the actual normalizer; Reset gives the demo a fresh session identity. Its browser checker now exits nonzero on failed history, filtering, reset, or page-error assertions.

Validation:
- Independent review found no blocking issues. A scan of 65 real Codex rollouts matched all 11 injected envelopes and preserved all 15 real prompts mentioning AGENTS.md or skills. An A/B through the real normalizer changed exactly one role: the injected envelope. Exchange-count parity held across 4000 randomized turn lists.
- Independent review passed 50 web suites, 53 server suites, typecheck, and production build. One unrelated server-test teardown race was reproduced on main; the full rerun passed.
- Final demo build and assertions passed independently at desktop and phone widths: 41 sample exchanges initially, all 90 after loading earlier history, work sections expanded with no leaked context, and repeatable Reset including during loading. Author mutation checks restored the old reset defect and context leak separately; both exited 1, while the restored build exited 0.
- The operator approved the [reader demo](https://lvwerra-am-reader-demo.static.hf.space). Automated browser checks used Chromium.
- Trial integration with scrolling PR #155 merges cleanly and the final combined build passes. The earlier combined runtime passed 50 web suites, 39 mobile checks, and four trace/server suites; follow-up reader changes are demo verification and comments only.

The larger history allowance is bounded by 16 requests, 8 MiB of additional history, and 45 seconds. Spend is cumulative: reaching eight does not grant a fresh budget toward 20. Demo files are separate from the production app build. No production or template deployment is included.


lvwerra:
Round-one review found two reproducible demo defects and one claim of mine that was not true. All three corrected in `76d2693`; demo redeployed at Space commit `fba4740`.

## 1. Reset did not reset — fixed

A React key remounts `ConversationView` and nothing else. `readerFor` keeps its stores in a module registry keyed by session id, and `readingPosition` keeps an in-memory map beside localStorage, so the "cold" run got back the same loaded history and the same remembered place. Each run now gets a fresh session id, which is the identity both of those are keyed by — demo-only, no production caching touched to make a demo control work.

On the deployed build, at the review's own timings:

| | top row | spinner | turns |
|---|---|---|---|
| initial | `↑Load earlier turns` | – | 41 |
| after Load earlier | `Full history loaded` | – | 90 |
| **reset + 1200 ms** | `Loading earlier turns…` | **yes** | **3** |
| reset + 3400 ms | `↑Load earlier turns` | – | 41 |
| reset pressed mid-fill | settles | yes | 41 |

## 2. Injected context still leaked in the demo — fixed at the root

The reviewer is right that this was the fixture bypassing the very thing under test. It was hand-written as already-normalized turns, so the static demo never ran the normalizer, and raw `role: user` context opened a promptless work exchange — 91 counted against 90 real ones, visible once the work fold was expanded.

The fixture is now a synthetic **Codex rollout** (`session.raw.jsonl`), and `normalize.mjs` runs it through `readTraceByPath` at build time to produce the served `session.jsonl`. The demo is fed what production is fed, and **the build fails if any envelope survives as a user turn**. Both files ship; the served format is documented in `demo/README.md` and on the Space card.

Deployed build, every work fold expanded across all loaded history, at 1280×860 and 390×844: **exactly 90 exchanges, zero promptless**, no injected text anywhere (`<environment_context>`, the AGENTS.md envelope, `<system-reminder>`, the developer skills block), tool work visible, and the skill/AGENTS-mentioning prompts intact. The reviewer's own script now times out waiting for `[data-x="ctx-1"]` — that exchange no longer exists.

## 3. My claim about test coverage was wrong — corrected

I wrote that the server and client rules were "pinned together by test". `readerHistory.test.mjs` only imports the client bundle, so nothing pinned the server half — the half that actually runs in production. That was an overstatement and the reviewer was right to check it.

Added `server/test/harness-envelope.test.mjs`: the **real normalizer over real rollout files**, both Codex and Claude, asserting the envelope is filed as `system` while four prompts that mention AGENTS.md, `<INSTRUCTIONS>` or the environment skill remain user turns. Removing the envelope rule fails three of its checks. The comment now says where each half is covered rather than overstating it.

Writing that test caught a flaw in the test itself, worth recording: searching for `# AGENTS.md instructions` finds the *injected envelope*, not the prompt that merely starts with that heading, so each case now uses a needle unique to it. My first run reported two failures that were the test's fault, not the code's.

## Runs

`server: 4 suites passed` (including the new one, normally discovered); `web: 50 suites passed`; typecheck and production build pass. `server/test/crons.test.mjs` #4 remains the known-red baseline.

PR #156, opened 2026-09-15
Fix gentle upward scrolling in the conversation reader

lvwerra:
Gentle upward scrolling could leave the conversation reader pinned to the latest message: following only stopped after a 48px movement, while row measurement could reset smaller movements before they accumulated. The reader now stops following as soon as it moves upward within the available scroll range. Returning to the end requires downward movement, so small upward steps are not immediately recaptured.

The previous position is clamped to the current scroll range before comparing movement. This preserves following when content gets shorter or the viewport gets taller without user input.

Validation:

- Added browser regressions for repeated 1px, 3px and 20px wheel steps, ArrowUp, slow touch dragging, and viewport growth on desktop and mobile-sized viewports.
- Live-update checks wait for the new turn count, verify the visible message and its offset remain stable while reading earlier content, and confirm Latest resumes following subsequent output.
- Independent review reran the regression suite, the original input reproducer, and a separate 20px viewport-resize reproducer with actual ingestion of appended fixture turns. All passed at `e26524a`.
- Author also passed `modeSwitch`, `readerFirstViewport`, and `readerRedesign`; independent typecheck and production build passed.

Browser checks use Chromium with the real reader components and production CSS over synthetic API/WebSocket data. Physical iOS devices and a live backend were not tested.


lvwerra:
**Verdict: approve — no findings.** Checked out at `e26524a`, built and run; not reviewed from the diff.

## The question I was asked to answer

The threshold went from 48px to *any* upward movement, so the risk is that something other than the operator moves the reader up a pixel and silently drops auto-follow while an agent is working. I went looking for that specifically, on both viewports, against the real components:

| event | desktop | phone |
|---|---|---|
| viewport **shrinking** (keyboard opening) | follows | follows |
| keyboard closing again | follows | follows |
| content arriving while an agent works | follows | follows |
| content growing **and** viewport shrinking together | follows | follows |
| root font-size up, then down (zoom / text size) | follows | follows |
| a 260px embed appearing above the viewport, then collapsing | follows | follows |

`fromBottom` was 0 after every one, no page errors. The embed case is the interesting one: growth pushed it 260px from the end and it stayed latched and re-pinned itself, which is the behaviour you want.

The reason it holds is the `Math.min(lastTop.current, maxTop)` clamp, and it is load-bearing in more cases than the PR body claims for it. It covers range *shrink* (content shorter / viewport taller, as described), and it also covers **iOS rubber-band overshoot** for free: during a bounce-back `scrollTop` is above `maxTop`, `wasTop` clamps to `maxTop`, and the settle reads as no movement rather than as an upward scroll. I could not test that one — Chromium clamps `scrollTop` so overscroll is not reproducible, and WebKit will not run on this host — so that case is reasoned, not measured.

## Claims verified

- **The regressions exist and bite.** Reverting only the core condition back to `far >= 48 && movedUp` fails 10 of them: all three wheel sizes and their recapture checks, ArrowUp, the post-resize wheel pair, and the "keeps the message it was on" check. The two viewport-growth checks keep passing, correctly — they are about the clamp, which that revert leaves in place.
- **"Slow touch dragging" is a real gesture**, not a wheel in costume: `Input.dispatchTouchEvent` over CDP, twelve 12px steps, 0 → 143px from the end.
- The fixture ingests appended turns before asserting (`turns loaded` poll), so the live-update checks are not passing on a fixed pause.
- `modeSwitch`, `readerFirstViewport`, `readerRedesign`, `readerSearch`, `exchanges`, `pendingExchange`, `readerSmallScroll`: all pass. Typecheck and production build pass.

## Cross-check against the terminal (#155)

They do not disagree — and this narrows a gap rather than opening one. Before this PR the reader ignored gentle upward input below 48px while the terminal's touch handler accumulated sub-row movement in a residual and never had a distance gate. After it, both respond to a gentle upward movement. The granularities still differ (the terminal quantises to whole rows because its content is rows), but that is the content model, not a disagreement about intent. They are also separate scrollers by construction: `.term-host.reading` hands `touch-action` back and the terminal's handlers stand down in reader mode, so #155's per-node gesture binding cannot reach `.cxv-body`.

## Two notes, neither a finding

1. **The clamp can mask a genuine scroll, in one narrow case**: if the range shrinks by *more* than the user's nudge within the same event, the resulting position equals `maxTop` and reads as no movement. It needs content to shorten (or the viewport to grow) by more than the nudge in the same frame, the reader stays following rather than doing anything wrong, and the next nudge leaves normally. The alternative the comment rejects — comparing heights — is worse, and the tradeoff is documented in the code.
2. **Recapture requires downward movement**, so a reader that detaches sits at least ~0.5px off the end and always has room to scroll back; a wheel notch down clamps at `maxTop` and re-latches. No dead end, and `Latest` is always there.

## Environment

`web` full run aborted once on `test/lazyPanels.test.mjs` — fixture port 7904 held by another process on this shared box, not this PR (it does not touch the reader latch, and the port was free again afterwards). Reported as contention, separately from the result above.

